PR #253, opened 2023-08-25
Provide a root flowr script

EagleoutIce:
Current State of the Repl

- [x] directly evaluate R expressions
- [x] call other scripts
- [x] help
- [x] automatic revival of the `RShellSesssion`
- [x] provide the current version of R
- [x] allow to present just the ast or dataflow graph of a given input
- [x] provide mermaid links if asked
- [x] allow to give a port to work in sever mode and not stdin/stdout communication
- [x] `-e`/`--execute` option to directly execute a command and exit
- [x] allow to pass multiple instructions, separated by a semicolon
- [x] Improve the currently convoluted separation of scripts.
- [x] set it as the default for the docker image. the repl should be everywhere :D

As originally unwanted side effect, this contains a *big* refactoring of flowR's core. 

- testing => outsourced to
  #272 
- document it in the wiki (probably as part of the overview) => outsourced to
  #276


EagleoutIce:
Additionally, it would be nice to have an `-e`/`--execute` option that just executes the given command and exits afterwards

EagleoutIce:
I currently have to continue porting the internal logic to the new `SteppingSlicer`

PR #377, opened 2023-10-04
Produce an Explicit CFG

EagleoutIce:
Still to be done:
 * [x] handle access tokens
 * [x] allow to access with the server and the REPL
 * [x] export as Quads (includes DF)
 * [x] testing (at least borderline...)


EagleoutIce:
@jofranky, sadly i was unable to complete the work today, but here is the current state for the CFG export in case you are interest (the pipeline is ensured to fail horribly because there are still TODO comments everywhere ^^).

PR #397, opened 2023-10-12
Basic Structure for the Abstract Interpretation

LukasPietzschmann:


src/core/output.ts
```diff
@@ -14,7 +14,7 @@ type StepResultsHelper<InterestedIn extends StepName> = {
 	'parse':       Out<'parse'>
 	'normalize':   StepResultsHelper<'parse'> & Out<'normalize'>
 	'dataflow':    StepResultsHelper<'normalize'> & Out<'dataflow'>
-	'slice':       StepResultsHelper<'dataflow'> & Out<'slice'>
+	'ai':          StepResultsHelper<'normalize'> & StepResultsHelper<'dataflow'> & Out<'ai'>
```
EagleoutIce, line 17:
```suggestion
	'ai':          StepResultsHelper<'dataflow'> & Out<'ai'>
```
The `normalize` step is included recursively

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+}
+
+interface IHandler<ValueType> {
+	name:  string,
+	enter: () => void
+	exit:  () => ValueType
+	next:  (value: ValueType) => void
+}
+
+type Intervals = Set<{min: number, minInclusive: boolean, max: number, maxInclusive: boolean}>
```
EagleoutIce, line 29: How about outsourcing the interval into its own interface?

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+
+interface Constraints {
+	node:      NodeId,
+	intervals: Intervals,
+	debugMsg:  string
+}
+
+function intervalFromNumber(n: number): Intervals {
```
EagleoutIce, line 37: we should separate between a single/continuous interval and a disjoint set of intervals (how about `domain`/`valueDomain`?) .
From that perspective, this lifts a scalar into the "domain"-world.

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+	const dfgNode: [DataflowGraphVertexInfo, OutgoingEdges] | undefined = dfg.graph.get(node)
+	guard(dfgNode !== undefined, `No DFG-Node found with ID ${node}`)
+	const [_, children] = dfgNode
+	const ids = Array.from(children.entries())
+		.filter(([_, edge]) => edge.types.has(EdgeType.Reads))
+		.map(([id, _]) => id)
+	if(ids.length === 0) {
+		return new Set()
+	}
+	guard(ids.length === 1, `Multiple reads-edges found for node ${node}`)
```
EagleoutIce, line 51: But this is definitely possible, consider:
```r
if(runif(1)) x <- 1; else x <- 2; 
print(x)
```

This should consist of the union of all domains (for now, i keep that name even if we later disagree).

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+	node:      NodeId,
+	intervals: Intervals,
+	debugMsg:  string
+}
+
+function intervalFromNumber(n: number): Intervals {
+	return new Set([{min: n, minInclusive: true, max: n, maxInclusive: true}])
+}
+
+function getIntervalsOfDfgChild(node: NodeId, dfg: DataflowInformation): Intervals {
```
EagleoutIce, line 41: Child appears to be a little vague. Maybe `leaf` is a better fit? Or just narrow it down to symbols atm.

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+	const id = ids[0]
+	const constraint = constraintMap.get(id)
+	guard(constraint !== undefined, `No constraint found for ID ${id}`)
+	return constraint.intervals
+}
+
+class Assignment implements IHandler<Constraints> {
+	private lhs:           NodeId | undefined
+	private rhs:           NodeId | undefined
+	private readonly node: RBinaryOp<ParentInformation>
```
EagleoutIce, line 61: There is a more narrow type alternative: `RAssignmentOp<ParentInformation>`.

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+
+	enter(): void {
+		console.log(`Entered ${this.name} ${this.node.info.id}`)
+	}
+
+	exit(): Constraints {
+		console.log(`Exited ${this.name} ${this.node.info.id}`)
+		return {
+			node:      this.lhs as NodeId,
+			intervals: new Set(), // TODO: check interval of the rhs
```
EagleoutIce, line 76: in general, not just the rhs but the _source_ of the assignment (`x <- source`, `source -> x`). This will be easier with the dataflow v2 update which will just have one ""assign"" it handles.

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+			debugMsg:  this.name
+		}
+	}
+
+	next(constraint: Constraints): void {
```
EagleoutIce, line 81: I know this is the way the cfg looks and we evaluate them. Yet it appears to be.. i dunno, so "magic" to just assume that next will yield the expected next child. Maybe we can leverage the node visiting context to at least guard we have not yet left the original expression (or ended up in a weird nesting).

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
-	console.log('hi from abstract interpretation')
+	const cfg = extractCFG(ast)
+	const operationStack = new Stack<IHandler<Constraints>>()
+	visitCfg(cfg, (node, _) => {
+		const astNode = ast.idMap.get(node.id)
+		// TODO: avoid if-else
+		if(astNode?.type === RType.BinaryOp) {
+			switch(astNode.flavor) {
+				case 'assignment': operationStack.push(new Assignment(astNode)).enter(); break
+				case 'arithmetic': operationStack.push(new BinOp(astNode)).enter; break
```
EagleoutIce, line 127: I think you forgot the parenthesis to call `enter`.

src/abstract-interpretation/processor.ts
```diff
@@ -1,7 +1,157 @@
...
+				default: guard(false, `Unknown binary op ${astNode.flavor}`)
+			}
+		} else if(astNode?.type === RType.Symbol) {
+			operationStack.peek()?.next({
```
EagleoutIce, line 131: we should robustify this. What if the program does have no assignment, binop, ...
For example, if its just `x`. While simply ignoring it is correct in this scenario, it is a potential problem with implicit returns and function definitions (again, dataflow v2 should replace those with explicit returns).

EagleoutIce:
Just a note so we do not forget it:
 * we need a way to detect if interval tracking is not beneficial/valid (e.g,. if we see an `x <- TRUE`).
 * we should have a single point to define what the worst-case assumption represents (do we want to use `Infinity` or mark such fields separately?

src/abstract-interpretation/processor.ts
LukasPietzschmann: Just to make sure, we're on the same page: You're suggesting to rename a continuous interval to `Domain`?

src/abstract-interpretation/processor.ts
LukasPietzschmann: Yeah, I'm aware that this can happen. I just wanted to ignore it for this commit. But a "TODO" would've been better than guarding against this case.

src/abstract-interpretation/processor.ts
LukasPietzschmann: Also fixed

src/abstract-interpretation/processor.ts
LukasPietzschmann: Oh yeah, I forgot that the assignments source must not always be on the rhs in R.

src/abstract-interpretation/processor.ts
```diff
@@ -0,0 +1,196 @@
...
+
+export function runAbstractInterpretation(ast: NormalizedAst, dfg: DataflowInformation): DataflowInformation {
+	const cfg = extractCFG(ast)
+	const operationStack = new Stack<IHandler<Constraints>>()
+	visitCfg(cfg, (node, _) => {
+		const astNode = ast.idMap.get(node.id)
+		// TODO: avoid if-else
+		if(astNode?.type === RType.BinaryOp) {
+			switch(astNode.flavor) {
+				case 'assignment': operationStack.push(new Assignment(astNode as RAssignmentOp<ParentInformation>)).enter(); break
```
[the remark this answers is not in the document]
LukasPietzschmann: Yeah, I thought so too. But TypeScript complaints without the cast.

src/abstract-interpretation/processor.ts
```diff
@@ -0,0 +1,196 @@
...
+	domains.forEach(domain => allIntervals.push(...domain))
+	allIntervals.sort(compareIntervals)
+	if(!allIntervals.length) {
+		return new Set<Interval>()
+	}
+
+	const resultingDomain = new Set<Interval>()
+	let currentInterval = allIntervals[0]
+	allIntervals.forEach(interval => {
+		if(interval.min < currentInterval.max) {
```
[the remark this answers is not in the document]
LukasPietzschmann: We now have `doIntervalsOverlap` which made the code way more readable (at least in my opinion xD)

src/abstract-interpretation/domain.ts
```diff
@@ -0,0 +1,166 @@
...
+export function unifyIntervals(intervals: Interval[]): Interval[] {
+	const sortedIntervals = intervals.sort(compareIntervalsByTheirMinimum)
+	if(sortedIntervals.length === 0) {
+		return []
+	}
+
+	const unifiedIntervals: Interval[] = []
+	let currentInterval = sortedIntervals[0]
+	for(const nextInterval of sortedIntervals) {
+		if(doIntervalsOverlap(currentInterval, nextInterval)) {
```
EagleoutIce, line 116: i mean theoretically we can heavily abuse sorting here (e.g. only update the max of following intervals and keep the interval start. Even if we do not require that optimization it may be a good idea to persist it in an issue.

src/abstract-interpretation/domain.ts
```diff
@@ -0,0 +1,166 @@
...
+			currentInterval = nextInterval
+		}
+	}
+	unifiedIntervals.push(currentInterval)
+	return unifiedIntervals
+}
+
+export function addDomains(domain1: Domain, domain2: Domain): Domain {
+	const intervals = new Set<Interval>()
+	for(const interval1 of domain1.intervals) {
```
EagleoutIce, line 131: Maybe we should abstract away the cartesian product to only define max and min.

src/abstract-interpretation/domain.ts
```diff
@@ -0,0 +1,166 @@
...
+			}, {
+				value:     interval1.max.value - interval2.min.value,
+				inclusive: interval1.max.inclusive && interval2.min.inclusive
+			}))
+		}
+	}
+	return new Domain(...intervals)
+}
+
+export function domainFromScalar(n: number): Domain {
```
EagleoutIce, line 161: i am not too sure but conceptually i am torn if this should be (statically) part of Domain.

src/abstract-interpretation/handler/binop/binop.ts
```diff
@@ -0,0 +1,45 @@
...
+	constructor(node: RBinaryOp<ParentInformation>) {
+		this.node = node
+	}
+
+	getName(): string {
+		return `Bin Op (${this.node.flavor})`
+	}
+
+	enter(): void {
+		console.log(`Entered ${this.getName()}`)
```
EagleoutIce, line 25: please switch to the flowr logger (you can create your own sublogger: `log.getSubLogger({ name: this.options.sessionName })`)

src/abstract-interpretation/handler/binop/binop.ts
```diff
@@ -0,0 +1,45 @@
+import {Handler} from '../handler'
+import {AINode} from '../../processor'
+import {BinaryOperatorFlavor, ParentInformation, RBinaryOp} from '../../../r-bridge'
+import {guard} from '../../../util/assert'
+import {operators} from './operators'
+
+export type BinOpOperators = {
```
EagleoutIce, line 7: why is the type here but the operators are not?

test/functionality/abstract-interpretation/abstract-interpretation.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+		domain1 = new Domain(new Interval({value: 6, inclusive: true}, {value: 9, inclusive: true}))
+		domain2 = new Domain(new Interval({value: 4, inclusive: true}, {value: 5, inclusive: true}))
+		assert.deepEqual(
+			subtractDomains(domain1, domain2),
+			new Domain(new Interval({value: 1, inclusive: true}, {value: 5, inclusive: true})),
+			'Subtracting two domains with overlapping intervals, subtracts the intervals'
+		)
+
+		// TODO: more tests for in and excluded bounds
+	})
```
EagleoutIce, line 137: Should i start writing tests for the handler/system as a whole?

src/abstract-interpretation/domain.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export class Interval {
+	readonly min: IntervalBound
+	readonly max: IntervalBound
+
+	constructor(min: IntervalBound, max: IntervalBound) {
+		this.min = min
+		this.max = max
+		guard(min.value <= max.value, `The interval ${this.toString()} has a minimum that is greater than its maximum`)
```
EagleoutIce, line 15: to string may be expensive (too expensive) wrt a closure, maybe use a call?

src/abstract-interpretation/handler/binop/binop.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	enter(): void {
+		console.log(`Entered ${this.getName()}`)
+	}
+
+	exit(): AINode {
+		console.log(`Exited ${this.getName()}`)
+		guard(this.lhs !== undefined, `No LHS found for assignment ${this.node.info.id}`)
+		guard(this.rhs !== undefined, `No RHS found for assignment ${this.node.info.id}`)
+		return operators[this.node.flavor](this.lhs, this.rhs, this.node)
```
EagleoutIce, line 32: Maybe we should abstract over the operators as well. If we want to do, say, type inference nothing from BinOp has to change besides the operator processing (and the generic but this is a type-level thing). So we could reuse these components.
Is a factory good for that? i am not too sure, but it may be sufficient to assign that statically for now (but outside of the BinOp scope).

src/abstract-interpretation/processor.ts
```diff
@@ -0,0 +1,82 @@
...
+import {BinOp} from './handler/binop/binop'
+import {Domain, domainFromScalar, unifyDomains} from './domain'
+
+export interface AINode {
+	readonly id:      NodeId
+	readonly domain:  Domain
+	readonly astNode: RNodeWithParent<ParentInformation>
+}
+
+const nodeMap = new Map<NodeId, AINode>()
```
EagleoutIce, line 17: this should not be global!

src/util/cfg/visitor.ts
```diff
@@ -0,0 +1,93 @@
...
+			const node = cfg.graph.vertices().get(id)
+			guard(node !== undefined, `Node with id ${id} not present`)
+			this.visitSingle(node, {parent: 'root', cfg, siblings: [...cfg.entryPoints], visited})
+		}
+	}
+
+}
+
+/**
+ * TODO
```
EagleoutIce, line 89: remove

test/functionality/abstract-interpretation/abstract-interpretation.spec.ts
EagleoutIce: Meeting: yes

EagleoutIce: ~> new issue

EagleoutIce (changes requested):
Make ready for merge!

src/abstract-interpretation/domain.ts
LukasPietzschmann: Tracked in #607

src/abstract-interpretation/handler/binop/binop.ts
LukasPietzschmann: Tracked in #608

src/abstract-interpretation/domain.ts
LukasPietzschmann: Tracked in #608

LukasPietzschmann:
@EagleoutIce So I cleared all open comments / TODOs. Theoretically, this PR should be mergeable now.

Buuuuut AI still makes many tests fail (mainly 'cause we don't know how to handle many cases). I think it's also a long term goal to make AI fail silently. Is this something for a new issue, or should this be implemented in this PR?

EagleoutIce:
As far as I am concerned, at least _passing_ these tests (e.g., by removing AI from the main pipeline for now) should be part of this PR.

src/abstract-interpretation/processor.ts
LukasPietzschmann: Tracked in #611

EagleoutIce (approved):

Issue #395, opened 2023-10-12
Memory leak in tmp folder for long-running parallel process!


EagleoutIce:

EagleoutIce:
we have a lot of `.R` files and many `Rtmp` folders when running parallel tasks!

EagleoutIce:
We still need a way to automatically unlink the (maybe empty) tempdir!

EagleoutIce:
Still exists on failed parse requests!

EagleoutIce:
Maybe, because the shell fails to parse and the recovery causes memory leak (on the read socket?)

EagleoutIce:
Correction: there are just files that are that big (or the DF extraction itself has the leak!).

Investigate: 

```
[...]

PR #411, opened 2023-10-13
Lint on git with clean history

EagleoutIce:


EagleoutIce:
Still has to be tested on windows

EagleoutIce:
Verified

.githooks/pre-push
```diff
@@ -2,16 +2,25 @@
...
+reset() {
+  echo "Restoring working tree"
+  git stash pop
+}
+
+if [ -n "$(git status --porcelain)" ]; then
+  echo "Working tree is not clean, temporary reset"
+  git stash save --keep-index --include-untracked
+  trap reset EXIT
+fi
```
LukasPietzschmann, line 20:
```suggestion
if [ -n "$(git status --porcelain)" ]; then
  echo "Working tree is not clean, make a temporary copy"
  tmpdir=$(mktemp --directory)
  cp -r . $tmpdir
  cd $tmpdir
  git stash save --quiet --keep-index --include-untracked
fi
```

LukasPietzschmann (changes requested):
Maybe not messing with the local version would be better. The overhead in copying the whole project to another folder is not great either, but (at least for me) it feels way better than stashing all the changes away while somebody could be working on them.

.githooks/pre-push
EagleoutIce: My version has a speedier copy by obeying the gitignore (the following reset is redundant as well but i placed it there to be save). Furthermore, i clean up with the reset trap to avoid overfilling the tmp folder.

LukasPietzschmann (approved):

.githooks/pre-push
LukasPietzschmann: > to avoid overfilling the tmp folder

Looks like you've had bad experiences with /tmp getting cluttered xD (cough #395)

EagleoutIce:
Again: verified on windows!

.githooks/pre-push
EagleoutIce: *cries in R*

PR #421, opened 2023-10-16
Test-fix: Global regex replacement to erase global ids even if there is more than one

EagleoutIce:


LukasPietzschmann (approved):
I've no idea whats going on, but [if it fits, it sits](https://knowyourmeme.com/memes/if-it-fits-i-sits)

Issue #471, opened 2023-11-02
Robustness against encoding problems


EagleoutIce:
Make flowR more robust against encoding problems in the data.

PR #477, opened 2023-11-05
Memory Leak on N-Quads Export

EagleoutIce:
It seems to be that the root issue (at least for the given file) was that the encoding caused the XML object conversion blew up beyond good and evil. I have added a filter, trying to be more robust (especially after #471).

@jofranky can you re-try your experiment with this state?


EagleoutIce:
New problematic file (txt as github seems to dislike .R files :c):
[midasqr.R.txt](https://github.com/Code-Inspect/flowr/files/13260554/midasqr.R.txt)

EagleoutIce:
Couldn't reproduce even with `export NODE_OPTIONS="--max-old-space-size=2048"`.

PR #517, opened 2023-11-24
Make it clearer, where to find linting information in the wiki

EagleoutIce:


EagleoutIce:
@Ehcsan, do you think that this makes it easier to know how to deal with the linter?

Ehcsan:
The changes make it easier to find information about the linter in its dedicated section. However while checking the changes I noticed that the wiki table of contend only redirects you to the home page (at least on the mobile version in chrome).

Ehcsan (approved):
The changes make the information about the linter easier to find

EagleoutIce:
> However while checking the changes I noticed that the wiki table of contend only redirects you to the home page (at least on the mobile version in chrome).

Yes, this is only until we merge this and the new wiki pages get deployed :)

LukasPietzschmann (approved):

PR #536, opened 2023-12-01
Keep Delimiters in normalized AST

EagleoutIce:


EagleoutIce:
@Ehcsan with this, you get the normalized tokens that i promised you :D

PR #571, opened 2023-12-15
Change _:cfg_ to _:cf:_ as a shortcut for _:controlflow_

### LINKED ISSUE #570: Change _:cfg_ to _:cf:_ as a shortcut for _:controlflow_
To show the dataflow graph for a given input, one can use _:dataflow_ or _:df_ (or _:d_).

_:controlflow_ does not follow this naming scheme, which is why I always mistype it 😢

LukasPietzschmann:
This PR resolves #570

### DIFF
--- a/src/cli/repl/commands/cfg.ts
+++ b/src/cli/repl/commands/cfg.ts
@@ -15,7 +15,7 @@ async function controlflow(shell: RShell, remainingLine: string) {
 export const controlflowCommand: ReplCommand = {
 	description:  'Get mermaid code for the control-flow graph of R code, start with \'file://\' to indicate a file',
 	usageExample: ':controlflow',
-	aliases:      [ 'cf' ],
+	aliases:      [ 'cfg', 'cf' ],
 	script:       false,
 	fn:           async(output, shell, remainingLine) => {
 		const result = await controlflow(shell, remainingLine)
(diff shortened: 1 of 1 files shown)


src/cli/repl/commands/cfg.ts
```diff
@@ -15,7 +15,7 @@ async function controlflow(shell: RShell, remainingLine: string) {
 export const controlflowCommand: ReplCommand = {
 	description:  'Get mermaid code for the control-flow graph of R code, start with \'file://\' to indicate a file',
 	usageExample: ':controlflow',
-	aliases:      [ 'cfg' ],
+	aliases:      [ 'cf' ],
```
EagleoutIce, line 18:
```suggestion
	aliases:      [ 'cfg', 'cf' ],
```

There should be nothing preventing us from allowing both!
Especially because this would essentially be a breaking change without allowing the old alias.

src/cli/repl/commands/cfg.ts
```diff
@@ -28,7 +28,7 @@ export const controlflowCommand: ReplCommand = {
 export const controlflowStarCommand: ReplCommand = {
 	description:  'Get a mermaid url of the control-flow graph of R code, start with \'file://\' to indicate a file',
 	usageExample: ':controlflow',
-	aliases:      [ 'cfg*' ],
+	aliases:      [ 'cf*' ],
```
EagleoutIce, line 31: Following the same comment as above:

```suggestion
	aliases:      [ 'cfg*', 'cf*' ],
```

EagleoutIce (approved):

PR #595, opened 2024-01-15
Drop `readlines/promises` Dependency

EagleoutIce:


EagleoutIce:
Additionally this makes the joi import more non-esm-interop compatible

PR #601, opened 2024-01-17
Fix: Escapes in the release title break everything

Ellpeck:
See https://stackoverflow.com/a/6697781

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,7 +37,7 @@ jobs:
         run: npm ci
       - name: Release
         run: |
-          msg="${{ github.event.head_commit.message }}"
+          msg='${{ github.event.head_commit.message }}'
           step=$(echo "$msg" | sed -n -E 's/\[release:(patch|minor|major)].*/\1/p')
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: npm ci
       - name: Release
         run: |
-          msg="${{ github.event.head_commit.message }}"
+          msg='${{ github.event.head_commit.message }}'
```
EagleoutIce, line 40: Let's hope that github will still evaluate it's variables :D furthermore, in general, single ticks could be used in the message to trick them again. but as long as only dedicated people can run this... we should be safe :eyes:

EagleoutIce (approved):

.github/workflows/release.yaml
Ellpeck: According to the docs, GH variables are evaluated in the workflow file before it is actually sent to a runner, so I doubt it cares about what is around the variables. 

Also yes the single tick thing iiiis a good point hmmmm

Issue #605, opened 2024-01-18
Feat: Follow up on `source` calls


EagleoutIce:
In R, users can include other files with the help of the [`source`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/source) function like this:

```R
# some amazing code
# source for the current environment
source("another/file/another/bit/of/fun.R")

# some amazing code which can refer to sourced variables
```

Now, we want to support this in `flowR`, for this we need to:
* [x] detect calls to `base::source` (i.e., check that `source` has not been redefined).
* [x] if found, follow the source (currently, only if the argument is given as a constant).
  * [x] this could be checked whenever the dataflow processor is checking a named function call
* [x] check if the file exists, if not, warn, otherwise retrieve the file
* [x] run `parse` and `normalize` on the file (in other words, the steps required before DF analysis)
* [x] run the dataflow analysis on the normalized, sourced file using the current environment

It should:
* [x] work recursively (`a.R` sources `b.r`, sources `c.R`, which may source `a.R` again). 
  it is fine if we use a hard cut-off to deal with recursion or reuse existing dataflow graphs to
  integrate until we reach a fixpoint
* [x] work within control flow constructs (i.e., if it happens within an if)
* [x] be general enough to be theoretically testable independent from the complete flowR pipeline

It would be nice if/it could be extended by:
* it is configurable (so that when starting the dataflow analysis one can configure flowR so that it ignores source calls to get the original behavior) (tracked in #642)
* [ ] graphs can be cached on a per-file basis in case one script chooses to source another one
  multiple times

- [x] For the tests, make sure that the "file provider abstraction" can be mocked so that we can use `source` in tests without actually referring to files on the host file system.

This is to be done using the `main` branch and independent from [`v2`](https://github.com/orgs/Code-Inspect/projects/5) (although the plan is to port it later).

PR #609, opened 2024-01-22
Feat: Follow up on source calls

Ellpeck:
VERY work in progress, I have no idea what I'm doing (YET)

### DIFF
--- a/src/dataflow/environments/environment.ts
+++ b/src/dataflow/environments/environment.ts
@@ -145,6 +145,14 @@ export const DefaultEnvironmentMemory = new Map<Identifier, IdentifierDefinition
 		definedAt: BuiltIn,
 		name:      'print',
 		nodeId:    BuiltIn
+	}]],
+	['source', [{
+		kind:      'built-in-function',
+		scope:     GlobalScope,
+		used:      'always',
+		definedAt: BuiltIn,
+		name:      'source',
+		nodeId:    BuiltIn
 	}]]
 ])
 
--- a/test/testfiles/source.R
+++ b/test/testfiles/source.R
@@ -1,0 +1,4 @@
+source("example.R")
+
+cat("-----\n")
+cat("Sourced N:", N, "\n")


src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -104,6 +104,27 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
+		if(sourceFile?.value?.type == RType.String) {
+			const request: RParseRequest = {
+				request:                'file',
+				content:                sourceFile.lexeme,
+				// TODO what does this do
+				ensurePackageInstalled: true
+			}
+			// TODO is this how to get a shell for this?
+			const shell = new RShell()
+			const slicer = new SteppingSlicer({
```
Ellpeck, line 118: For #615 to be useful for this, it looks like we might have to refactor and/or make a version of `SteppingSlicer` that uses the new synchronous shell. What do you think @EagleoutIce?

EagleoutIce: or we defer that to the pipeline executor (with an issue) and, for now, do the steps ourselves.

Ellpeck: That doesn't appear possible, since some of the steps themselves require `RShell` and are, thus, inherently asynchronous. For example, `parse` and `normalize` both do.

Ellpeck:
am I too stupid for a rebase, I click the "rebase" button and then it does this

Ellpeck:
@EagleoutIce as you can see, all regular tests will also fail with the same `dataflow` being `undefined` exception just due to the fact that the [source parsing code](https://github.com/Code-Inspect/flowr/pull/609/commits/c1d13460e252b6b65d8c16b7dd2504c6828a8bcb) *exists* - regardless of whether it's actually executed.

EagleoutIce:
First, I probably want to add `"importsNotUsedAsValues": "error",` to the tsconfig to prevent some of the possibilities (and deal with the upcoming errors) -> #616. 

Besides it may very well be a problem of you now requiring the steps in the dataflow, while the steps require the dataflow. Can you check if something like https://github.com/import-js/eslint-plugin-import is sufficient to detect this? (and if so, add it in a new issue?). 
Besides, can you break this cycle?

Ellpeck:
It looks like the [no-cycle rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) doesn't detect any issues with dependency cycles. I tried a few other, seemingly related rules, but none of them seemed to catch it either. Did you have a specific rule in mind?

Ellpeck:
`import type` did nothing to fix the cyclic dependency issue so that's just great

src/core/steps.ts
```diff
@@ -96,7 +96,8 @@ export const STEPS_PER_FILE = {
 	} satisfies IStep<typeof normalize>,
 	'dataflow': {
 		description: 'Construct the dataflow graph',
-		processor:   produceDataFlowGraph,
+		// TODO avoid cyclic dependency when using step executor in function-call (feels like a dirty hack)
```
EagleoutIce, line 99: technically it is no longer a todo? besides, it isn't tooo dirty

src/dataflow/environments/environment.ts
```diff
@@ -147,6 +147,15 @@ export const DefaultEnvironmentMemory = new Map<Identifier, IdentifierDefinition
 		definedAt: BuiltIn,
 		name:      'print',
 		nodeId:    BuiltIn
+	}]],
+	// TODO is this the correct definition for source?
```
EagleoutIce, line 151: from our current perspective, yes :)

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,6 +110,32 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
 		inIds.push(...functionName.in, ...functionName.unknownReferences)
 	}
 
+	// parse a source call and analyze the referenced code
+	if(named && functionCallName == 'source') {
```
EagleoutIce, line 114: this fails, whenever source is overwritten, you have to check if the function name really resolves to the builtin `source` for the current environment.

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,6 +110,32 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
+		const sourceFile = functionCall.arguments[0]
```
EagleoutIce, line 115: uff, relying on implicit undefined :D

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,6 +110,32 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
+		if(sourceFile?.value?.type == RType.String) {
+			const executor = new RShellExecutor()
+			const path = removeTokenMapQuotationMarks(sourceFile.lexeme)
+
+			// parse, normalize and dataflow the sourced file
+			const parsed = executeSingleSubStep('parse', {
+				request:                'file',
+				content:                path,
+				ensurePackageInstalled: true
```
EagleoutIce, line 124: i mean technically we shouldn't be here if it isn't but better be safe than sorry i guess.

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,6 +110,32 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
+			}, executor) as string
+			const normalized = executeSingleSubStep('normalize', parsed, executor.getTokenMap(), undefined, fileNameDeterministicCountingIdGenerator(path)) as NormalizedAst<OtherInfo & ParentInformation>
+			const dataflow = processDataflowFor(normalized.ast, { ...data, environments: finalEnv })
+
+			// update our graph with the sourced file's information
+			// TODO just set finalEnv, use overwriteEnvironments or appendEnvironments? makes no difference in the current example
```
EagleoutIce, line 130: the behavior is similar to that of an expression list. With the knowledge that the environment handles all set semantics, overwrite should be correct.

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,6 +110,32 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
+			finalEnv = overwriteEnvironments(finalEnv, dataflow.environments)
+			finalGraph.mergeWith(dataflow.graph)
+			// TODO is this the way it should be?? just changing the data ast seems fishy
```
EagleoutIce, line 133: hmm in a way the idmap should be everything, but know the id map and the ast are out of sync. maybe we should extend our id mapping to handle a collection of maps #abstracttheproblemsaway

Ellpeck:
did you just magically fix my rebase issues

src/r-bridge/lang-4.x/ast/parser/xml/internal/xml-to-json.ts
```diff
@@ -8,8 +8,11 @@ import type { XmlBasedJson } from '../input-format'
  * @param config    - The configuration to use (i.e., what names should be used for the attributes, children, ...)
  * @param xmlString - The xml input to parse
  */
-export function xlm2jsonObject(config: XmlParserConfig, xmlString: string): Promise<XmlBasedJson> {
-	return xml2js.parseStringPromise(xmlString, {
+export function xlm2jsonObject(config: XmlParserConfig, xmlString: string): XmlBasedJson {
+	let result: XmlBasedJson = {}
+	xml2js.parseString(xmlString, {
+		// we want this to be strictly synchronous!
```
EagleoutIce, line 14: maybe add the reason as well? so that the callback is synchronous? 

Yet we should maybe incorporate a test of transforming an artificailly huge file so that if this becomes problematic in the future... at least we have a chance of knowing that?
#newissuetime

src/r-bridge/retriever.ts
```diff
@@ -54,22 +55,37 @@ const ErrorMarker = 'err'
...
+		guard(output !== ErrorMarker, () => `unable to parse R code (see the log for more information) for request ${JSON.stringify(request)}}`)
+		return output
+	} else {
+		const run = async() => {
+			if(request.ensurePackageInstalled)
+				await shell.ensurePackageInstalled('xmlparsedata', true)
+
+			shell.sendCommands(...setupCommands)
+			const output = (await shell.sendCommandWithOutput(outputCommand)).join(shell.options.eol)
+			guard(output !== ErrorMarker, () => `unable to parse R code (see the log for more information) for request ${JSON.stringify(request)}}`)
```
EagleoutIce, line 83: somewhat of an ugly duplication for an error check :sweat_smile:

test/functionality/r-bridge/lang/ast/parse-values.ts
```diff
@@ -29,7 +29,7 @@ describe('Constant Parsing',
 					request:                'text',
 					content:                '{',
 					ensurePackageInstalled: true
-				}, shell))
+				}, shell) as Promise<string>)
```
EagleoutIce, line 32: without being able to track the context in GitHub (thanks GitHub) how about [AsyncOrSync](https://github.com/ts-essentials/ts-essentials/tree/master/lib/async-or-sync)?

EagleoutIce:
Still, we have to think about a way of mocking the file retrieval in the test environment.

src/dataflow/internal/process/functions/function-call.ts
Ellpeck: oh, I think I thought that's what `named` did

EagleoutIce:
additionally, we may want to make the sourcing-mechanic to be reusable so that we can, theoretically, work with libraries etc. as well (although attaching them actually adds a new environment in the env stack).

Furthermore, know is the best (#newissuetime) time to think about actually start testing behavioral equivalence. Or in other words: use the DF analysis to predict simple dependencies in data and than actually execute (known and safe) scripts to see if they actually show that behavior. 
This would help, especially when we start supporting more and more R versions.

src/dataflow/internal/process/functions/function-call.ts
Ellpeck: if we don't amend the id map like this, the final graph seems to be missing a bunch of id mappings, displaying variables & dependencies as `???` instead

src/dataflow/internal/process/functions/function-call.ts
EagleoutIce: named indicates that the function is not anonymous (e.g. the second call in ` produceCounter(0)()`).

src/dataflow/internal/process/functions/function-call.ts
EagleoutIce: yes because it relies on the map to produce all nodes. i think we can safely defer this is in an issue like "multi-file id handling". because now, with out-of-sync maps there should be better ways when we have something like an ID-Provider.

src/dataflow/internal/process/functions/function-call.ts
EagleoutIce: this should probably be part of a comment

src/dataflow/internal/process/functions/function-call.ts
Ellpeck: So we should just leave this code like it is (amending the original ast) and open an issue to improve upon that?

src/dataflow/internal/process/functions/source.ts
```diff
@@ -6,7 +7,7 @@ import {overwriteEnvironments} from '../../../environments'
 import type {DataflowInformation} from '../../info'
 
 export function processSourceCall<OtherInfo>(functionCall: RFunctionCall<OtherInfo & ParentInformation>, data: DataflowProcessorInformation<OtherInfo & ParentInformation>, information: DataflowInformation): DataflowInformation {
-	const sourceFile = functionCall.arguments[0]
+	const sourceFile = functionCall.arguments[0] as RArgument<ParentInformation>
```
EagleoutIce, line 10: actually i meant it the other way around as the 0th index can, theoretically, be undefined for if we encounter `source()`

Ellpeck: well you should've said so

Ellpeck: but i'm checking for that in the `if` right below: `sourceFile?.value?.type == RType.String`

src/dataflow/internal/process/functions/function-call.ts
EagleoutIce: yes

src/dataflow/internal/process/functions/source.ts
```diff
@@ -6,14 +6,18 @@ import {type DataflowProcessorInformation, processDataflowFor} from '../../../pr
...
-export const sourceFileProvider: (path: string) => RParseRequest = path => {
+let sourceFileProvider: (path: string) => RParseRequest = path => {
 	return {
 		request:                'file',
 		content:                path,
 		ensurePackageInstalled: true
 	}
 }
 
+export function setSourceFileProvider(provider: (path: string) => RParseRequest): void {
```
EagleoutIce, line 17: please refactor the type into a FileProvider, use the interface syntax.

src/dataflow/internal/process/functions/source.ts
```diff
@@ -6,16 +6,26 @@ import {type DataflowProcessorInformation, processDataflowFor} from '../../../pr
...
+			request:                'file',
+			content:                path,
+			ensurePackageInstalled: true
+		}
 	}
 }
 
-export function setSourceFileProvider(provider: (path: string) => RParseRequest): void {
-	sourceFileProvider = provider
+export function setTextSourceProvider(sources: Map<string,string>): void {
```
EagleoutIce, line 19: hmmm  ob man das im allgemeinen will? Also eig will ich imm allgemeinen ja schon auch dann für andere Semantiken ne path->request funktion basteln.

test/functionality/dataflow/processing-of-elements/functions/source-tests.ts
```diff
@@ -1,21 +1,14 @@
...
-import {setSourceFileProvider} from '../../../../../src/dataflow/internal/process/functions/source'
+import { setTextSourceProvider} from '../../../../../src/dataflow/internal/process/functions/source'
 import {BuiltIn, DataflowGraph, EdgeType, initializeCleanEnvironments} from '../../../../../src'
 import {LocalScope} from '../../../../../src/dataflow/environments/scopes'
 import {UnnamedArgumentPrefix} from '../../../../../src/dataflow/internal/process/functions/argument'
 import {define} from '../../../../../src/dataflow/environments'
 
 describe('source', withShell(shell => {
-	const sources = new Map<string, string>([
+	setTextSourceProvider(new Map<string, string>([
```
EagleoutIce, line 9: using varargs we could get rid of the map construction (at least in the tests :D)
just passing an object like:
`{ simple: 'N <- 9' }` should be more readable but this is not a must-have.

src/dataflow/processor.ts
```diff
@@ -13,20 +13,23 @@ export interface DataflowProcessorInformation<OtherInfo> {
...
    */
-	readonly activeScope:  DataflowScopeName
+	readonly activeScope:      DataflowScopeName
 	/**
    * Other processors to be called by the given functions
    */
-	readonly processors:   DataflowProcessors<OtherInfo>
+	readonly processors:       DataflowProcessors<OtherInfo>
+	// TODO using "initial" as the default path doesn't allow us to skip re-sourcing the initial file - how do we find out the initial file's name/path?
+	readonly currentPath:      string | 'initial'
```
Ellpeck, line 31: I'm really not sure if this is a good and/or valid way of dealing with "what file are we currently analyzing", since I also don't know if this is already being stored some way somewhere. Is this acceptable as a way of tracking this @EagleoutIce?

EagleoutIce: It is how i would have done it as, currently, there is only the current 'request'.... Yet instead of a path we should probably store the retrieval request!

Ellpeck: That's a good idea, I was actually wondering what to do when the request is just text instead of a file, but just storing the request directly would solve that as well. Yay!

src/dataflow/internal/process/functions/source.ts
```diff
@@ -28,6 +28,7 @@ export function processSourceCall<OtherInfo>(functionCall: RFunctionCall<OtherIn
 		const path = removeTokenMapQuotationMarks(sourceFile.lexeme)
 		const request = sourceProvider.createRequest(path)
 
+		// TODO we shouldn't skip a re-analysis *always*, just when it's a cycle - right?
```
Ellpeck, line 31: *Note to self*: It would make a lot more sense for this to, instead of storing a dictionary, just storing the "path" of files that we went down to get to the current source call. If the path already contains this file somewhere in it, it means we're in a loop!

Ellpeck:
I think this is mostly finished? I might be missing something huge and just totally forgetting about it, but all the major items from the issue #605 are implemented and should be working according to the tests :)

src/dataflow/internal/process/functions/function-call.ts
```diff
@@ -107,13 +107,19 @@ export function processFunctionCall<OtherInfo>(functionCall: RFunctionCall<Other
...
 		unknownReferences: [],
 		in:                inIds,
 		out:               functionName.out, // we do not keep argument out as it has been linked by the function
 		graph:             finalGraph,
 		environments:      finalEnv,
 		scope:             data.activeScope
 	}
+
+	// parse a source call and analyze the referenced code
+	if(isSourceCall(functionCallName, data.activeScope,finalEnv))
```
EagleoutIce, line 120: no braces :C

src/dataflow/internal/process/functions/source.ts
```diff
@@ -0,0 +1,68 @@
...
+import {dataflowLogger} from '../../../index'
+
+let sourceProvider = requestProviderFromFile()
+
+export function setSourceProvider(provider: RParseRequestProvider): void {
+	sourceProvider = provider
+}
+
+export function isSourceCall(name: Identifier, scope: DataflowScopeName, environments: REnvironmentInformation): boolean {
+	if(name != 'source')
```
EagleoutIce, line 19: this is not correct. Theoretically one could alias source like this:

`a <- source` 

and now the import can happen like `a("other.R")`

EagleoutIce: the check should happen for the resolved definition

src/dataflow/internal/process/functions/source.ts
```diff
@@ -0,0 +1,68 @@
...
+		return false
+	const definitions = resolveByName(name, scope, environments)
+	return definitions !== undefined && definitions.some(d => d.kind == 'built-in-function')
```
EagleoutIce, line 22: hmmm for now it is probably best to fail/warn if there are multiple definitions because then we must treat the complete import as a _maybe_ because it might do something different.

src/dataflow/internal/process/functions/source.ts
```diff
@@ -0,0 +1,68 @@
...
+export function processSourceCall<OtherInfo>(functionCall: RFunctionCall<OtherInfo & ParentInformation>, data: DataflowProcessorInformation<OtherInfo & ParentInformation>, information: DataflowInformation): DataflowInformation {
+	const sourceFile = functionCall.arguments[0] as RArgument<ParentInformation> | undefined
+	if(sourceFile?.value?.type == RType.String) {
+		const executor = new RShellExecutor()
+		const path = removeTokenMapQuotationMarks(sourceFile.lexeme)
+		const request = sourceProvider.createRequest(path)
+
+		// check if the sourced file has already been dataflow analyzed, and if so, skip it
+		const requestString = JSON.stringify(request)
+		if(data.referenceChain.some(r => JSON.stringify(r) == requestString)) {
```
EagleoutIce, line 34: given the repeated use -> please at least outsource a fingerprinting function. 
Furthermore, the check appears to be rather expensive for deeper reference chains.

src/dataflow/internal/process/functions/source.ts
```diff
@@ -0,0 +1,68 @@
...
+			dataflowLogger.info(`Found loop in dataflow analysis for ${requestString}: ${JSON.stringify(data.referenceChain)}, skipping further dataflow analysis`)
+			return information
+		}
+
+		// parse, normalize and dataflow the sourced file
+		let normalized: NormalizedAst<OtherInfo & ParentInformation>
+		let dataflow: DataflowInformation
+		try {
```
EagleoutIce, line 42: maybe we should outsource (haha) the actual sourcing part for reusability/to test it separately

src/dataflow/internal/process/functions/source.ts
```diff
@@ -0,0 +1,68 @@
...
+			dataflowLogger.warn(`Failed to analyze sourced file ${requestString}, skipping: ${(e as Error).message}`)
+			return information
+		}
+
+		// update our graph with the sourced file's information
+		const newInformation = {...information}
+		newInformation.environments = overwriteEnvironments(information.environments, dataflow.environments)
+		newInformation.graph.mergeWith(dataflow.graph)
+		// this can be improved, see issue #628
+		for(const [k, v] of normalized.idMap)
```
EagleoutIce, line 61: embrace the brace heavens!

src/dataflow/processor.ts
```diff
@@ -13,20 +13,22 @@ export interface DataflowProcessorInformation<OtherInfo> {
...
    * Name of the currently active scope, (hopefully) always {@link LocalScope | Local}
    */
-	readonly activeScope:  DataflowScopeName
+	readonly activeScope:    DataflowScopeName
 	/**
    * Other processors to be called by the given functions
    */
-	readonly processors:   DataflowProcessors<OtherInfo>
+	readonly processors:     DataflowProcessors<OtherInfo>
+	readonly currentRequest: RParseRequest
```
EagleoutIce, line 30: maybe a small comment explaining why we have two new fields and what they are for?

src/r-bridge/retriever.ts
```diff
@@ -25,6 +26,10 @@ interface RParseRequestBase {
 	ensurePackageInstalled: boolean
 }
 
+export interface RParseRequestProvider {
```
EagleoutIce, line 29: a small comment for why it exists (test stubs etc.)

test/functionality/dataflow/processing-of-elements/functions/source-tests.ts
```diff
@@ -0,0 +1,250 @@
...
+		.addEdge('3', '2', EdgeType.Argument, 'always')
+		.addEdge('3', BuiltIn, EdgeType.Reads, 'always')
+	)
+
+	const envWithX = define(
+		{nodeId: '0', scope: 'local', name: 'x', used: 'always', kind: 'variable', definedAt: '2' },
+		LocalScope,
+		initializeCleanEnvironments()
+	)
+	const recursive2Prefix = 'recursive2-2:1-2:6-'
```
EagleoutIce, line 183: better to construct this using the original function or? I mean now we can start a debate regarding what we want to test but i think the tests are fragile enough as-is.

src/dataflow/internal/process/functions/function-call.ts
Ellpeck: you're still so wrong for this but i will ""fix"" it anyway ❤️

EagleoutIce:
Please add an afterEach hook (?) that clears the source provider so that we can be sure they do not accidentally leak into other tests (keep it local to the dataflow tests, that should suffice).

EagleoutIce (approved):
Looks fine to me!

EagleoutIce:

PR #615, opened 2024-01-24
Refactor: Avoid async when interfacing with the R shell

Ellpeck:
Also known as: add entirely synchronous `RShellExecutor` class

### DIFF
--- a/src/r-bridge/shell-executor.ts
+++ b/src/r-bridge/shell-executor.ts
@@ -1,4 +1,4 @@
-import {DEFAULT_R_SHELL_OPTIONS, RShellOptions} from './shell'
+import {DEFAULT_R_SHELL_OPTIONS, RShellExecutionOptions} from './shell'
 import {deepMergeObject} from '../util/objects'
 import {spawnSync} from 'child_process'
 import {ts2r} from './lang-4.x'
@@ -9,14 +9,14 @@ import {log} from '../util/log'
 import fs from 'fs'
 
 export class RShellExecutor {
-	// TODO use a custom options class that doesn't have all the session stuff?
-	public readonly options:        Readonly<RShellOptions>
+
+	public readonly options:        Readonly<RShellExecutionOptions>
 	private readonly log:           Logger<ILogObj>
 	private readonly prerequisites: string[] = []
 
-	public constructor(options?: Partial<RShellOptions>) {
+	public constructor(options?: Partial<RShellExecutionOptions>) {
 		this.options = deepMergeObject(DEFAULT_R_SHELL_OPTIONS, options)
-		this.log = log.getSubLogger({ name: this.options.sessionName })
+		this.log = log.getSubLogger({ name: 'RShellExecutor' })
 	}
 
 	public continueOnError(): void {
--- a/src/r-bridge/shell-executor.ts
+++ b/src/r-bridge/shell-executor.ts
@@ -21,17 +21,19 @@ export class RShellExecutor {
 		this.log = log.getSubLogger({ name: 'RShellExecutor' })
 	}
 
-	public continueOnError(): void {
+	public continueOnError(): RShellExecutor {
 		this.log.info('continue in case of Errors')
 		this.prerequisites.push('options(error=function() {})')
+		return this
 	}
 
-	public injectLibPaths(...paths: string[]): void {
+	public injectLibPaths(...paths: string[]): RShellExecutor {
 		this.log.debug(`injecting lib paths ${JSON.stringify(paths)}`)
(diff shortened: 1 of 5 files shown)


src/r-bridge/shell-executor.ts
```diff
@@ -0,0 +1,40 @@
...
+import {deepMergeObject} from '../util/objects'
+import {spawnSync} from 'child_process'
+import {ts2r} from './lang-4.x'
+import {SemVer} from 'semver'
+import semver from 'semver/preload'
+import type {ILogObj, Logger} from 'tslog'
+import {log} from '../util/log'
+
+export class RShellExecutor {
+	// TODO use a custom options class that doesn't have all the session stuff?
```
EagleoutIce, line 11: Yes, we could let them extend each other

src/r-bridge/shell-executor.ts
```diff
@@ -0,0 +1,132 @@
...
+	public readonly options:        Readonly<RShellExecutionOptions>
+	private readonly log:           Logger<ILogObj>
+	private readonly prerequisites: string[] = []
+
+	public constructor(options?: Partial<RShellExecutionOptions>) {
+		this.options = deepMergeObject(DEFAULT_R_SHELL_OPTIONS, options)
+		this.log = log.getSubLogger({ name: 'RShellExecutor' })
+	}
+
+	public continueOnError(): void {
```
EagleoutIce, line 24: Given that these are constructor/builder functions? Why not encouraging the builder pattern by returning this? 
(although this is not necessary given that we will probably configure the executor only once).

src/r-bridge/shell-executor.ts
```diff
@@ -0,0 +1,132 @@
...
+		const tempDir = this.run([
+			'temp <- tempdir()',
+			`cat(temp, ${ts2r(this.options.eol)})`
+		])
+		log.info(`obtained temp dir ${tempDir}`)
+
+		const deleteOnExit = function() {
+			if(fs.existsSync(tempDir)) {
+				log.info(`deleting temp dir ${tempDir}`)
+				fs.rmSync(tempDir, {recursive: true, force: true})
```
EagleoutIce, line 95: things which destroyed steam... (fine)

src/r-bridge/shell-executor.ts
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	public run(commands: string | string[], returnErr = false): string {
+		this.log.trace(`> ${JSON.stringify(commands)}`)
+
+		const returns = spawnSync(this.options.pathToRExecutable, this.options.commandLineOptions, {
+			env:      this.options.env,
+			cwd:      this.options.cwd,
+			encoding: 'utf8',
+			input:    this.prerequisites.concat(typeof commands == 'string' ? [commands] : commands).join(this.options.eol)
```
EagleoutIce, line 127: You could use the spread operator here.

test/functionality/r-bridge/executor.ts
```diff
@@ -0,0 +1,77 @@
...
+		assert.isNotNull(semver.valid(version), `the version ${JSON.stringify(version)} should be a valid semver`)
+		assert.isTrue(semver.gt(version, '0.0.0'), `the version ${JSON.stringify(version)} should not be 0.0.0`)
+	})
+
+	it('ignore errors', () => {
+		const executor = new RShellExecutor()
+
+		// check regular result when an error occurs
+		const error = executor.run('a', true)
+		assert.match(error, /Error.*'a'/g)
```
EagleoutIce, line 23: this depends on the localization, so you require:

`options(warn=-1); invisible(Sys.setlocale("LC_MESSAGES", 'en_GB.UTF-8'))`

A test of the R sessions does exactly that (although i am still not 100% sure if the lines above are sufficient for every R version).

src/r-bridge/shell-executor.ts
```diff
@@ -0,0 +1,132 @@
...
+import {log} from '../util/log'
+import fs from 'fs'
+import {removeTokenMapQuotationMarks, TokenMap} from './retriever'
+import {DeepWritable} from 'ts-essentials'
+
+export class RShellExecutor {
+
+	public readonly options:        Readonly<RShellExecutionOptions>
+	private readonly log:           Logger<ILogObj>
+	private readonly prerequisites: string[] = []
```
EagleoutIce, line 17: Maybe allow to add arbitrary commands as prerequisites? 
In a way it may be nice to toggle some of those in the future? 

Yet, this can be outsourced into another issue :)

src/r-bridge/shell-executor.ts
EagleoutIce: that's why i added the `(fine)` :D

EagleoutIce (approved):

PR #625, opened 2024-01-31
Migrate CI `setup-node`

Ellpeck:
Use `v4` of `setup-node` in the setup action (it was already being used in the release action)


EagleoutIce:
we wait dearly for the current MR...

PR #640, opened 2024-02-05
Force Checked Index Access

Ellpeck:
It turns out that this causes a ***lot*** of errors, so I was thinking we should potentially use a different option than `noUncheckedIndexedAccess`, since a lot of the situations that this causes a warning in seem pretty safe tbh


Ellpeck:
(See [this comment](https://github.com/Code-Inspect/flowr/issues/634#issuecomment-1927127252) for why this likely isn't happening)

PR #641, opened 2024-02-05
Persist history of flowR REPL sessions

Ellpeck:
persistory


EagleoutIce:
Theoretically, we have a problem when someone runs multiple repls at the same time which overwrite each other but hey... we live only once!

EagleoutIce:
Yes, or we could keep a History-Tree but I think the current handling is sufficient as the most frequent use case will be the restart of the repl after compilation

EagleoutIce:

PR #651, opened 2024-02-07
 Add a startup configuration file for flowr

Ellpeck:


src/flowr.ts
```diff
@@ -157,6 +169,28 @@ async function mainServer(backend: Server = new NetServer()) {
...
+				// assign default values to all config options except for the specified ones
+				const ret = Object.assign(defaultConfigOptions, JSON.parse(read) as FlowrConfigOptions)
+				log.info(`Using config ${JSON.stringify(ret)} from ${configPath}`)
+				return ret
+			} catch(e) {
+				log.error(`Failed to parse config file at ${configPath}: ${(e as Error).message}`)
+			}
+		}
+		// move up to parent directory (apparently this is somehow the best way to do it in node, what)
+		searchPath = searchPath.split(path.sep).slice(0, -1).join(path.sep)
```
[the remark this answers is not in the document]
Ellpeck: unfortunately there are only 7 other ways and all of them are equally bad to this

Ellpeck:
I DID IT MOM MAIN IS HERE WITHOUT A MILLION COMMITS BEING MADE PART OF THE PR

src/config.ts
```diff
@@ -0,0 +1,38 @@
+import type { MergeableRecord} from './util/objects'
```
EagleoutIce, line 1: uff uneven spacing :C

EagleoutIce: let's burn this (are there eslint rules to prevent this? :D)

src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+import {deepMergeObject} from './util/objects'
+import path from 'path'
+import fs from 'fs'
+import {log} from './util/log'
+
+export interface FlowrConfigOptions extends MergeableRecord {
+	ignoreSourceCalls: boolean
+}
+
+export const defaultConfigFile = 'flowr.json'
```
EagleoutIce, line 11: i think i weakly recall using this as the example output file for the benchmarking somewhere. Could you please check?

src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+export const defaultConfigOptions: FlowrConfigOptions = {
+	ignoreSourceCalls: false
+}
+export const config = parseConfigOptions(process.cwd())
```
EagleoutIce, line 15: the over-engineer part of me wants to allow cli options for this but i think it is fine for now :)

src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function parseConfigOptions(workingDirectory: string, configFile = defaultConfigFile): FlowrConfigOptions {
+	let searchPath = path.resolve(workingDirectory)
+	do{
+		const configPath = path.join(searchPath, configFile)
+		if(fs.existsSync(configPath)) {
+			try {
+				const read = fs.readFileSync(configPath,{encoding: 'utf-8'})
+				// assign default values to all config options except for the specified ones
+				const ret = deepMergeObject(defaultConfigOptions, JSON.parse(read) as FlowrConfigOptions)
```
EagleoutIce, line 25: I think we should safe guard here with a strict json schema (as used for the network messages). Just to be safe(r) :D

src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+				log.info(`Using config ${JSON.stringify(ret)} from ${configPath}`)
+				return ret
+			} catch(e) {
+				log.error(`Failed to parse config file at ${configPath}: ${(e as Error).message}`)
+			}
+		}
+		// move up to parent directory (apparently this is somehow the best way to do it in node, what)
+		searchPath = searchPath.split(path.sep).slice(0, -1).join(path.sep)
```
EagleoutIce, line 33: i know its ugly and i am unaware of a better solution, but shouldn't we at least outsource this into a function with a nice name?

src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+	} while(fs.existsSync(searchPath))
+
+	log.info('Using default config')
```
EagleoutIce, line 36: should we reprint the config for more info in the future in case the default config changes?

src/config.ts
EagleoutIce: But at least the config file name should be configurable :eyes:

src/config.ts
Ellpeck: Seems like they're all called `benchmark.json` or `benchmark-[something].json`, so doesn't seem to be an issue. Nothing else seems to be called `flowr.json` either.

src/config.ts
```diff
@@ -0,0 +1,73 @@
...
+	ignoreSourceCalls: boolean
+}
+
+export const defaultConfigOptions: FlowrConfigOptions = {
+	ignoreSourceCalls: false
+}
+export const defaultConfigFile = 'flowr.json'
+
+const schema = Joi.object({
+	ignoreSourceCalls: Joi.string().optional()
```
EagleoutIce, line 19: Why a string and not a boolean?

src/config.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	// reset the config so it's reloaded next time
+	currentConfig = undefined
+}
+
+export function setConfig(config: FlowrConfigOptions) {
+	currentConfig = config
+}
+
+export function getConfig(): FlowrConfigOptions {
```
EagleoutIce, line 38: Sometime... we could rename this to flowrs global configuration :P

I understand why you made it lazy, but then, theoretically, as a user - i would assume that flowr would use the config file as it was when it started (e.g. when i am currently editing it to find out about its behavior) and not randomly whenever it is needed somewhere, so maybe we want to force the lazy eval in the main flowr scripts (especially to make it future proof with more and more configuration options being added :D)

src/config.ts
```diff
@@ -0,0 +1,73 @@
+import type {MergeableRecord} from './util/objects'
+import {deepMergeObject} from './util/objects'
+import path from 'path'
+import fs from 'fs'
+import {log} from './util/log'
+import {getParentDirectory} from './util/files'
+import Joi from 'joi'
+
+export interface FlowrConfigOptions extends MergeableRecord {
+	ignoreSourceCalls: boolean
```
EagleoutIce, line 10: we should at least document here, in a comment, what the purpose of the flag should be.

Furthermore, we may add documentation oft his to the wiki (you can defer the issue if you do not want to do this right now).

src/dataflow/internal/process/functions/source.ts
```diff
@@ -32,6 +33,11 @@ export function isSourceCall(name: Identifier, scope: DataflowScopeName, environ
 export function processSourceCall<OtherInfo>(functionCall: RFunctionCall<OtherInfo & ParentInformation>, data: DataflowProcessorInformation<OtherInfo & ParentInformation>, information: DataflowInformation): DataflowInformation {
 	const sourceFile = functionCall.arguments[0] as RArgument<ParentInformation> | undefined
 	if(sourceFile?.value?.type == RType.String) {
+		if(getConfig().ignoreSourceCalls) {
```
EagleoutIce, line 36: i am still unsure if i would want the check to happen as the first thing in this processor. Especially because we may add behavior for non-constant-string arguments with abstract interpretation etc. and to me it appears to be cleaner if it is a general guard

src/flowr.ts
```diff
@@ -81,6 +84,8 @@ if(options['no-ansi']) {
 	setFormatter(voidFormatter)
 }
 
+if(options['config-file'])
```
EagleoutIce, line 87: braces :sob: (:soap:)

but just as a quick check: does it works as well when someone gives relative and absolute paths (`../../.dotfiles/flowr.mega.config.json`, `/etc/trash/oldstuff/config`)?

src/config.ts
Ellpeck: because i'm incredibly smart

EagleoutIce (approved):

EagleoutIce:

EagleoutIce:

PR #665, opened 2024-02-20
Simplify Dataflow Tests

### LINKED ISSUE #158: Make the Creation of Dataflow Graphs in the Tests More Readable

bjthehun:
Closes #158, when merged.

- [x] Introduces a DataflowGraphBuilder.
- [x] The builder has convenience methods to easily add vertices/edges of a given type.
- [x] These methods replace addVertex/addEdge in dataflow tests.
- [x] Also removed default/optional parameters.


src/dataflow/graph/graph.ts
```diff
@@ -177,6 +179,33 @@ export class DataflowGraph {
...
+	/**
+	 * Adds a vertex for variable use. Intended for creating dataflow graphs as part of function tests.
+	 * @param id - AST node id
+	 * @param name - Variable name
+	 * @param when - always or maybe; defaults to always
+	 * @param environment - Environment the use occurs in
+	 * @param asRoot - boolean; defaults to true.
+	 * @param scope - string
+	 */
+	public uses(id: string, name: string, when: DataflowGraphEdgeAttribute = 'always', environment?: REnvironmentInformation, asRoot: boolean = true) {
```
bjthehun, line 192: The core of these changes are helper methods like `uses` here, a specialized version of `addVertex` with less noise (no object syntax, explicit tags, etc.). Eventually, I plan to add specialized versions of all vertex types.

src/dataflow/graph/graph.ts
```diff
@@ -267,6 +296,16 @@ export class DataflowGraph {
...
 	}
 
+	/**
+	 * Adds a read edge for simple testing.
+	 * 
+	 * @param from - Vertex/Node id
+	 * @param to   - see from
+	 * @param when - always (default), or maybe
+	 */
+	public reads(from: NodeId, to: NodeId, when: DataflowGraphEdgeAttribute = 'always') {
```
bjthehun, line 306: Like `uses`, `reads` is a simpler/more concise version of `addEdge`.

bjthehun: Of course, other versions for calls, arguments, defined-by etc. are still required.

EagleoutIce:
Looks great so far!

Hopefully,  I will be able to take a closer look in roughly 12-14 hours! 

Thank you very much for your great work!

src/dataflow/graph/graph.ts
```diff
@@ -177,6 +179,33 @@ export class DataflowGraph {
...
+	/**
+	 * Adds a vertex for a variable definition.
+	 * @param id - AST node ID
+	 * @param name - Variable name
+	 * @param scope - Scope (global/local/custom), defaults to local.
+	 * @param when - always or maybe; defaults to maybe.
+	 * @param environment - Environment the variable is defined in.
+	 * @param asRoot - boolean; defaults to true.
+	 */
+	public definesVariable(id: string, name: string, scope: string = LocalScope, when: DataflowGraphEdgeAttribute = 'always', environment?: REnvironmentInformation, asRoot: boolean = true) {
```
EagleoutIce, line 205: Is there a reason you do not use the `NodeId` type alias for the id? I consider changing the type.

src/dataflow/graph/graph.ts
```diff
@@ -177,6 +179,33 @@ export class DataflowGraph {
...
+	}
+
+	/**
+	 * Adds a vertex for a variable definition.
+	 * @param id - AST node ID
+	 * @param name - Variable name
+	 * @param scope - Scope (global/local/custom), defaults to local.
+	 * @param when - always or maybe; defaults to maybe.
+	 * @param environment - Environment the variable is defined in.
+	 * @param asRoot - boolean; defaults to true.
```
EagleoutIce, line 203:
```suggestion
	 * @param asRoot - should the vertex be part of the root vertex set of the graph (i.e., be a valid entry point) or is it nested (e.g., as part of a function definition)
```

src/dataflow/graph/graph.ts
EagleoutIce: Besides I am not too sure about the name `definesVariable` (which is long and focuses on the variable which is missing from a function like `uses`). 
How do you like `define(variable|function, ...)`, `use`, ...?

src/dataflow/graph/graph.ts
EagleoutIce: Furthermore, these are many parameters. What about a mixed approach? 
`uses(id: NodeId, name: string, args: { when: ... })` this would save us from writing a lot of `undefine`s if the `asRoot` flag is to be set explicitly and it wouldn't create too much overhead (we could use a deep merge for the defaults or even, if we use a builder wrapper, configure those on-the-fly).

test/functionality/dataflow/processing-of-elements/functions/function-call-tests.ts
```diff
@@ -310,9 +310,9 @@ a()()`,
 		assertDataflow('Calling with 1 + x', shell, 'foo(1 + x)',
 			new DataflowGraph()
 				.addVertex({ tag: 'function-call', id: '5', name: 'foo', environment: initializeCleanEnvironments(), args: [{ nodeId: '4', name: `${UnnamedArgumentPrefix}4`, scope: LocalScope, used: 'always' }]})
-				.addVertex({ tag: 'use', id: '4', name: `${UnnamedArgumentPrefix}4`})
-				.addVertex({ tag: 'use', id: '2', name: 'x'})
-				.addEdge('4', '2', EdgeType.Reads, 'always')
+				.uses('4', `${UnnamedArgumentPrefix}4`)
```
EagleoutIce, line 313: maybe there is even a way to avoid writing all of these unnamed argument prefixes :eyes:.

EagleoutIce (commented):
Thank you very much!

test/functionality/_helper/dataflowgraph-builder.ts
```diff
@@ -0,0 +1,46 @@
+import type { DataflowGraphEdgeAttribute, NodeId, REnvironmentInformation } from '../../../src'
+import { DataflowGraph, EdgeType } from '../../../src'
+import { LocalScope } from '../../../src/dataflow/environments/scopes'
+
+export function emptyGraph() {
+	return new DataflowGraphBuilder
```
EagleoutIce, line 6: I thought i had an eslint rule disallowing constructor calls without parenthesis (now tracked in https://github.com/Code-Inspect/flowr/issues/667).

EagleoutIce:
```suggestion
	return new DataflowGraphBuilder()
```

EagleoutIce (commented):
Thank you for switching to the builder layer!

test/functionality/dataflow/graph/equal-tests.ts
```diff
@@ -51,7 +51,7 @@ describe('Equal', () => {
 				const rhs = emptyGraph().reads('0', '1')
 				neq('Source Id', emptyGraph().reads('2', '1'), rhs)
 				neq('Target Id', emptyGraph().reads('0', '2'), rhs)
-				neq('Type', emptyGraph().addEdge('0', '1', EdgeType.Calls, 'always'), rhs)
+				neq('Type', emptyGraph().argument('0', '1', 'always'), rhs)
```
EagleoutIce, line 54: is there are reason you change calls edge to an argument?

bjthehun:
Dataflow graph building itself has been simplified. However, defining/using references and environments still is comparatively cumbersome.
I'd probably wait for the code review, plus #673, to work on that, because inconsistent spacing gave me quite some trouble while switching to the builder methods.

EagleoutIce:
I have just implemented #673.

bjthehun:
> I have just implemented #673.
> 
> @bjthehun Do you want me to handle all the new conflicts resulting from the unified spacing?

If you want to, then yes.

EagleoutIce:

EagleoutIce:

PR #676, opened 2024-02-23
lint: add curly rule and fix curly braces

Ellpeck:


Ellpeck:
@EagleoutIce we have to think about implementing a consistent brace style (see https://eslint.org/docs/latest/rules/brace-style) because right now, this PR causes the `--fix` action to create some *wild* formatting.

Let me know if we should make that part of this PR or open another issue for it!

EagleoutIce:
I think it should be part of this PR because otherwise fix makes the code... urgh

EagleoutIce:
I want the #onetruebracestyle

EagleoutIce (approved):

EagleoutIce:

PR #701, opened 2024-03-06
Linter: Combine Imports

Ellpeck:
So, all of the options for this rule kind of suck, so we have to compromise.

- [the eslint rule](https://eslint.org/docs/next/rules/no-duplicate-imports) doesn't allow auto-fixing, making fixing hundreds of source files a bit of a pain.
- The `import/no-duplicates` rule, which we're using now, is fine. However, we'd like to enable the `prefer-inline` option as well, to allow for type imports and regular imports to be merged. Unfortunately, that behavior [is currently broken](https://github.com/import-js/eslint-plugin-import/issues/2792), so we're not enabling it for now.

It seems like, for the time being, this is the best we can do.


EagleoutIce:
Thank you!

> Unfortunately, that behavior https://github.com/import-js/eslint-plugin-import/issues/2792, so we're not enabling it for now.

Can you create an issue for this?

EagleoutIce (approved):

EagleoutIce:

Issue #588, opened 2024-01-11
Test Abstract Interpretation


EagleoutIce:
Unit and basic system-level tests for abstract interpretation.

PR #706, opened 2024-03-08
Start work on supporting conditionals (let's start with if-else)

### LINKED ISSUE #617: Start work on supporting conditionals (let's start with if-else)
We need basic support for if-else structures so we can show something cool at the intermediate fair.

LukasPietzschmann:
closes #617

### DIFF
--- a/src/abstract-interpretation/ainode.ts
+++ b/src/abstract-interpretation/ainode.ts
@@ -2,13 +2,15 @@ import {NodeId, ParentInformation, RNodeWithParent} from '../r-bridge'
 import {assertUnreachable, guard} from '../util/assert'
 import {Domain, unifyDomains} from './domain'
 
-export interface AINode {
-	/** The ID of the node that logically holds the domain (e.g. the lhs of an assignment) */
-	readonly nodeId:       NodeId
-	/** The ID of the whole expression that the domain was calculated from (e.g. the whole assignment expression) */
-	readonly expressionId: NodeId
-	readonly domain:       Domain
-	readonly astNode:      RNodeWithParent<ParentInformation>
+export class AINode {
+	constructor(
+		public readonly domain: Domain,
+		public readonly astNode: RNodeWithParent<ParentInformation>,
+		/** The ID of the whole expression that the domain was calculated from (e.g. the whole assignment expression) */
+		public readonly expressionId: NodeId = astNode.info.id,
+		/** The ID of the node that logically holds the domain (e.g. the lhs of an assignment) */
+		public readonly nodeId: NodeId = astNode.info.id
+	) {}
 }
 
 export const enum RegisterBehavior {
@@ -80,7 +82,7 @@ export class AINodeStore implements Iterable<AINode> {
 				case RegisterBehavior.Merge: {
 					const existing = this.map.get(node.nodeId)
 					guard(existing !== undefined, `Node with ID ${node.nodeId} should exist`)
-					this.register({...node, domain: unifyDomains([existing.domain, node.domain])}, RegisterBehavior.Overwrite)
+					this.register(new AINode(unifyDomains([existing.domain, node.domain]), node.astNode, node.expressionId), RegisterBehavior.Overwrite)
 					break
 				}
 				default: assertUnreachable(behavior)
--- a/src/abstract-interpretation/ainode.ts
+++ b/src/abstract-interpretation/ainode.ts
@@ -11,6 +11,15 @@ export class AINode {
 		/** The ID of the node that logically holds the domain (e.g. the lhs of an assignment) */
 		public readonly nodeId: NodeId = astNode.info.id
 	) {}
+
+	static copy(node: AINode, changes: Partial<AINode>): AINode {
+		return new AINode(
+			changes.domain ?? node.domain,
+			changes.astNode ?? node.astNode,
+			changes.expressionId ?? node.expressionId,
+			changes.nodeId ?? node.nodeId
+		)
+	}
 }
 
 export const enum RegisterBehavior {
@@ -82,7 +91,7 @@ export class AINodeStore implements Iterable<AINode> {
 				case RegisterBehavior.Merge: {
 					const existing = this.map.get(node.nodeId)
 					guard(existing !== undefined, `Node with ID ${node.nodeId} should exist`)
-					this.register(new AINode(unifyDomains([existing.domain, node.domain]), node.astNode, node.expressionId), RegisterBehavior.Overwrite)
+					this.register(AINode.copy(node, {domain: unifyDomains([existing.domain, node.domain])}), RegisterBehavior.Overwrite)
 					break
 				}
 				default: assertUnreachable(behavior)
--- a/src/abstract-interpretation/ainode.ts
+++ b/src/abstract-interpretation/ainode.ts
@@ -3,9 +3,9 @@ import {assertUnreachable, guard} from '../util/assert'
 import {Domain, unifyDomains} from './domain'
 
 export interface AINode {
-	// The ID of the node that logically holds the domain (e.g. the lhs of an assignment)
+	/** The ID of the node that logically holds the domain (e.g. the lhs of an assignment) */
 	readonly nodeId:       NodeId
-	// The ID of the whole expression that the domain was calculated from (e.g. the whole assignment expression)
+	/** The ID of the whole expression that the domain was calculated from (e.g. the whole assignment expression) */
 	readonly expressionId: NodeId
 	readonly domain:       Domain
 	readonly astNode:      RNodeWithParent<ParentInformation>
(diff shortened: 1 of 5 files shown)


src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
+import { NodeId, ParentInformation, RNodeWithParent } from '../r-bridge'
+import { assertUnreachable, guard } from '../util/assert'
+import { Domain, unifyDomains } from './domain'
+
+export interface AINode {
+	// The ID of the node that logically holds the domain
+	readonly nodeId:       NodeId
+	// The ID of the whole expression that the domain was calculated from
+	readonly expressionId: NodeId
```
EagleoutIce, line 9: is expression the correct reference here? With the separation being arbitrary an example may be beneficial

src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
+import { NodeId, ParentInformation, RNodeWithParent } from '../r-bridge'
+import { assertUnreachable, guard } from '../util/assert'
+import { Domain, unifyDomains } from './domain'
+
+export interface AINode {
+	// The ID of the node that logically holds the domain
+	readonly nodeId:       NodeId
```
EagleoutIce, line 7: why not join this with the respective astNode? as they are inherently linked by the traversal it could be a direct decoration.

src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
...
+		} else if(content !== undefined) {
+			super([[content.nodeId, content]])
+		} else if(content === undefined) {
+			super()
+		} else {
+			assertUnreachable(content)
+		}
+	}
+
+	push(node: AINode): void {
```
EagleoutIce, line 28: given the map nature, something like `register` is probably a better name.

EagleoutIce: \+ given that we add a wrapper to use, how about guarding against overwriting?
I mean yes, we assume that every node is unique, but what if we add the same node multiple times.

src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
...
+		this.set(node.nodeId, node)
+	}
+}
+
+export function mergeDomainStores(...stores: AINodeStore[]): AINodeStore {
```
EagleoutIce, line 33:
```suggestion
export function mergeDomainStores(...stores: readonly AINodeStore[]): AINodeStore {
```

src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
...
+	const result = new AINodeStore()
+	for(const store of stores) {
+		for(const [id, node] of store) {
+			if(result.has(id)) {
+				const existing = result.get(id)
+				guard(existing !== undefined, `Domain for ID ${id} is missing`)
+				const unified = unifyDomains([existing.domain, node.domain])
+				result.set(id, {...node, domain: unified})
```
EagleoutIce, line 41: why is this not using the `push`/`register` functionality?

src/abstract-interpretation/ainode.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function mergeDomainStores(...stores: AINodeStore[]): AINodeStore {
+	const result = new AINodeStore()
+	for(const store of stores) {
+		for(const [id, node] of store) {
+			if(result.has(id)) {
+				const existing = result.get(id)
+				guard(existing !== undefined, `Domain for ID ${id} is missing`)
+				const unified = unifyDomains([existing.domain, node.domain])
```
EagleoutIce, line 40: Theoretically we can combine this as a fold/reduction providing the merging function - just in case in the future we add more than the domain.

src/abstract-interpretation/domain.ts
```diff
@@ -90,23 +90,41 @@ function compareIntervalsByTheirMaximum(interval1: Interval, interval2: Interval
 	return compareIntervals(CompareType.Max, interval1.max, interval2.max)
 }
 
-export function doIntervalsOverlap(interval1: Interval, interval2: Interval): boolean {
+export const enum OverlapKind {
+	Overlap = 0,
+	Touch = 1,
+}
+
+export function doIntervalsOverlap(interval1: Interval, interval2: Interval, kind: OverlapKind = OverlapKind.Overlap): boolean {
```
EagleoutIce, line 98: We definitely need tests for this.

src/abstract-interpretation/domain.ts
```diff
@@ -90,23 +90,41 @@ function compareIntervalsByTheirMaximum(interval1: Interval, interval2: Interval
...
+	let doIntervalsTouch = true
+
 	// If one interval ends before the other starts, they don't overlap
 	if(diff1 < 0 || diff2 < 0) {
-		return false
-	}
-	// If their end and start are equal, they only overlap if both are inclusive
-	if(diff1 === 0) {
-		return interval1.max.inclusive && interval2.min.inclusive
+		doIntervalsOverlap = false
```
EagleoutIce, line 107: It should suffice to track Overlap based on the requested mode.

src/abstract-interpretation/domain.ts
EagleoutIce: and fundamentally i think the function would be simpler if handled as an intersection of interval, checking if it is empty/scalar.

src/abstract-interpretation/domain.ts
```diff
@@ -166,4 +184,82 @@ export function subtractDomains(domain1: Domain, domain2: Domain): Domain {
...
+	intersection: Interval | undefined,
+	larger:       Interval | undefined
+}
+
+function flipInclusiveness(intervalBound: IntervalBound): IntervalBound {
+	return {value: intervalBound.value, inclusive: !intervalBound.inclusive}
+
+}
+
+export function overlapIntervals(interval1: Interval, interval2: Interval): IntervalOverlap {
```
EagleoutIce, line 206: warum das und `doIntervalsOverlap`?

src/util/cfg/cfg.ts
```diff
@@ -524,24 +524,25 @@ function cfgUnaryOp(unary: RNodeWithParent, operand: ControlFlowInformation): Co
 
 
 function cfgExprList(_node: RNodeWithParent, expressions: ControlFlowInformation[]): ControlFlowInformation {
```
EagleoutIce, line 526: not too sure whether we just just switch to basic blocks

src/abstract-interpretation/ainode.ts
LukasPietzschmann: I added an example to each comment. But as we're currently not using the `expressionId` at all, I might just remove it in the future (tbh I can't really remember why I added it in the first place 😅)

EagleoutIce: I think it was to indicated the respective parent expression when operating on the sub-expression level.

src/abstract-interpretation/ainode.ts
LukasPietzschmann: Sure, fixed it :) I added a `guard(...)` that guards against attempts that want to overwrite a value. When I spot a case that triggers the guard, I'll reevaluate if this is correct behavior (but it's probably not).

src/abstract-interpretation/ainode.ts
LukasPietzschmann: I also just noticed this when renaming the function 😅 I probably just forgot

src/abstract-interpretation/domain.ts
LukasPietzschmann: > We definitely need tests for this.

We already have some (six), but they don't yet cover all edge cases.

> and fundamentally i think the function would be simpler if handled as an intersection of interval, checking if it is empty/scalar.

This would be definitly more elegant and reusable. I'll write it down somewhere and change it when I get to it. For now, I prioritize making progress; refactorings like this can come later.

src/abstract-interpretation/domain.ts
LukasPietzschmann: `overlapIntervals` currently uses `doIntervalsOverlap`, but this could definitly be refactored, then `doIntervalsOverlap` would be no longer needed (or maybe just as a small wrapper for `overlapIntervals`).

src/abstract-interpretation/ainode.ts
LukasPietzschmann: Nvm, we do indeed use `register` to overwrite values. I added an extra flag that should be set when overwriting should be possible.

src/abstract-interpretation/domain.ts
LukasPietzschmann: This is now tracked in #588

src/abstract-interpretation/ainode.ts
LukasPietzschmann: Addressed in e6d3add0

src/util/cfg/cfg.ts
LukasPietzschmann: Tracked in #812

PR #714, opened 2024-03-11
Improve dependency management for flowr

Ellpeck:
Apparently webpack, in its ***infinite wisdom***, thinks `node:readline` means "readline is to be found by using the protocol `node://`", which is.. what.

But alas, `node:` is a redundant prefix so we just remove it. That should fix it. Apparently. I'm going to cry.


Ellpeck:
There doesn't appear to be a rule for *not* preferring the `node:` protocol prefix :(

EagleoutIce:

PR #717, opened 2024-03-11
Ensure low logging level on npm by default

Ellpeck:
All exisiting extension points, like the repl, the slicer and the tests, automatically set their log level to their desired level already, so this change will only have an effect on the npm package.


EagleoutIce (approved):

EagleoutIce:

PR #722, opened 2024-03-14
Fix: Repl Multi-Expression Prompts Are Still a Problem

Ellpeck:


Ellpeck:
Is this an acceptable way to deal with this issue? Strings are still left quoted for all REPL commands, but commands can then unquote them if they need them to be. I think this is the.. safest option? Maybe?

EagleoutIce (approved):

EagleoutIce:
I think they are fine

EagleoutIce:

PR #723, opened 2024-03-14
Remove references to xmlparsedata from docs

Ellpeck:


Ellpeck:
I'm not actually sure if removing the entire meta table makes sense, but I think those are all only used in the xml generated by xmlparsedata?

EagleoutIce:
The tokens are still very well of importance! I updated the patch accordingly.

EagleoutIce:

PR #727, opened 2024-03-15
Fix Repl Crashes for Invalid Input

Ellpeck:


Ellpeck:
hmm I'm realizing now that catching every exception in every command manually (in addition to the default catching i added in the first commit) may be a bit.. unnecessary? @EagleoutIce your opinion ✨

EagleoutIce:
I agree. Although it has the potential of allowing for :sparkles: better errors :sparkles:, the default handling appears to be sufficient.

EagleoutIce (approved):

EagleoutIce:

PR #745, opened 2024-03-25
refactor: move modules into their own subdirectory

Ellpeck:

PR #747, opened 2024-03-27
Update Dockerfile to reflect module splitting

Ellpeck:
Note that this cannot be tested yet since the modules don't have their dependencies resolved fully yet.

Also depends on #745.


Ellpeck:
This is done, but ***will only work once the privately discussed version path issue is solved.***

EagleoutIce:
_This pull request is included in v1.5.1 (see [Release v1.5.1 (Declutter flowR)](https://github.com/Code-Inspect/flowr/releases/tag/v1.5.1))._

PR #754, opened 2024-04-16
feat-fix: improve usedRVersion to return after a reasonable timeout

Ellpeck:
This improvement/fix also includes an additional property for `CollectorTimeout` that allows skipping the rejection when the timeout is reached.

### DIFF
--- a/src/r-bridge/shell.ts
+++ b/src/r-bridge/shell.ts
@@ -21,6 +21,10 @@ export interface CollectorTimeout extends MergeableRecord {
    * if true, the timeout will reset whenever we receive new data
    */
 	resetOnNewData: boolean
+	/**
+	 * If false, the promise will resolve with empty data, instead of rejecting, if the timeout is reached
+	 */
+	reject:         boolean
 }
 
 interface CollectorUntil extends MergeableRecord {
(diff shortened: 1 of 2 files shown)


src/r-bridge/shell.ts
```diff
@@ -53,7 +57,8 @@ export const DEFAULT_OUTPUT_COLLECTOR_CONFIGURATION: OutputCollectorConfiguratio
 	postamble: `🐧${'-'.repeat(5)}🐧`,
 	timeout:   {
 		ms:             750_000,
-		resetOnNewData: true
+		resetOnNewData: true,
+		reject:         true
```
EagleoutIce, line 61: i am not such a big fan of a `reject` flag. how about a more general purpose `onError` handler?

Ellpeck: fiiine

src/r-bridge/shell.ts
```diff
@@ -21,6 +22,10 @@ export interface CollectorTimeout extends MergeableRecord {
    * if true, the timeout will reset whenever we receive new data
    */
 	resetOnNewData: boolean
+	/**
+	 * invoked when the timeout is reached. If not set, the promise will be rejected with an error.
+	 */
+	onTimeout?:     (resolve: (value: AsyncOrSync<string[]>) => void, reject: (value: AsyncOrSync<string[]>) => void, partialResult: string[]) => void
```
EagleoutIce, line 28: i expected a handler for the error case in general but that is fine :)

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.0.0 (see [Release v2.0.0 (Dataflow v2)](https://github.com/Code-Inspect/flowr/releases/tag/v2.0.0))._

PR #760, opened 2024-04-30
feat-fix: increase r version timeout in r shell

Ellpeck:
This is an amend for #754, as it turns out through [testing](https://github.com/Code-Inspect/vscode-flowr/pull/79/commits/4569bf46c802ab1d549ea4c045c6e831e0056f26) that a one second timeout isn't enough on a slower machine.


EagleoutIce:
_This pull request is included in v2.0.0 (see [Release v2.0.0 (Dataflow v2)](https://github.com/Code-Inspect/flowr/releases/tag/v2.0.0))._

PR #799, opened 2024-05-18
Allow to check outputs of tests

EagleoutIce:
See `Slice for initial x should return noting else` in `calls-tests` for an example.


EagleoutIce:
_This pull request is included in v2.0.2 (see [Release v2.0.2 (Fixing Control-Flow, Markdown-Exports, and Handling of Unnamed Closures)](https://github.com/Code-Inspect/flowr/releases/tag/v2.0.2))._

PR #808, opened 2024-05-20
Improve Mermaid Escaping

EagleoutIce:
Now, we use the html shortcodes to avoid escaping problems.


EagleoutIce:
Currently, Oparu is down, which explains the failing tests.

PR #830, opened 2024-05-29
Include times per line in statistics

Ellpeck:


Ellpeck:
TODO: standard deviation for 1 run is 0 because it gets summarized from just one run, and then the standard deviation for the total is also 0 because it gets calculated based on the other standard deviations. that's wrong & bad

EagleoutIce:
_This pull request is included in v2.0.3 (see [Release v2.0.3 (Fine-Grained Benchmarks)](https://github.com/Code-Inspect/flowr/releases/tag/v2.0.3))._

EagleoutIce:
_This pull request is included in v2.0.3 (see [Release v2.0.3 (Fine-Grained Benchmarks)](https://github.com/Code-Inspect/flowr/releases/tag/v2.0.3))._

PR #897, opened 2024-08-19
 Server does not accept IDs generated by a multi-file-request

LukasPietzschmann:


src/cli/repl/server/messages/slice.ts
```diff
@@ -23,7 +23,7 @@ export const requestSliceMessage: MessageDefinition<SliceRequestMessage> = {
 		type:      Joi.string().valid('request-slice').required(),
 		id:        Joi.string().optional(),
 		filetoken: Joi.string().required(),
-		criterion: Joi.array().items(Joi.string().regex(/\d+:\d+|\d+@.*|\$\d+/)).min(0).required()
+		criterion: Joi.array().items(Joi.string().regex(/\d+:\d+|\d+@.*|\$(\d+|.+@root-\d+-\d+)/)).min(0).required()
```
LukasPietzschmann, line 26: Not sure if we should narrow down the path of the ID. E.g. `/a/b/c.R` in `/a/b/c.R|root-1-1`. Currently, it is matched by `.+`. But we could also be a bit more specific as it could be a real path, or just "-inline-"

EagleoutIce:
I get to it soon. Is this urgent? On that note: id Format validation in general may be too early here, given that it is combined with a semantic validation (existence) later and our only definition of validity is that the id exists. So maybe purging the check would be "cleaner"

LukasPietzschmann:
> Is this urgent?

Nope definitely not. Take your time!

EagleoutIce:
Reconsidering, I would prefer removing the regex check from the joi Definition. Do you want to do this @LukasPietzschmann? Otherwise I probably have some time tomorrow

LukasPietzschmann:
Yeah, I can do this later today as part of this PR.

EagleoutIce:
Thank you!

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.0.15 (see [Release v2.0.15 (Magic Comments and Subgraph Checks)](https://github.com/flowr-analysis/flowr/releases/tag/v2.0.15))._

PR #965, opened 2024-09-10
 Expose Multi-File AST

LukasPietzschmann:


LukasPietzschmann:
@EagleoutIce if you've got time, you can take a look at the code. There is still a TODO I have to implement, but the rest should reviewable :)

EagleoutIce:
sadly i do not have a lot of time at the moment, but looking at it the first time, suggest that it may be the `as` cast which wrongly tries to handle the striing aray as a strin

LukasPietzschmann:
I rebased this branch onto main ':)

EagleoutIce:
I pushed a respective fix which causes the parsing to happen sequentially! However, I still noticed that any additional file is normalized an additional time; i.e. there is another normalize call for this. Can you track that down? it is probably due to someone somewhere still not assuming the multifile ast exists (probably my old code) and just triggering a new parsing

EagleoutIce:
If you are ready, i will take on the merge regarding the 3-version upgrades we have

EagleoutIce:
Additionally, the case you established should probably be added as a test case to avoid regression.

EagleoutIce:
This is to be reconsidered alongside #1143.

PR #1046, opened 2024-10-08
[Query API] Location Query

LukasPietzschmann:


EagleoutIce:
Should i defer review until this is marked as ready?

LukasPietzschmann:
Yes please, the grouping is still completely missing.

LukasPietzschmann:
@EagleoutIce What do you think about adding a flag which, when set, will make the query return the `fullRange` of a node (if present)?

EagleoutIce:
But if this is too much, outsource this to another issue and we focus on what you require here.

src/queries/catalog/location-query/location-query-executor.ts
```diff
@@ -0,0 +1,17 @@
+import type { BasicQueryData } from '../../query';
+import type { LocationQuery, LocationQueryResult } from './location-query-format';
+import { labelSourceRange } from './location-query-format';
+
+
+export function executeLocationQuery({ ast }: BasicQueryData, queries: readonly LocationQuery[]): LocationQueryResult {
+	return {
+		'.meta': {
+			/* there is no sense in measuring a get */
+			timing: 0
```
EagleoutIce, line 10: given there are multiple hashmap access operations, i would in-fact measure the time, as we are no longer constant!

EagleoutIce:
We currently defer this based on the [location map query](https://github.com/flowr-analysis/flowr/wiki/Query-API#location-map-query) that got merged in the meantime!

PR #1065, opened 2024-10-11
Clean up query definition registration

Ellpeck:


src/queries/query.ts
```diff
@@ -14,6 +15,14 @@ import { executeIdMapQuery } from './catalog/id-map-query/id-map-query-executor'
...
+import { bold, type OutputFormatter } from '../util/ansi';
+import { printAsMs } from '../util/time';
+import { asciiCallContext, summarizeIdsIfTooLong } from '../cli/repl/commands/repl-query';
+import type { PipelineOutput } from '../core/steps/pipeline/pipeline';
+import type { DEFAULT_DATAFLOW_PIPELINE } from '../core/steps/pipeline/default-pipelines';
+import { graphToMermaidUrl } from '../util/mermaid/dfg';
+import { normalizedAstToMermaidUrl } from '../util/mermaid/ast';
+import Joi from 'joi';
 
 export type Query = CallContextQuery | DataflowQuery | NormalizedAstQuery | IdMapQuery;
```
EagleoutIce, line 27: Cold we please move this type down so it can be inferred from Supported Queries and not the other way around?

src/queries/query.ts
```diff
@@ -28,26 +37,86 @@ export interface BasicQueryData {
 export type QueryExecutor<Query extends BaseQueryFormat, Result extends BaseQueryResult> = (data: BasicQueryData, query: readonly Query[]) => Result;
 
 type SupportedQueries = {
-	[QueryType in Query['type']]: QueryExecutor<QueryArgumentsWithType<QueryType>, BaseQueryResult>
+	[QueryType in Query['type']]: SupportedQuery<QueryType>
+}
```
EagleoutIce, line 41: this should probably be just removed now, types are to be implcitly given by the single map below.

src/queries/query.ts
```diff
@@ -28,26 +37,86 @@ export interface BasicQueryData {
...
+	asciiSummarizer: (formatter: OutputFormatter, processed: PipelineOutput<typeof DEFAULT_DATAFLOW_PIPELINE>, queryResults: BaseQueryResult, resultStrings: string[]) => boolean
+	schema:          Joi.ObjectSchema
 }
 
 export const SupportedQueries = {
-	'call-context':   executeCallContextQueries,
-	'dataflow':       executeDataflowQuery,
-	'id-map':         executeIdMapQuery,
-	'normalized-ast': executeNormalizedAstQuery
+	'call-context': {
```
EagleoutIce, line 50: ok this is no half the way i wanted it to be :sweat_smile: could we move these definitions away from here so that the overview is nice and readable? 
For example:

`'call-context': CallContextQueryDefinition`

EagleoutIce (changes requested):

EagleoutIce:
_This pull request is included in v2.1.4 (see [Release v2.1.4 (Dependencies Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.1.4))._

PR #1121, opened 2024-11-04
[Abstract Domains] Validate the correctness of resolvesToBuiltInConstant

by-lars:


test/functionality/dataflow/environments/resolve-tests.ts
```diff
@@ -42,4 +44,28 @@ describe('Resolve', () => {
 			});
 		});
 	});
+	describe('Builtin Constants', () => {
+		it('Unknown Identifier', () => {
+			const env = defaultEnv();
+			const result = resolvesToBuiltInConstant(undefined, env, undefined);
+			expect(result, 'should be Ternary.Never').to.be.equal(Ternary.Never);
```
EagleoutIce, line 51: So in general we would prefer to compact this here. (e.g., avoid the `env` variable).
Besides, we only rarely use BDD here and usually use the normal assertion style.

test/functionality/dataflow/environments/resolve-tests.ts
```diff
@@ -42,4 +44,28 @@ describe('Resolve', () => {
...
+		});
+		it('Undefined Identifier', () => {
+			const env = defaultEnv();
+			const result = resolvesToBuiltInConstant('foo', env, undefined);
+			expect(result, 'should be Ternary.Never').to.be.equal(Ternary.Never);
+		});
+
+		const testSingle = (identifier: Identifier, expected: unknown) => function() {
```
EagleoutIce, line 59: Usually we use such a style to include the `it` right away, just so we can theoretically abstract this into many `it`s later :D So could you refactor this into something like:

```ts
function test(label: string, identifier: Identifier, expected: unknown) {
  // ...
}
```

test/functionality/dataflow/environments/resolve-tests.ts
```diff
@@ -42,4 +44,28 @@ describe('Resolve', () => {
...
+			const env = defaultEnv();
+			const result = resolvesToBuiltInConstant(identifier, env, expected);
+			expect(result, 'should be Ternary.Always').to.be.equal(Ternary.Always);
+		};
+		it('Resolve TRUE',	testSingle('TRUE', true));
+		it('Resolve T',		testSingle('T', true));
+		it('Resolve FALSE',	testSingle('FALSE', false));
+		it('Resolve F',		testSingle('F', false));
+		it('Resolve NULL',	testSingle('NULL', null));
+		it('Resolve NA',	testSingle('NA', null));
```
EagleoutIce, line 69: Currently missing are negative tests although they may not be as interesting. Yet something like "this does not resolve to anything", "this does not resolve to what i want", or this "may resolve to what we want" - may be of interest.

test/functionality/dataflow/environments/resolve-tests.ts
by-lars: I used the style that was used in the file. How would the preferred style look like?

test/functionality/dataflow/environments/resolve-tests.ts
by-lars: I now have tests for an Identifier which is not defined and an Identifier that does not resolve to what we wanted.
But I am not quite sure how to implement the test case for "maybe resolves to what we want".

test/functionality/dataflow/environments/resolve-tests.ts
EagleoutIce: `assert.strictEqual` etc :D

test/functionality/dataflow/environments/resolve-tests.ts
```diff
@@ -42,4 +45,33 @@ describe('Resolve', () => {
...
+		testResolve('Resolve T',		    'T', 		true,		  Ternary.Always);
+		testResolve('Resolve FALSE',	    'FALSE',	false,		  Ternary.Always);
+		testResolve('Resolve F',		    'F',		false,		  Ternary.Always);
+		testResolve('Resolve NULL',		    'NULL', 	null,		  Ternary.Always);
+		testResolve('Resolve NA',		    'NA', 		null,		  Ternary.Always);
+
+		testResolve('Maybe Resolves TRUE',  'TRUE', 	true,		  Ternary.Maybe, 
+			defaultEnv().defineInEnv({ name: 'TRUE', nodeId: 0, definedAt: 1, type: ReferenceType.Constant, controlDependencies: [{ id: 42, when: true }] })
+		);
+		testResolve('Maybe Resolves FALSE', 'FALSE', 	false,		  Ternary.Maybe,
```
EagleoutIce, line 66: From a technical perspective this is probably redundant, aber let's keept it this way :)

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.1.6 (see [Release v2.1.6 (Include Source-File Information, Vitest, and Documentation)](https://github.com/flowr-analysis/flowr/releases/tag/v2.1.6))._

PR #1176, opened 2024-11-25
[Tests] Basic System Tests

gigalasr:


gigalasr:
@EagleoutIce How should the system tests be executed in the ci? SHould there be an extra step or should they be executed with test-full (change package.json) ?

EagleoutIce:
> @EagleoutIce Would you like me to add any of the examples you use often that you mentioned? If yes could you provide them?

For now it is fine to test something like `:df x <- 3` and `:df "x <- 3\nprint(x)"`. I can do the rest.

EagleoutIce:
> @EagleoutIce How should the system tests be executed in the ci? SHould there be an extra step or should they be executed with test-full (change package.json) ?

A separate step would be fine to have `test:system` or something like that. This should be not part of `test-full` but part of `checkup`

EagleoutIce:
_This pull request is included in v2.1.9 (see [Release v2.1.9 (Improved Closure Slicing, DFG docs, and Support for the Car Package)](https://github.com/flowr-analysis/flowr/releases/tag/v2.1.9))._

Issue #1135, opened 2024-11-13
Support Aliasses in Dependencies Query


EagleoutIce:
To identify, e.g., the libraries loaded (after #1106).

EagleoutIce:
This is now in #1258.

PR #1201, opened 2024-12-05
[Abstract Domains iii]: Resolve constant aliases

gigalasr:


src/dataflow/environments/resolve-by-name.ts
```diff
@@ -105,3 +110,88 @@ export function resolveToConstants(name: Identifier | undefined, environment: RE
...
+export function getAliases(sourceIds: readonly NodeId[], dataflow: DataflowGraph, environment: REnvironmentInformation): NodeId[] | undefined {
+	const definitions: NodeId[] = [];
+
+	for(const sourceId of sourceIds) {
+		const info = dataflow.getVertex(sourceId);
+		if(info === undefined) {
+			return undefined; 
+		}
+
+		if(info.tag === VertexType.Value) {
```
EagleoutIce, line 124: maybe use an object so that we can write `aliasHandler[info.tag](...)`

src/dataflow/environments/resolve-by-name.ts
```diff
@@ -105,3 +110,88 @@ export function resolveToConstants(name: Identifier | undefined, environment: RE
 		from:  def.type
 	}));
 }
+
+
+export function getAliases(sourceIds: readonly NodeId[], dataflow: DataflowGraph, environment: REnvironmentInformation): NodeId[] | undefined {
```
EagleoutIce, line 115: In general it would be nice to have a config option whether to use getAlias or resolves to built in constant. Can you explore this within this PR? (or defer as next step after this issue?)

EagleoutIce (approved):

EagleoutIce:
Besides the two cases i sent to you, you can use `call-context-query.test.ts#Aliases` replacing the `print` call (or whatever it is we search for in the test) by a value such as `3` or `"hey"`.

EagleoutIce:
Maybe consider doing #1135 afterward c:

src/dataflow/environments/resolve-by-name.ts
```diff
@@ -105,3 +110,88 @@ export function resolveToConstants(name: Identifier | undefined, environment: RE
...
+	if(defs === undefined) {
+		return undefined;
+	}
+
+	const values: unknown[] = [];
+	for(const def of defs) {
+		if(def.type === ReferenceType.BuiltInConstant) {
+			values.push(def.value);
+		} else if(def.type === ReferenceType.BuiltInFunction) {
+			// TODO: nothing?
```
EagleoutIce, line 178: You can make an issue for this ~> track function aliasses

src/dataflow/environments/resolve-by-name.ts
EagleoutIce: This switch is probably best implemented by abstracting the call to resolve, ideally we add a flag to the config format which then configure an appropriate handler/resolve object on startup

EagleoutIce:
_This pull request is included in v2.1.10 (see [Release v2.1.10 (First support for pointer analysis (named list arguments with `$`))](https://github.com/flowr-analysis/flowr/releases/tag/v2.1.10))._

PR #1160, opened 2024-11-19
 [Code Search, II] A tree-sitter parser alternative

Ellpeck:

PR #1221, opened 2024-12-23
New Parser Backend (Tree-Sitter)

EagleoutIce:


Ellpeck:
Still left to do (carried over from #1160):

* [ ] switch to npm package once available, to avoid wasm and (hopefully `init`)
   * [ ] Still allow the wasm backend (e.g., if we want to develop/explore part of it) with
         a user-supplied `.wasm` file.
* [ ] Document the new backend and the way we can configure it (**Flo**, later)
* [ ] Also execute benchmarks with tree-sitter and r shell separately
* [ ] Fix the three-ish errors that occur with the tree sitter parser in the social-science benchmark suite

EagleoutIce:
_This pull request is included in v2.2.0 (see [Release v2.2.0 (Tree-Sitter Engine)](https://github.com/flowr-analysis/flowr/releases/tag/v2.2.0))._

PR #1231, opened 2025-01-06
Throw an error if the requested config file does not exist/is not found

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.1.11 (see [Release v2.1.11 (Improved Configuration Retrieve and Config Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.1.11))._

PR #1267, opened 2025-01-28
[Abstract Domains IV] A resolve value query

gigalasr:


src/documentation/print-query-wiki.ts
```diff
@@ -221,6 +221,28 @@ ${
 	}
 });
 
+registerQueryDocumentation('resolve-value', {
+	name:             'Resolve Value Query',
+	type:             'active',
+	shortDescription: 'Provides access to flowR\'s alias tracking',
```
EagleoutIce, line 227:
```suggestion
	shortDescription: 'Provides access to flowR\'s value tracking (which is configurable)',
```

src/documentation/print-query-wiki.ts
```diff
@@ -221,6 +221,28 @@ ${
...
+	functionName:     executeSearch.name,
+	functionFile:     '../queries/catalog/resolve-value-query/resolve-value-query-executor.ts',
+	buildExplanation: async(shell: RShell) => {
+		const exampleCode = 'x <- 1\nprint(x)';
+		return `
+With this query you can use flowR's alias-tracking capabilities to resolve identifiers to their values.
```
EagleoutIce, line 233:
```suggestion
With this query you can use flowR's value-tracking capabilities to resolve identifiers to all potential values they may have at runtime (if possible). The extend to which flowR traces values (e.g. built-ins vs. constants) can be configured in flowR's Configuration file (see the [Interface](${FlowrWikiBaseRef}/Interface) wiki page for more information). 
```

src/queries/catalog/resolve-value-query/resolve-value-query-executor.ts
```diff
@@ -0,0 +1,43 @@
...
+		throw new Error('idMap was undefined');
+	}
+
+	const start = Date.now();
+	const results: ResolveValueQueryResult['results'] = {};
+	for(const query of queries) {
+		const key = fingerPrintOfQuery(query);
+		
+		if(results[key]) {
+			log.warn(`Duplicate Key for slicing-query: ${key}, skipping...`);
```
EagleoutIce, line 25:
```suggestion
			log.warn(`Duplicate Key for resolve-value-query: ${key}, skipping...`);
```

src/queries/catalog/resolve-value-query/resolve-value-query-executor.ts
```diff
@@ -0,0 +1,43 @@
...
+export function fingerPrintOfQuery(query: ResolveValueQuery): string {
+	return JSON.stringify(query);
+}
+
+export function executeResolveValueQuery({ dataflow: { graph, environment } }: BasicQueryData, queries: readonly ResolveValueQuery[]): ResolveValueQueryResult {
+	const idMap = graph.idMap;
+
+	if(!idMap) {
+		throw new Error('idMap was undefined');
+	}
```
EagleoutIce, line 17: Please use flowR's `guard` function for this, to ensure that the idMap is not undefined.
In general, you can always fallback to the idMap given in then normalized AST (as part of the BasicQueryData). So we should do this first.

src/queries/catalog/resolve-value-query/resolve-value-query-format.ts
```diff
@@ -0,0 +1,58 @@
...
+	results: Record<string, {values: unknown[]}>
+}
+
+function rValueToAscii(value: string | RNumberValue | RStringValue): string {
+	if(value === null || value === undefined) {
+		return 'undefined';
+	} else if(typeof value === 'string') {
+		return value;
+	} else if(typeof value === 'object') {	
+		if('num' in value) {
```
EagleoutIce, line 29: that is one way to do it :D

test/functionality/dataflow/query/resolve-value-query.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+				results[key] = {
+					values: [... new Set(values)]
+				};
+			}
+
+			return {
+				'resolve-value': { results }
+			};
+		});
```
EagleoutIce, line 36: i do not think this is a good way to test the query - simply by clone and own. I think it would be better if we explicitly specify the expected result here (i.e., accept another parameter to `testQuery` to specify the expected result)

EagleoutIce (changes requested):

test/functionality/dataflow/query/resolve-value-query.test.ts
gigalasr: Oki, I did it this way because the other tests also followed a similar approach. I will change the code to explicitly allowed to provide the expected result.

src/queries/catalog/resolve-value-query/resolve-value-query-format.ts
gigalasr: Meaning, I should do it differently, or is this ok?

test/functionality/dataflow/query/resolve-value-query.test.ts
EagleoutIce: jeah but this was originally only meant to be the case for simple getter queries

src/queries/catalog/resolve-value-query/resolve-value-query-format.ts
```diff
@@ -0,0 +1,58 @@
...
+		for(const [fingerprint, obj] of Object.entries(out.results)) {
+			const { criteria } = JSON.parse(fingerprint) as ResolveValueQuery;
+			result.push(`   ╰ Values for {${criteria.join(', ')}}`);
+			result.push(`   	╰ ${obj.values.map(v => rValueToAscii(v as string | RNumberValue | RStringValue)).join(', ')}`);
+		}
+		return true;
+	},
+	schema: Joi.object({
+		type:     Joi.string().valid('resolve-value').required().description('The type of the query.'),
+		criteria: Joi.array().items(Joi.string()).min(0).required().description('The slicing criteria to use.'),
```
EagleoutIce, line 56:
```suggestion
		criteria: Joi.array().items(Joi.string()).min(1).required().description('The slicing criteria to use.'),
```

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.2.2 (see [Release v2.2.2 (Robustify tree-sitter Integration, Resolve-Value Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.2.2))._

EagleoutIce:
_This pull request is included in v2.2.4 (see [Release v2.2.4 (Improved Docs and Dependency Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.2.4))._

PR #1400, opened 2025-02-28
Added pointer analysis data to benchmark stats

### LINKED ISSUE #1398: Add pointer analysis data to benchmark stats
**Is your feature request related to a problem? Please describe.**
The newly developed pointer analysis stores indices for pointers in the data flow graph. Now, we want to include the number of stored indices in the benchmark statistics so that we can compare the impact of the pointer analysis.

**Describe the solution you would like**
Count the number of stored pointer indices and stored it in the benchmark statistics.

**Additional context**

Slartibartfass2:
Closes #1398


EagleoutIce (approved):
Great! Could you please just add a single test case (+-) to verify that this counts indices for a case in which they exist? If you want to, adapt an existing one :)

Slartibartfass2:
> Great! Could you please just add a single test case (+-) to verify that this counts indices for a case in which they exist? If you want to, adapt an existing one :)

Sorry, I wasn't finished yet. Should've set in on draft. I'm currently adding a test.

EagleoutIce:
Ok, feel free to ping me whenever this is ready @Slartibartfass2 :)

Slartibartfass2:
@EagleoutIce Can you please take a look again.

src/benchmark/slicer.ts
```diff
@@ -112,7 +115,7 @@ export class BenchmarkSlicer {
 	 * Initialize the slicer on the given request.
 	 * Can only be called once for each instance.
 	 */
-	public async init(request: RParseRequestFromFile | RParseRequestFromText, autoSelectIf?: AutoSelectPredicate) {
+	public async init(request: RParseRequestFromFile | RParseRequestFromText, autoSelectIf?: AutoSelectPredicate, threshold?: number) {
```
EagleoutIce, line 118: Since then this can be configured in the configuration file from flowR, but i am fine with this!

src/benchmark/slicer.ts
```diff
@@ -329,24 +387,28 @@ export class BenchmarkSlicer {
...
 		report: (current: number, total: number, allCriteria: SlicingCriteria[]) => void = () => { /* do nothing */ },
-		sampleRandom = -1
+		sampleRandom = -1,
+		maxSliceCount = -1,
 	): Promise<number> {
 		this.guardActive();
 		let count = 0;
 		const allCriteria = [...collectAllSlicingCriteria((this.normalizedAst as NormalizedAst).ast, filter)];
+		if(maxSliceCount > 0 && allCriteria.length > maxSliceCount) {
+			return -allCriteria.length;
```
EagleoutIce, line 410: Shouldn't this just sample in that case? as it hapens with sampleRandom below?
A hard cutoff sounds like something which introduces a lot of bias

Slartibartfass2: I thought this was the idea, we discussed. To skip files with too many slices, we can use this flag.
Or did I get this wrong?

src/benchmark/slicer.ts
Slartibartfass2: I can also just use the config. What do you prefer?

src/benchmark/slicer.ts
EagleoutIce: Oh okay, I was under the impression we would just sample in these cases as this is equivalent to "we ignore them". I am fine with either as long as it is documented.

Slartibartfass2: The CLI argument is documented, and I added a comment in the code. Do we need documentation somewhere else I've missed?

EagleoutIce (approved):

EagleoutIce:

PR #1416, opened 2025-03-10
Bug-Fix: Processing of nested Arguments

EagleoutIce:
Before PA this should have been fixable just by allowing recursive resolves of `$<-` and I think this was simply lost when rewriting the built-in environment to use the json config format. Now, with pointer analysis, we require much more sparkles.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -164,7 +164,7 @@ export const defaultConfigOptions: FlowrConfigOptions = {
 	defaultEngine: 'r-shell',
 	solver:        {
 		variables:       VariableResolve.Alias,
-		pointerTracking: false,
+		pointerTracking: true,
 		resolveSource:   {
 			dropPaths:             DropPathsOption.No,
 			ignoreCapitalization:  true,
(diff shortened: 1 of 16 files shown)


src/config.ts
```diff
@@ -164,7 +164,7 @@ export const defaultConfigOptions: FlowrConfigOptions = {
 	defaultEngine: 'r-shell',
 	solver:        {
 		variables:       VariableResolve.Alias,
-		pointerTracking: false,
+		pointerTracking: true,
```
Slartibartfass2, line 167: Are we sure to enable it back again? I'm a bit afraid that it might still break something.

src/dataflow/internal/process/functions/call/built-in/built-in-replacement.ts
```diff
@@ -5,31 +5,40 @@ import { processKnownFunctionCall } from '../known-call-handling';
...
 import { dataflowLogger } from '../../../../../logger';
-import type { ContainerIndices, ContainerIndicesCollection, ContainerLeafIndex, IndexIdentifier } from '../../../../../graph/vertex';
-import { VertexType } from '../../../../../graph/vertex';
+import type {
+	ContainerIndices,
+	ContainerIndicesCollection,
+	ContainerLeafIndex,
+	IndexIdentifier } from '../../../../../graph/vertex';
+
+
```
Slartibartfass2, line 23: Are these blank lines wanted?

Slartibartfass2 (approved):
I only partially understand the fix, but I can confirm that the bug is fixed.

src/config.ts
EagleoutIce: Hm unsure, I was under the assumption that this is now the best way we have to find new bugs in the wild :D

EagleoutIce:

PR #1438, opened 2025-03-12
Fix: the pointer analysis scaling upsie

EagleoutIce:


src/dataflow/internal/process/functions/call/built-in/built-in-vector.ts
```diff
@@ -65,6 +65,12 @@ export function processVector<OtherInfo>(
 		}
 	}
 
+	const threshold = getPointerAnalysisThreshold();
+
+	if(threshold !== 'unlimited' && (threshold === 'disabled' || vectorArgs.length > threshold)) {
+		return fnCall.information;
+	}
```
Slartibartfass2, line 72: Maybe extract this check, it's both in `built-in-vector` and in `built-in-list`. If we add the support for more data types, we want to reuse it again.

Slartibartfass2 (changes requested):
Thank you very much for the fix!
Just some minor things.

Slartibartfass2 (approved):
Looks great now! I'm excited to run the eval with this again.

EagleoutIce:

PR #1439, opened 2025-03-13
Add equidistant selection for slicing criterias with the benchmark slicer

### LINKED ISSUE #1433: Benchmark: Replace random sampling with equidistant sampling
**Is your feature request related to a problem? Please describe.**

Currently, if we would rather not slice all available slices in the benchmark, we can use sampling, e.g., only slicing 10 available slices. This uses randomness to determine the sampled slicing criteria. For comparing two benchmarks, this is bad because the position of the slices has a high impact on the output. For instance, if in the first benchmark the first slice is sampled and in the second one the last slice is sampled, and the file is huge, the first slice cannot possibly contain as many lines as the second one.

TLDR; random slice sampling leads to results that are challenging to compare.

**Describe the solution you would like**

Instead of choosing random samples, use equidistant samples, e.g., for 10 samples, use every 10th slice. This way, it is deterministic and comparable with other benchmarks that use the same arguments.

**Additional context**
Random sampling with a seed would also be possible.

Slartibartfass2:
Closes #1433 

See issue for details.

### DIFF
--- a/src/benchmark/slicer.ts
+++ b/src/benchmark/slicer.ts
@@ -402,6 +402,7 @@ export class BenchmarkSlicer {
 		filter: SlicingCriteriaFilter,
 		report: (current: number, total: number, allCriteria: SlicingCriteria[]) => void = () => { /* do nothing */ },
 		sampleCount = -1,
+		sampleStrategy: 'random' | 'equidistant' = 'random',
 		maxSliceCount = -1,
 	): Promise<number> {
 		this.guardActive();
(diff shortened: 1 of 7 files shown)


src/benchmark/slicer.ts
```diff
@@ -400,19 +401,18 @@ export class BenchmarkSlicer {
 	public async sliceForAll(
 		filter: SlicingCriteriaFilter,
 		report: (current: number, total: number, allCriteria: SlicingCriteria[]) => void = () => { /* do nothing */ },
-		sampleRandom = -1,
```
EagleoutIce, line 403: hmm so I do not want this option to be gone... maybe we keep `sampleCount` but produce another key which indicates the `sampleStrategy` and just defaults to `random`? There are still flowR scenarios in which we want to sample randomly.

EagleoutIce (changes requested):
Besides retaining the random sampling, i am fine :)

src/benchmark/slicer.ts
Slartibartfass2: I brought it back.

Slartibartfass2 (commented):
I also disabled pointer tracking when the flag is not passed to the benchmark tool. Now that we made it default again, it would be enabled either way.

PR #1466, opened 2025-03-14
[AbsInt] Tracking Data-Frame Shapes

EagleoutIce:


src/abstract-interpretation/data-frame/abstract-interpretation.ts
```diff
@@ -0,0 +1,158 @@
...
+	node: RNode<ParentInformation>
+): node is RConstant<ParentInformation> {
+	return node.type === RType.String || node.type === RType.Number || node.type === RType.Logical;
+}
+
+function isRSingleNode(
+	node: RNode<ParentInformation>
+): node is RSingleNode<ParentInformation> {
+	return isRConstant(node) || node.type === RType.Symbol || node.type === RType.Break || node.type === RType.Next || node.type === RType.Comment || node.type === RType.LineDirective;
+}
```
EagleoutIce, line 106: I think we should move these functions to a more general position - maybe alongside the model?

OliverGerstl: These functions do no longer exist, as we can now directly check whether a CFG vertex represents a simple or complex AST node by checking whether the vertex has an end marker

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
+import type { RNode } from '../../r-bridge/lang-4.x/ast/model/model';
+import type { ParentInformation } from '../../r-bridge/lang-4.x/ast/model/processing/decorate';
+import type { NodeId } from '../../r-bridge/lang-4.x/ast/model/processing/node-id';
+import type { DataFrameStateDomain } from './domain';
+import type { ConstraintType, DataFrameOperationArgs, DataFrameOperationName, DataFrameOperationOptions } from './semantics';
+
+/**
+ * An abstract data frame operation without additional options.
+ * - `operation` contains the type of the abstract operations (see {@link DataFrameOperationName})
```
EagleoutIce:
```suggestion
 * - `operation` contains the type of the abstract operation (see {@link DataFrameOperationName})
```

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+    [Name in OperationName]: {
+		operation: Name,
+		operand:   NodeId | undefined,
+		type?:     ConstraintType,
+		options?:  DataFrameOperationOptions<Name>
+	} & DataFrameOperationArgs<Name>;
+}[OperationName];
+
+interface DataFrameInfoBase {
+	type?:   string,
```
EagleoutIce, line 44: I think we should document this as well - given that the `type` is optional this can also happen as a union afterward.

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/** Enum to mark nodes during the data frame shape inference */
+export enum DataFrameInfoMarker {
+	/** Marks the target symbol of assignments as "unassigned" until the assigned expression is evaluated */
+	Unassigned = 'unassigned'
+}
+
+/**
+ * Represents a node without extra data frame information.
```
EagleoutIce: Please clarify whether this is "any other node" or a node for which we think it may be a data frame/only know that it is one.

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+ * The `marker` can be used to mark nodes during the data frame shape inference
+ */
+interface DataFrameEmptyInfo extends DataFrameInfoBase {
+	type?:   never,
+	marker?: DataFrameInfoMarker
+}
+
+/**
+ * Represents a symbol assignment with a target identifier (symbol) and assigned expression.
```
EagleoutIce: *symbol* assignment? Given that you dispatch over the CFG this should include, e.g. `"a" <- 42`

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export interface DataFrameAssignmentInfo extends DataFrameInfoBase {
+	type:       'assignment',
+	identifier: NodeId,
+	expression: NodeId
+}
+
+/**
+ * Represents a data frame expression with mapped abstract operations.
```
EagleoutIce: is there any comment to be made wrt. to side effects that this expression may or may not have

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export interface DataFrameExpressionInfo extends DataFrameInfoBase {
+	type:       'expression',
+	operations: DataFrameOperation[]
```
EagleoutIce, line 88: Are these operations expected to be in sequence, confluent, any order, ...?

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export interface AbstractInterpretationInfo {
+	dataFrame?: DataFrameInfo
+}
+
+/**
+ * Checks whether an AST node has attached data frame assignment information.
+ */
+export function hasDataFrameAssignmentInfo(
+	node: RNode<ParentInformation & AbstractInterpretationInfo>
```
EagleoutIce: In general we pass along `OtherInfo` for these functions to allow for additional decorations (newer TS versions are better in handling those but in general we want to state that "this does not remove this `OtherInfo`"

src/abstract-interpretation/data-frame/absint-info.ts
EagleoutIce: Additionally, maybe we can add the checks like `hasDataFrameAssignmentInfo`?

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+ * The control flow graph visitor to infer the shape of data frames using abstract interpretation
+ */
+export class DataFrameShapeInferenceVisitor<
+	OtherInfo = NoInfo,
+	ControlFlow extends ControlFlowInformation = ControlFlowInformation,
+	Ast extends NormalizedAst<OtherInfo & AbstractInterpretationInfo> = NormalizedAst<OtherInfo & AbstractInterpretationInfo>,
+	Dfg extends DataflowGraph = DataflowGraph,
+	Config extends DataFrameShapeInferenceVisitorConfiguration<OtherInfo, ControlFlow, Ast, Dfg> = DataFrameShapeInferenceVisitorConfiguration<OtherInfo, ControlFlow, Ast, Dfg>
+> extends SemanticCfgGuidedVisitor<OtherInfo & AbstractInterpretationInfo, ControlFlow, Ast, Dfg, Config & { defaultVisitingOrder: 'forward', defaultVisitingType: 'exit' }> {
+	/** The old domain of an AST node before proccessing the node */
```
EagleoutIce: Maybe add why we do require the "pre" state

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+
+		const visitedCount = this.visited.get(vertex.id) ?? 0;
+		this.visited.set(vertex.id, visitedCount + 1);
+
+		// only continue visitor if the node has not been visited before or the data frane value of the node changed
+		return visitedCount === 0 || !equalDataFrameState(this.oldDomain, this.newDomain);
+	}
+
+	protected override visitDataflowNode(vertex: Exclude<CfgSimpleVertex, CfgBasicBlockVertex>): void {
+		const node = this.getNormalizedAst(isMarkerVertex(vertex) ? vertex.root : vertex.id);
```
EagleoutIce: Maybe outsource the id retrieval of a `vertex` to make these and other requests more readable?

src/abstract-interpretation/data-frame/absint-visitor.ts
EagleoutIce: additionally, make clear to what extend this maps global state (after visiting the root vertex)

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+			return;
+		}
+		this.oldDomain = node.info.dataFrame?.domain ?? new Map<NodeId, DataFrameDomain>();
+		super.visitDataflowNode(vertex);
+
+		if(this.shouldWiden(vertex)) {
+			this.newDomain = wideningDataFrameStates(this.oldDomain, this.newDomain);
+		}
+		// mark variable definitions as "unassigned", as the evaluation of the assigned expression is delayed until processing the assignment
+		const variableDefinition = isVariableDefinitionVertex(this.getDataflowGraph(vertex.id));
```
EagleoutIce: maybe `isVariableDef` to make sure it is a flag

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+		node.info.dataFrame ??= variableDefinition ? { marker: DataFrameInfoMarker.Unassigned } : {};
```
EagleoutIce: couldn't this happen in the variableDefinition handler of the dataflow processor? given that the domain is within the attribute and the visitor should be called by the super call anyway?

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+		node.info.dataFrame.domain = this.newDomain;
+	}
+
+	protected override onAssignmentCall({ call, target, source }: { call: DataflowGraphVertexFunctionCall, target?: NodeId, source?: NodeId }): void {
+		const node = this.getNormalizedAst(call.id);
+		const targetNode = target !== undefined ? this.getNormalizedAst(target) : undefined;
```
EagleoutIce: Reading this i would be happy to move the `undefined` case into the `getNormalizedAst`

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+		let value = DataFrameTop;
+
+		for(const { operation, operand, type, options, ...args } of node.info.dataFrame.operations) {
+			const operandValue = operand !== undefined ? resolveIdToDataFrameShape(operand, this.config.dfg, this.newDomain) : value;
+			value = applySemantics(operation, operandValue ?? DataFrameTop, args, options);
+			const constraintType = type ?? getConstraintType(operation);
+
+			if(operand !== undefined && constraintType === ConstraintType.OperandModification) {
+				this.newDomain.set(operand, value);
+				getVariableOrigins(operand, this.config.dfg).forEach(origin => this.newDomain.set(origin.info.id, value));
```
EagleoutIce: we shouldn't use `.forEach` - it tends to be much much slower in certain scenarios

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+			} else if(constraintType === ConstraintType.ResultPostcondition) {
+				this.newDomain.set(node.info.id, value);
+			}
+		}
+	}
+
+	// We only process vertices of leaf nodes and exit vertices (no entry nodes of complex nodes)
```
EagleoutIce: Why is this not a tsdoc comment?

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+	private shouldSkipVertex(vertex: CfgSimpleVertex) {
+		return isMarkerVertex(vertex) ? vertex.type !== CfgVertexType.EndMarker : vertex.end !== undefined;
+	}
+
+	private getPredecessorNodes(vertexId: NodeId): RNode<ParentInformation & AbstractInterpretationInfo>[] {
+		return this.config.controlFlow.graph.outgoingEdges(vertexId)?.keys()  // outgoing dependency edges are incoming CFG edges
```
EagleoutIce, line 187: cursed :P

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+				if(vertex !== undefined && this.shouldSkipVertex(vertex)) {
+					return this.getPredecessorNodes(vertex.id);
+				} else if(vertex?.type === CfgVertexType.EndMarker) {
+					const nodeId = vertex.root;
+					return nodeId ? [this.getNormalizedAst(nodeId)] : [];
+				} else {
+					return vertex ? [this.getNormalizedAst(vertex.id)] : [];
+				}
+			})
+			.filter(node => node !== undefined)
```
EagleoutIce: Please use the `isNotUndefined` helper here so the type system knows what is going on

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -0,0 +1,199 @@
...
+			.toArray() ?? [];
```
EagleoutIce, line 199: why exactly `toArray` ?

src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -0,0 +1,324 @@
+import { defaultConfigOptions } from '../../config';
+import type { NodeId } from '../../r-bridge/lang-4.x/ast/model/processing/node-id';
+import { setEquals } from '../../util/collections/set';
+
+const MaxColNames = defaultConfigOptions.abstractInterpretation.dataFrame.maxColNames;
+
+type Interval = [number, number];
+
+/** The bottom element (least element) of the positive interval domain representing no possible values, explicitly given as "bottom". */
+export const IntervalBottom = Symbol('bottom');
```
EagleoutIce: Using a Symbol for this technically breaks reusability over several runs/storing this bottom in JSON but if this is fine with yo

src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -0,0 +1,324 @@
...
+/** The top element (greatest element) of the positive interval domain representing all possible values, defined as the interval from 0 to infinity. */
+export const IntervalTop: Interval = [0, Infinity];
+/** The positive interval domain representing possible integer values. */
+export type IntervalDomain = Interval | typeof IntervalBottom;
+
+/** The bottom element (least element) of the column names domain representing no possible column name, defined as the empty list []. */
+export const ColNamesBottom: string[] = [];
```
EagleoutIce: this way, technically, the type system looses the information:
1) that this should not be changed (it is not readonly)
2) that it is empty (you say it is `string[]`).

shouldnt this be better?

```suggestion
export const ColNamesBottom = [] as const;
```

src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -0,0 +1,324 @@
...
+    colnames: ColNamesDomain,
+    cols:     IntervalDomain,
+    rows:     IntervalDomain
+}
+
+/**
+ * The bottom element (least element) of the data frame shape domain representing no possible value, mapping the columns names to {@link ColNamesBottom}
+ * and the number of columns and rows to {@link IntervalBottom}.
+ */
+export const DataFrameBottom: DataFrameDomain = {
```
EagleoutIce: again we could use `as const satisfies...` to make it more informative for the type system

src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/** Subtracts an abstract value from another abstract value of the column names domain by performing a set minus. */
+export function subtractColNames(set1: ColNamesDomain, set2: ColNamesDomain): ColNamesDomain {
+	if(set1 === ColNamesTop) {
+		return ColNamesTop;
+	} else if(set2 === ColNamesTop) {
+		return set1;
+	} else {
+		return Array.from(new Set(set1).difference(new Set(set2)));
```
EagleoutIce, line 96: if column names would imply order, this would destroy that. If they do not, why use arrays in the first place?

src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -0,0 +1,188 @@
...
+					colnames:  columns?.map(col => typeof col === 'string' ? col : undefined)
+				});
+			}
+			operand = undefined;
+		}
+	}
+	return result;
+}
+
+function getEffectiveArgs(
```
EagleoutIce: Maybe document that this returns all non special, named arguments?

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/**
+ * Gets all effective argument from a list of arguments by removing all arguments whose names should be excluded.
+ *
+ * @param args     - The list of arguments to filter
+ * @param excluded - The names of the arguments to exclude
+ * @returns The filtered list of arguments
+ */
+export function getEffectiveArgs(
```
EagleoutIce, line 101: given that this name is also used at a another position where it does additionally consider special operators, I would prefer one of those being renamed.

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,270 @@
...
+export function getUnresolvedSymbolsInExpression(
+	expression: RNode<ParentInformation> | typeof EmptyArgument | undefined,
+	dfg?: DataflowGraph
+): string[] {
+	if(expression === undefined || expression === EmptyArgument || dfg === undefined) {
+		return [];
+	}
+	switch(expression.type) {
+		case RType.ExpressionList:
+			return [...expression.children.flatMap(child => getUnresolvedSymbolsInExpression(child, dfg))];
```
EagleoutIce: why re-spread? this should just be a performance hazard :see_no_evil:

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,270 @@
...
+		case RType.Argument:
+			return [...getUnresolvedSymbolsInExpression(expression.value, dfg)];
+		case RType.Symbol:
+			if(isUseVertex(dfg.getVertex(expression.info.id)) && (dfg.outgoingEdges(expression.info.id)?.size ?? 0) === 0) {
+				return [unquoteArgument(expression.content)];
+			} else {
+				return [];
+			}
+		default:
+			return [];
```
EagleoutIce: Is this the reason you do not use one of the visitors? In general I would prefer you use the visitor as you may miss some children etc. when we change the AST structure in the fturue

src/abstract-interpretation/data-frame/mappers/assignment-mapper.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @returns Data frame assignment info containing the IDs of the identifier and expression, or `undefined` if the node does not represent a data frame assignment
+ */
+export function mapDataFrameVariableAssignment(
+	identifier: RSymbol<ParentInformation> | RString<ParentInformation>,
+	expression: RNode<ParentInformation>,
+	dfg: DataflowGraph
+): DataFrameAssignmentInfo | undefined {
+	const resolveInfo = { graph: dfg, idMap: dfg.idMap, full: true, resolve: VariableResolve.Alias };
+
+	if(!isDataFrameArgument(expression, resolveInfo)) {
```
EagleoutIce, line 27: So this way we die in functions? this is fine but maybe we should document this

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+	Tibble = 'tibble',
+	DataTable = 'data.table'
+}
+
+/**
+ * Mapper for mapping the supported concrete data frame functions to mapper functions,
+ * including information about the origin library of the functions and the type of the returned data frame.
+ */
+const DataFrameFunctionMapper = {
+	'data.frame':    { mapper: mapDataFrameCreate },
```
EagleoutIce: maybe make the return type explicit for all of them and add the library for a base version too?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+	'full_join':     { mapper: mapDataFrameJoin, library: 'dplyr' },
+	'merge':         { mapper: mapDataFrameJoin },
+	'relocate':      { mapper: mapDataFrameIdentity, library: 'dplyr' },
+	'arrange':       { mapper: mapDataFrameIdentity, library: 'dplyr' }
+} as const satisfies Record<string, DataFrameFunctionMapperInfo<never>>;
+
+/**
+ * List of data frame functions that are not explicitly supported but may return data frames.
+ */
+const UnsupportedDataFrameFunctions = [
```
EagleoutIce: Unsupported sounds like something that any not explicitly stated function is. Maybe Just `Other`?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+
+interface DataFrameFunctionMapperInfo<Params extends object> {
+	readonly mapper:      DataFrameFunctionMapping<Params>;
+	readonly library?:    string;
+	readonly returnType?: Exclude<DataFrameType, DataFrameType.DataFrame>;
+}
+
+interface UnsupportedDataFrameFunctionBase {
+	readonly type:        string;
+	readonly names:       string[];
```
EagleoutIce: why not add the `readonly` to the array as well?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+type UnsupportedDataFrameFunctionMapping = UnsupportedDataFrameEntryPoint | UnsupportedDataFrameTransformation | UnsupportedDataFrameModification;
+
+/**
+ * Data frame function mapper for mapping a concrete data frame function to abstract data frame operations.
+ * - `args` contains the function call arguments
+ * - `params` contains the expected argument location for each parameter of the function
+ * - `info` contains the resolve information
+ * - `config` contains the flowR configuration
+ */
+type DataFrameFunctionMapping<Params extends object> = (
```
EagleoutIce, line 572: Maybe add *Abstraction* into the name?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+/**
+ * Mapper type for mapping each supported data frame function of {@link DataFrameFunctionMapper} to the parameters required for the respective data frame function mapper.
+ * - `critical` contains all function parameters in which case the function call is unsupported and has to be over-approximated by the abstract operation `unknown`
+ */
+type DataFrameFunctionParamsMapping = {
+	[Name in DataFrameFunction]: DataFrameFunctionParams<Name> & { critical?: FunctionParameterLocation<unknown>[] }
+}
+
+/**
+ * Maps a concrete data frame function call to abstract data frame operations.
```
EagleoutIce, line 594: then call it `abstract...`?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+ */
+export function mapDataFrameFunctionCall<Name extends DataFrameFunction>(
+	node: RNode<ParentInformation>,
+	dfg: DataflowGraph,
+	config: FlowrConfigOptions = defaultConfigOptions
+): DataFrameExpressionInfo | undefined {
+	const resolveInfo = { graph: dfg, idMap: dfg.idMap, full: true, resolve: VariableResolve.Alias };
+	let operations: DataFrameOperation[] | undefined;
+
+	if(node.type === RType.FunctionCall && node.named) {
```
EagleoutIce: Maybe invert this guard? because this causes a lot of nesting

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+ *
+ * @param node   - The R node of the function call
+ * @param dfg    - The data flow graph for resolving the arguments
+ * @param config - The flowr configuration to use
+ * @returns Data frame expression info containing the mapped abstract data frame operations, or `undefined` if the node does not represent a data frame function call
+ */
+export function mapDataFrameFunctionCall<Name extends DataFrameFunction>(
+	node: RNode<ParentInformation>,
+	dfg: DataflowGraph,
+	config: FlowrConfigOptions = defaultConfigOptions
```
EagleoutIce: I do not think these functions should have defaults for the flowR config. The only thing this causes is someone forgetting to pass it a long and then have weirdly inconsistent behavior as we use different configurations

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+			}
+		} else {
+			const mapping = getUnsupportedDataFrameFunction(node.functionName.content);
+
+			if(mapping?.type === 'entry_point') {
+				operations = [{ operation: 'unknown', operand: undefined }];
+			} else if(mapping?.type === 'transformation' || mapping?.type === 'modification') {
+				const args = getFunctionArguments(node, dfg);
+				operations = mapDataFrameUnknown(args, mapping, resolveInfo);
+			}
```
EagleoutIce: Maybe make such separations covering including `assertUnreachable`?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+	}
+	return [{
+		operation: 'identity',
+		operand:   dataFrame.value.info.id
+	}];
+}
+
+function mapDataFrameRead(
+	args: readonly RFunctionArgument<ParentInformation>[],
+	params: {
```
EagleoutIce, line 704: This should be an interface i think, this is just.... a lot :D

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -0,0 +1,1440 @@
...
+		fileName:      FunctionParameterLocation,
```
EagleoutIce, line 705: Why no explicit generic here? It feels like a situation where you "have" to know the default to know what `text` may be

src/abstract-interpretation/data-frame/semantics.ts
```diff
@@ -0,0 +1,491 @@
...
+
+function applyCreateSemantics(
+	value: DataFrameDomain,
+	{ colnames, rows }: { colnames: (string | undefined)[] | undefined, rows: number | [number, number] | undefined }
+): DataFrameDomain {
+	const cols = colnames?.length;
+
+	return {
+		colnames: colnames?.every(isNotUndefined) ? colnames : ColNamesTop,
+		cols:     cols !== undefined ? [cols, cols] : IntervalTop,
```
EagleoutIce, line 112: does this already filter duplicate/illegal names?

src/dataflow/environments/built-in.ts
```diff
@@ -215,8 +215,9 @@ export const BuiltInProcessorMapper = {
 } as const satisfies Record<`builtin:${string}`, BuiltInIdentifierProcessorWithConfig<never>>;
 
 export const BuiltInEvalHandlerMapper = {
-	'built-in:c':     resolveAsVector,
-	'builtin:vector': resolveAsVector
+	'built-in:c': resolveAsVector,
+	'built-in::': resolveAsSeq,
+	'built-in:-': resolveAsMinus
```
EagleoutIce, line 221: "Why only `-`"?

src/documentation/doc-util/doc-query.ts
```diff
@@ -44,6 +45,14 @@ The analysis required _${printAsMs(duration)}_ (including parsing and normalizat
 
 ${codeBlock('json', collapseQuery ? str.split('\n').join(' ').replace(/([{[])\s{2,}/g,'$1 ').replace(/\s{2,}([\]}])/g,' $1') : str)}
 
+${(function() {
```
EagleoutIce, line 48: why is this listed as... added?

src/documentation/print-interface-wiki.ts
```diff
@@ -226,6 +227,16 @@ ${codeBlock('json', JSON.stringify(
...
 					threshold: 50
 				}
+			},
+			abstractInterpretation: {
+				dataFrame: {
+					maxColNames:       20,
+					wideningThreshold: 4,
+					readLoadedData:    {
+						readExternalFiles: true,
+						maxReadLines:      10_000_000
```
EagleoutIce: That is a lot lot lot of lines to load by default (given that the columncount may scale that easily)

src/abstract-interpretation/data-frame/semantics.ts
```diff
@@ -0,0 +1,491 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+	_args: {}
+): DataFrameDomain {
+	return value;
+}
+
+function applyUnknownSemantics(
+	_value: DataFrameDomain,
+	// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+	_args: {}
```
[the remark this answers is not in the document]
OliverGerstl: The empty object type is actually intended and required here, as were are creating a type union of this type with another type in `absint-info.ts` via `DataFrameOperationArgs<Name>`, which would not work for `Record<string, never>`

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: This is still the part that annoys me most about the control flow graph :)

EagleoutIce: yes i can see why.. but given that the CFG itself is unaware of the inversion... hard to fix from the outside or do you have an idea? I would be happy to given the new use of the CFG just optimize this within the graph class

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: The `toArray` method is used to convert the `MapIterator` resulting from `keys()` into an array

EagleoutIce: Oh yes, missed that we are not within arrays :D

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: It's really not a big thing, I would just rename the “DF/CF Dependency Edges” generally to “DF/CF Edges”, change "outgoing" to "incoming" and "incoming" to "outgoing", and reverse the edges in the CFG Mermaid diagrams accordingly, as their are currently a bit difficult to read.
But if you know the concept/idea behind it, it's not a big deal :)

src/abstract-interpretation/data-frame/domain.ts
OliverGerstl: Oh yes, you are right.
I didn't even notice, that the symbols are not JSON serialized at all (e.g. in the query) :sweat_smile:
I added the string literals back again

src/abstract-interpretation/data-frame/absint-visitor.ts
EagleoutIce: but if we swap this other algorithms that want to traverse them in the other direction... break but i see the problem

src/abstract-interpretation/data-frame/absint-info.ts
OliverGerstl: What exactly do you mean by "side effects that this expression may or may not have" in the context of the data frame operations?

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: For me, it's just the naming that is a bit unintuitive. Of course, backward traversals are also very important :)

src/abstract-interpretation/data-frame/absint-info.ts
EagleoutIce: conventionally an `expression` would refer to something that does not have a side effect. E.g. a `remove(df, "foo")` that returns a _new_ dataframe. However, given that R does not make that separation and calls something like `removeInPlace(df, "foo")` an expression as well, it would be nice to document which naming scheme we follow

src/abstract-interpretation/data-frame/domain.ts
OliverGerstl: The columns names do not imply order, it was just a (weird) design decision at the beginning to use arrays instead of sets. This will be changed in #1752 with #1760

EagleoutIce: Ok this is fine, but i guess then #1760 is scheduled to happen _after_ this PR?

src/abstract-interpretation/data-frame/absint-info.ts
OliverGerstl: Ah ok. The `type` property of an `DataFrameOperation` can be used to overwrite the `ConstraintType` (operand precondition, operand modification, or result postcondition) of an abstract operation, which should already been documented, but I may extend the documentation there

src/abstract-interpretation/data-frame/domain.ts
OliverGerstl: Yes, unfortunately this is currently out of scope for the thesis :(

EagleoutIce: No worries, this is perfectly fine :)

EagleoutIce: Although afterward, I would be interested whether we get performance benefits just from changing the structure and avoiding the recasting on operations

src/abstract-interpretation/data-frame/mappers/assignment-mapper.ts
OliverGerstl: What do you mean by "we die in functions"? :sweat_smile: 
Do you mean that we cannot detect side effects of unknown data frame functions?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: But wouldn't this make it harder to differentiate if we have to check imported libraries in the future?
For example, if the data frame functions come from different base R packages like "base", "stats", and "utils".
Wouldn't it be more difficult to differentiate which functions are from base R that we do not have to check, and which functions are from external packages for which we have to check if they were imported before?

src/abstract-interpretation/data-frame/mappers/assignment-mapper.ts
EagleoutIce: no i mean we cannot detect a dataframe transformer function within a function definition but i think we should just document that that in general these only kick when we know it is a data frame

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: Not really no, because ideally you would then integrate with the project analysis context from the Project Builder which tells you all loaded libraries, functions etc :)

src/abstract-interpretation/data-frame/mappers/assignment-mapper.ts
OliverGerstl: Ah ok, yes we currently do not support custom data frame functions and function definitions dealing with data frames

src/abstract-interpretation/data-frame/semantics.ts
OliverGerstl: Duplicate and illegal names are filtered in the function mappers mapping the concrete data frame functions to the abstract operations (e.g. [here](https://github.com/flowr-analysis/flowr/blob/ea95a96d1026d04244acf73d7331392f8056f714/src/abstract-interpretation/data-frame/mappers/function-mapper.ts#L643)), since replacing duplicate/illegal names is not done by every function and can often be disabled via arguments, depending on the concrete data frame function

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: There is no generic here since I do not allow a default value for the `fileName` argument

src/abstract-interpretation/data-frame/semantics.ts
EagleoutIce: Ok, but then this _expects_ the name to be sanitized - which is fine, but something we should document imho

src/documentation/doc-util/doc-query.ts
OliverGerstl: You added this when you added the " df-shape" query :sweat_smile:

EagleoutIce: oh :see_no_evil:  :grin:

EagleoutIce: I just thought "I added this" but I forgot this isn't on main yet

src/dataflow/environments/built-in.ts
OliverGerstl: What do you mean by "Why only `-`"?
For resolving vectors, we support negated values and negated values in vectors like `c(-1, -2)` and `-c(1, 2)`. Therefore, I added the `resolveAsMinus` eval handler.

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: hmm I am fine with that for now, but in the future maybe something like `FunctionParmaterLocationWithDefault` and ...WithoutDefault?

src/dataflow/environments/built-in.ts
EagleoutIce: Oh... sorry the second half of the comment appears to be missing. I was just thinking that `+c(1,2)` should be easy to handle and feels like the "right" counterpart - but this is just something we can outsource to another issue and it is by god not required or anything for the thesis :D

src/documentation/print-interface-wiki.ts
OliverGerstl: Mmh, that was the value I tested on a CSV file with 10 columns and it took only a few seconds on my machine.
But maybe I reduce the value by one digit, just to be safe :+1:

EagleoutIce: Given that this is a default I would tend to lower values and then add something to flowR that adds information like: "blocked inclusion of a CSV with 60 million lines" and then the extension (e.g.) provides a button to automatically raise the threshold to include this file :D

EagleoutIce: Additionally I think it is more of a future-proofing fear. If we ever extend the data frame domain to include information about the, e.g., distribution of data, this gets much more expensive

src/dataflow/environments/built-in.ts
OliverGerstl: Ah ok, but that's not much effort, I added the eval handler for `+` :)

EagleoutIce: ok :D (although now we have an eval for your thesis that did not have this handler so it may be better to still do this extra)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Mmh, I actually quite like the fact that the parameters are all equally structured in the function definitions, and I think that the readability is not greatly restricted by this

EagleoutIce: Allright, in that case let's leave it at that for now :)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Mmh, I think this should be clear enough with the comments, since the type is only used locally within the file. Otherwise, I would have to rename all other types as well, making the type names unnecessarily long in my opinion

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Mmh, I think this should be clear enough with the comments and, at the moment, I can't think of a meaningful, not too long name. Otherwise, I would have to rename all other mapping functions as well, making the names unnecessarily long in my opinion

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: ok :)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: :P

src/dataflow/environments/built-in.ts
OliverGerstl: Yes, I have to do the benchmarking again anyway because I changed several things.
That can then run nicely during the Nabada :D

EagleoutIce: :P :+1:

EagleoutIce:
_This pull request is included in v2.3.0 (see [Release v2.3.0 (Data Frame Shape Inference)](https://github.com/flowr-analysis/flowr/releases/tag/v2.3.0))._

PR #1469, opened 2025-03-15
Added `safeSizeOf` method to handle `sizeOfObject` bug

### LINKED ISSUE #1468: [Bug]: `sizeOfObject` is object instead of number
### What happened?

While running benchmarks for flowr with the pointer analysis enabled, the size of the vertices got so big that the used `sizeOf` method returned an error object, which wasn't handled before.

When calculating the sum of `sizeOfObject` and this happens, the output will result in a long string containing all numbers concatenated because the object couldn't be added and was concatenated instead.

_No response_

v2.2.11 (Mar 2, 2025)

### Related input code

```R

```

### Relevant log output

```Shell

```

### Meta Information

- [ ] The Issue is Urgent
- [ ] I Used the Dockerized Version of flowr

Slartibartfass2:
Closes #1468 

If an error object is returned by the `sizeOf` method, the array is split in half and the size is calculated recursively.


EagleoutIce (approved):

EagleoutIce:

PR #1484, opened 2025-03-18
Restructure absint info for data frames

OliverGerstl:


src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -1,36 +1,34 @@
...
-	arguments: (NodeId | undefined)[]
+	arguments: (NodeId | undefined)[],
+	modify?:   boolean
 }
 
-interface DataFrameStatementInfo {
-	type:   'statement',
-	domain: Map<NodeId, DataFrameDomain>
+interface DataFrameInfo {
+	type:    string;
```
EagleoutIce, line 13: We can discuss this on friday, for now I am fine with leaving the super-type unbounded.

EagleoutIce (approved):
This looks fine but is a lot of semantics, we can discuss them on friday :)

PR #1518, opened 2025-03-25
[Dependencies, III.1]: Types for R Values 

gigalasr:


EagleoutIce (approved):
I mean we definitely need more tests for this later, but for now this is great :D

Issue #1520, opened 2025-03-27
The new CFG


EagleoutIce:
* [x] #1526 
  * every function call should get a link to the processor type that created it
  * additionally it should be added with package information (unknown at first if we are unable to resolve)
  * include configuration to the nodes but ensure that the memory requirements for this are manageable
* [ ] #1527
  * Allow to check for, say `base::<-` within a functin call
  * receive the call, with the dfg and prepared information on what it resolve to etc.
* [ ] #1528 
  * Allow to traverse the CFG in either order
  * Allow for repeated link inlining or selective links. 
* [ ] #1529

PR #1530, opened 2025-04-04
Implement fold over CFG for forward traversal

OliverGerstl:
The CFG will be replaced with the new CFG (#1520)

### DIFF
--- a/src/abstract-interpretation/data-frame/domain.ts
+++ b/src/abstract-interpretation/data-frame/domain.ts
@@ -60,10 +60,10 @@ export function meetColNames(X1: ColNamesDomain, X2: ColNamesDomain): ColNamesDo
 }
 
 export function subtractColNames(X1: ColNamesDomain, X2: ColNamesDomain): ColNamesDomain {
-	if(X2 === ColNamesTop) {
-		return ColNamesBottom;
-	} else if(X1 === ColNamesTop) {
+	if(X1 === ColNamesTop) {
 		return ColNamesTop;
+	} else if(X2 === ColNamesTop) {
+		return X1;
 	} else {
 		return Array.from(new Set(X1).difference(new Set(X2)));
 	}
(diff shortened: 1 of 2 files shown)


src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -26,6 +29,12 @@ export const DataFrameTop: DataFrameDomain = {
 	rows:     IntervalTop
 };
 
+export type DataFrameStateDomain = Map<NodeId, DataFrameDomain>;
+
+export function equalColNames(X1: ColNamesDomain, X2: ColNamesDomain): boolean {
+	return X1 === X2 || (X1 !== ColNamesTop && setEquals(new Set(X1), new Set(X2)));
+}
+
 export function leqColNames(X1: ColNamesDomain, X2: ColNamesDomain): boolean {
```
EagleoutIce, line 38: I think there is a merit in generalizing these operations so that they are not specific to the colnames domain - given that they just represent the exact set domain

src/abstract-interpretation/data-frame/processor.ts
```diff
@@ -0,0 +1,44 @@
...
+	node: RNode<OtherInfo & ParentInformation>,
+	domain: DataFrameStateDomain,
+	dfg: DataflowGraph
+): DataFrameStateDomain | ConditionalDataFrameState {
+	switch(node.type) {
+		case RType.BinaryOp:
+			return processBinaryOp(node, domain, dfg);
+		case RType.FunctionCall:
+			return processFunctionCall(node, domain, dfg);
+		default:
```
EagleoutIce, line 23: why not go to top?

src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -60,6 +69,10 @@ export function subtractColNames(X1: ColNamesDomain, X2: ColNamesDomain): ColNam
 	}
```
EagleoutIce, line 69: regarding subtract col names: why does `x - TOP` return Bottom?

EagleoutIce (approved):

src/abstract-interpretation/data-frame/domain.ts
EagleoutIce: but isn't TOP equivalent to "we do not know which column names are removed"?

If i do `select(-contains(u))` to remove all columns that contain an unknown string, then the resulting dataframe might be empty (which woudln't be bot in our definition but an empty list right?) or it could contain elements - but it would not be invalid

PR #1531, opened 2025-04-04
Adapt the data frame processors and semantics to the new CFG

OliverGerstl:


EagleoutIce (approved):
the good thin is, the changes for the new cfg should be rather small. 
Let's use our next meeting to discuss the details of the individual mappings like `mapDataFrameCreate`

PR #1535, opened 2025-04-06
Infer types for constant expressions

loki259:
This will add type inference for logical, numeric and string constants 😎

### DIFF
--- a/src/typing/type-inferencer.ts
+++ b/src/typing/type-inferencer.ts
@@ -1,4 +1,5 @@
 import { DefaultNormalizedAstFold } from '../abstract-interpretation/normalized-ast-fold';
+import type { RExpressionList } from '../r-bridge/lang-4.x/ast/model/nodes/r-expression-list';
 import type { RLogical } from '../r-bridge/lang-4.x/ast/model/nodes/r-logical';
 import type { RNumber } from '../r-bridge/lang-4.x/ast/model/nodes/r-number';
 import type { RString } from '../r-bridge/lang-4.x/ast/model/nodes/r-string';
@@ -6,19 +7,21 @@ import type { RString } from '../r-bridge/lang-4.x/ast/model/nodes/r-string';
 export enum RDataType {
     Logical = 'LogicalType',
     Numeric = 'NumericType',
-    String = 'StringType',
+    String  = 'StringType',
+	Null    = 'NullType',
 }
 
-export type TypingInfo = RDataType | undefined;
+export type TypingInfo = RDataType
+					   | undefined; // Represents missing type information
 
 export class TypeInferencer<Info> extends DefaultNormalizedAstFold<TypingInfo, Info> {
 	constructor() {
 		super(undefined);
 	}
 
-	protected override concat(a: TypingInfo, b: TypingInfo): TypingInfo {
+	protected override concat(_a: TypingInfo, _b: TypingInfo): TypingInfo {
 		// Do not combine typing information for now
-		return b;
+		return undefined;
 	}
 
 	override foldRLogical(_node: RLogical<Info>): TypingInfo {
@@ -31,5 +34,14 @@ export class TypeInferencer<Info> extends DefaultNormalizedAstFold<TypingInfo, I
 
 	override foldRString(_node: RString<Info>): TypingInfo {
 		return RDataType.String;
+	}
+
+	override foldRExpressionList(exprList: RExpressionList<Info>): TypingInfo {
+		if(exprList.children.length === 0) {
+			return RDataType.Null;
+		} else {
+			// Do not handle early returns for now
+			return this.fold(exprList.children[exprList.children.length - 1]);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/typing/type-inferencer.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { RNumber } from '../r-bridge/lang-4.x/ast/model/nodes/r-number';
+import type { RString } from '../r-bridge/lang-4.x/ast/model/nodes/r-string';
+
+export enum RDataType {
+    Logical = 'LogicalType',
+    Numeric = 'NumericType',
+    String = 'StringType',
+}
+
+export type TypingInfo = RDataType | undefined;
```
EagleoutIce, line 12: What is the intended Semantic of `undefined`? Is it to be equivalent to `any` or should it just signal that we have no type information? 
(I mean given an incremental typing approach, TOP may be equialvent to "no information" but we may want to separate "we know that we know nothing" from "we haven't tried (yet)")

So if this should be TOP/any I would make it explicit, otherwise I would at least clarify the meaning of `undefined` with a comment.

src/typing/type-inferencer.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class TypeInferencer<Info> extends DefaultNormalizedAstFold<TypingInfo, Info> {
+	constructor() {
+		super(undefined);
+	}
+
+	protected override concat(a: TypingInfo, b: TypingInfo): TypingInfo {
+		// Do not combine typing information for now
+		return b;
```
EagleoutIce, line 21: I think to stay correct this has to yield TOP. Because otherwise `1 + TRUE` should yield `logical`, and even function calls may just "pass along" the type.

src/typing/type-inferencer.ts
```diff
@@ -0,0 +1,35 @@
...
+	override foldRLogical(_node: RLogical<Info>): TypingInfo {
+		return RDataType.Logical;
+	}
+
+	override foldRNumber(_node: RNumber<Info>): TypingInfo {
+		return RDataType.Numeric;
+	}
+
+	override foldRString(_node: RString<Info>): TypingInfo {
+		return RDataType.String;
```
EagleoutIce, line 33: Just as food for thought: do you want to tag raw strings specifically in the future?

src/typing/type-inferencer.ts
```diff
@@ -0,0 +1,35 @@
...
+		// Do not combine typing information for now
+		return b;
+	}
+
+	override foldRLogical(_node: RLogical<Info>): TypingInfo {
+		return RDataType.Logical;
+	}
+
+	override foldRNumber(_node: RNumber<Info>): TypingInfo {
+		return RDataType.Numeric;
```
EagleoutIce, line 29: You should do the `complex` recognition here if you want to separate `1i` from `1`.
Additionally: How do you want to map values such as `NA`, `NaN` etc.?

test/functionality/typing/constant-type-inference.test.ts
```diff
@@ -0,0 +1,33 @@
+import { beforeAll, describe, expect, test } from 'vitest';
+import { retrieveNormalizedAst, withShell } from '../_helper/shell';
+import type { NormalizedAst } from '../../../src/r-bridge/lang-4.x/ast/model/processing/decorate';
+import { RDataType, TypeInferencer } from '../../../src/typing/type-inferencer';
+
+describe.sequential('Infer types for constant expressions', withShell(shell => {
```
EagleoutIce, line 6: I think it is safe to switch to the Tree-Sitter engine for the tests (which should speed them up as well :D). We can discuss this in the next meeting if you want

test/functionality/typing/constant-type-inference.test.ts
```diff
@@ -0,0 +1,33 @@
...
+	let logicalAst: NormalizedAst | undefined;
+	let numberAst: NormalizedAst | undefined;
+	let stringAst: NormalizedAst | undefined;
+	const inferencer = new TypeInferencer();
+
+	beforeAll(async() => {
+		logicalAst = await retrieveNormalizedAst(shell, 'TRUE');
```
EagleoutIce, line 13: Doing it this way for a quick test is perfectly fine! 
For the future we may want to separate these parse calls and potentially just move them to the respective test which then becomes a wrapper function similar to `assertDataflow` etc.

test/functionality/typing/constant-type-inference.test.ts
```diff
@@ -0,0 +1,33 @@
...
+		numberAst  = await retrieveNormalizedAst(shell, '42');
+		stringAst  = await retrieveNormalizedAst(shell, '"Hello, world!"');
+
+	});
+
+	test(`Infer ${RDataType.Logical} for logical constants`, () => {
```
EagleoutIce, line 19: If you want to keep this style for simple tests, maybe you want to use [`test.each`](https://vitest.dev/api/#test-each)?

src/typing/type-inferencer.ts
EagleoutIce: I think it is a good idea to add 2-3 more tests for cases like `1 + TRUE` just to make sure we do not wrongly return a specific type.

EagleoutIce (approved):

src/typing/type-inferencer.ts
loki259: In the early stages of the system the plan is to not distinguish between normal values and NA/NaN, just like we do not distinguish between scalars and vectors for now. In later design iterations such distinctions will be made on the basis of subtyping.

src/typing/type-inferencer.ts
loki259: Not for now, but maybe we can incorporate this distinction when we implement subtyping.

PR #1546, opened 2025-04-11
create search api queries from old query api queries

Ellpeck:


Ellpeck:
should probably have more tests but i would like input on whether this is what you had in mind :)

EagleoutIce:
Yes this is roughly what i wanted.

The call context query should return _every_ resulting call but of course not the linked ids etc.

EagleoutIce:
If you want you can add an argument to the flattening which can then be used in the query as well.

PR #1588, opened 2025-04-22
Linter Rule 2: File Path Boogaloo

Ellpeck:


Ellpeck:
Before this is review-ready, we need to get #1487 in line with the current version of main so that this PR doesn't look insane

EagleoutIce (approved):

PR #1642, opened 2025-05-11
Remove singleton config instance

### LINKED ISSUE #1641: Remove singleton config instance
Instead of a single config instance instance in `config.ts`, we want to pass the config down to where it is needed. This will allow us to have a config scoped to each analyzer instance created via the planned builder API (#1512).

MaxAtoms:
When your first PR touches like half of the project's files ✌🏽 

Remaining `getConfig()` call-sites:

```
        src/cli  (6 usages found)
            benchmark-helper-app.ts  (1 usage found)
                benchmark  (1 usage found)
                    62 let config = getConfig();
            export-quads-app.ts  (1 usage found)
                27 const shell = new RShell(getEngineConfig(getConfig(), 'r-shell'));
            flowr.ts  (1 usage found)
                createConfig  (1 usage found)
                    109 config = getConfig(options['config-file'] ?? defaultConfigFile);
            slicer-app.ts  (1 usage found)
                getSlice  (1 usage found)
                    48 const config = getConfig();
            statistics-app.ts  (1 usage found)
                28 void flowrScriptGetStats(scriptOptions, getConfig());
            statistics-helper-app.ts  (1 usage found)
                27 void getStatsForSingleFile(scriptOptions, getConfig());
```

Closes #1641

### DIFF
--- a/src/benchmark/slicer.ts
+++ b/src/benchmark/slicer.ts
@@ -24,7 +24,7 @@ import type {
 } from './stats/stats';
 import type { NormalizedAst } from '../r-bridge/lang-4.x/ast/model/processing/decorate';
 import type { SlicingCriteria } from '../slicing/criterion/parse';
-import type { TREE_SITTER_SLICING_PIPELINE, DEFAULT_SLICING_PIPELINE } from '../core/steps/pipeline/default-pipelines';
+import type { DEFAULT_SLICING_PIPELINE, TREE_SITTER_SLICING_PIPELINE } from '../core/steps/pipeline/default-pipelines';
 import { createSlicePipeline } from '../core/steps/pipeline/default-pipelines';
 
 
@@ -135,7 +135,7 @@ export class BenchmarkSlicer {
 				if(this.parserName === 'r-shell') {
 					return new RShell(getEngineConfig(config, 'r-shell'));
 				} else {
-					await TreeSitterExecutor.initTreeSitter(config);
+					await TreeSitterExecutor.initTreeSitter(getEngineConfig(config, 'tree-sitter'));
 					return new TreeSitterExecutor();
 				}
 			}
--- a/src/benchmark/slicer.ts
+++ b/src/benchmark/slicer.ts
@@ -46,6 +46,8 @@ import type { InGraphIdentifierDefinition } from '../dataflow/environments/ident
 import type { ContainerIndicesCollection } from '../dataflow/graph/vertex';
 import { isParentContainerIndex } from '../dataflow/graph/vertex';
 import { equidistantSampling } from '../util/collections/arrays';
+import type { FlowrConfigOptions } from '../config';
+import { getEngineConfig } from '../config';
 
 /**
  * The logger to be used for benchmarking as a global object.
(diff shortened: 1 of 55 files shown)


MaxAtoms:
@EagleoutIce I have not yet updated the tests and documentation in fear of duplicate work if `config.ts` changes again. Maybe you want to have a look now and let me know if you have any comments. But no need to hurry from my side :)

src/benchmark/slicer.ts
```diff
@@ -123,16 +125,17 @@ export class BenchmarkSlicer {
 	 * Initialize the slicer on the given request.
 	 * Can only be called once for each instance.
 	 */
-	public async init(request: RParseRequestFromFile | RParseRequestFromText, autoSelectIf?: AutoSelectPredicate, threshold?: number) {
+	public async init(request: RParseRequestFromFile | RParseRequestFromText, config: FlowrConfigOptions,
```
EagleoutIce, line 128: Looking forward to chaining that once we have the project interface :D

src/cli/flowr.ts
```diff
@@ -88,47 +88,58 @@ if(options['no-ansi']) {
...
 }
 
-let usedConfig = false;
-if(options['config-json']) {
-	const config = parseConfig(options['config-json']);
-	if(config) {
-		log.info(`Using passed config ${JSON.stringify(config)}`);
-		setConfig(config);
-		usedConfig = true;
+function createConfig() : FlowrConfigOptions {
```
EagleoutIce, line 91: In the future we probably want to have one general create, but for now this is fine :)

src/cli/flowr.ts
```diff
@@ -150,7 +161,22 @@ async function retrieveEngineInstances(): Promise<{ engines: KnownEngines, defau
 	return { engines, default: defaultEngine };
 }
 
+function hookSignalHandlers(engines: { engines: KnownEngines; default: keyof KnownEngines }) {
```
EagleoutIce, line 164: In the long term this should move to the flowR project wrapper

src/cli/repl/commands/repl-cfg.ts
```diff
@@ -44,8 +44,8 @@ export const controlflowCommand: ReplCommand = {
 	usageExample: ':controlflow',
 	aliases:      [ 'cfg', 'cf' ],
 	script:       false,
-	fn:           async(output, shell, remainingLine) => {
-		await produceAndPrintCfg(shell, remainingLine, output, [], cfgToMermaid);
+	fn:           async(output, shell, remainingLine, _, config) => {
```
EagleoutIce, line 47: Given that this now gets longer and longer we should transition to an object based passing style so that we can select the keys of interest! What do you say @MaxAtoms?

src/cli/repl/commands/repl-commands.ts
```diff
@@ -52,7 +52,7 @@ export const helpCommand: ReplCommand = {
 	script:       false,
 	usageExample: ':help',
 	aliases:      [ 'h', '?' ],
-	fn:           output => {
+	fn:           (output) => {
```
EagleoutIce, line 55: Was this your linters doing? in general i think we omit the parentheses for single-arg arrow syntax

src/dataflow/processor.ts
```diff
@@ -40,7 +41,10 @@ export interface DataflowProcessorInformation<OtherInfo> {
 	 * The chain of control-flow {@link NodeId}s that lead to the current node (e.g., of known ifs).
 	 */
 	readonly controlDependencies: ControlDependency[] | undefined
-	
+	/**
+	 * The config
```
EagleoutIce, line 45:
```suggestion
	 * The flowr configuration used for environment seeding, and precision control
```

src/slicing/static/static-slicer.ts
```diff
@@ -33,7 +32,7 @@ export function staticSlicing(
 	graph: DataflowGraph,
 	{ idMap }: NormalizedAst,
 	criteria: SlicingCriteria,
-	threshold = getConfig().solver.slicer?.threshold ?? 75,
+	threshold = 75,
```
EagleoutIce, line 35: We should definitely not lose the configurable threshold! 
Let's update the parameter to have an undefined default, or accept the config explicitly

EagleoutIce (commented):
Right now it is hard to check whether config was adapted consistently, so I wait with that until the tests pass so we have a safe-guard :D

src/cli/repl/commands/repl-cfg.ts
MaxAtoms: Sound good :) I added this with 6d15b86c87902ce753c21f18c46f7209cd15d98a

PR #1700, opened 2025-05-28
Forward Slicing

EagleoutIce:
To be done (at least things i can think of right now):

* [x] New Pipeline Steps for forward slicing (-> add a flag to the old step to make it go forward)
* [x] New Query for the forward slice
* [x] Flag to `assertSliced` which allows to check for forward slicing
* [x] A bunch of new tests 
  * [x] (as a starting point, we can just try forward slicing on all variables in the normal backward slicing tests and check whether a guard fails) -> **works!**
* [x] Documentation


EagleoutIce:
_This pull request is included in v2.4.0 (see [Release v2.4.0 (Forward-Slicing, Minor Fixes)](https://github.com/flowr-analysis/flowr/releases/tag/v2.4.0))._

PR #1725, opened 2025-06-04
Change `amendConfig` to accept an anonymous function parameter

### LINKED ISSUE #1521: [Bug]: Cannot amend config to use `maxIndexCount`
### What happened?

When one tries to amend the current config by passing an object for `pointerTracking` to set the `maxIndexCount` an exception is thrown because `deepMergeObject` can't handle union types.

https://github.com/flowr-analysis/flowr/blob/d012d3224839dd2c3e67c311de916444c293bddc/src/util/objects.ts#L29

This happens because the default config has `pointerTracking` set to true, and now `deepMergeObject` can't merge a boolean with an object.

_No response_

v2.2.12 (Mar 17, 2025)

### Related input code

```R

```

### Relevant log output

```Shell

```

### Meta Information

- [ ] The Issue is Urgent
- [ ] I Used the Dockerized Version of flowr

### LINKED ISSUE #1623: `amendConfig` should accept a DeepPartial rather than just a Partial
You should be able to override single nested config options without having to do the whole `...` thing all the time

MaxAtoms:
As discussed, the arrays in `FlowrConfigOptions` need to be mutable in order to use `DeepWritable`.

closes #1521, closes #1623

### DIFF
--- a/src/dataflow/eval/values/string/string-constants.ts
+++ b/src/dataflow/eval/values/string/string-constants.ts
@@ -21,7 +21,7 @@ export function liftString(str: Lift<RStringValue>): ValueString {
 	};
 }
 
-export function collectStrings(a: Value[]): string[] | undefined {
+export function collectStrings(a: Value[], withQuotes: boolean = false): string[] | undefined {
 	if(bottomTopGuard(a)) {
 		return undefined;
 	}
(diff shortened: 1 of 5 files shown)


test/functionality/dataflow/processing-of-elements/functions/dataflow-source.test.ts
```diff
@@ -21,7 +28,11 @@ describe.sequential('source', withShell(shell => {
 	};
 	beforeAll(() => {
 		setSourceProvider(requestProviderFromText(sources));
-		amendConfig({ solver: { resolveSource: { repeatedSourceLimit: 0 } } });
+		amendConfig(c =>
+			c.solver.resolveSource = deepMergeObject(
```
MaxAtoms: This is the only ugly case because `resolveSource` is optional.
Alternative: `@ts-ignore`, I guess 😅

PR #1750, opened 2025-06-15
Use the new CFG visitor for abstract interpretation

OliverGerstl:
This
- merges the new control flow graph into the current abstract interpretation implementation,
- extends alias tracking for vectors to support the sequence operator `:` and negations,
- uses the new semantic-guided CFG visistor for abstract interpretation,
- and only adds data frame shape information to the AST information for nodes affecting the shape of a data frame


EagleoutIce:
_This pull request is included in v2.2.16 (see [Release v2.2.16 (Local Config, Fixes, New Linter Rules)](https://github.com/flowr-analysis/flowr/releases/tag/v2.2.16))._

PR #1766, opened 2025-06-19
refactor(datatype-query): reuse typed AST for all criteria

### LINKED ISSUE #1765: [Query] Datatype Query Repeatedly Infers Types For The Same AST In Case Of Multiple Queries
The query implementation should be optimized to reuse a single typed AST for all given criteria.

loki259:
Closes #1765


src/queries/catalog/datatype-query/datatype-query-executor.ts
```diff
@@ -8,12 +8,13 @@ import { inferDataTypes } from '../../../typing/infer';
 export function executeDatatypeQuery({ dataflow, ast }: BasicQueryData, queries: readonly DatatypeQuery[]): DatatypeQueryResult {
 	const start = Date.now();
 	const result: DatatypeQueryResult['inferredTypes'] = {};
+	const typedAst = inferDataTypes(ast as NormalizedAst< ParentInformation & { typeVariable?: undefined }>, dataflow);
```
EagleoutIce:
```suggestion
	const typedAst = inferDataTypes(ast as NormalizedAst<ParentInformation & { typeVariable?: undefined }>, dataflow);
```

If we are already at it :D

Issue #1597, opened 2025-04-29
Add config entry for widening threshold and max col names


OliverGerstl:
Add a config entries for the abstract interpretation of data frames to the flowR config for
- the data frame state widening threshold (`MaxColNames`)
- the maximum number of column names for the column names join operator (`DefaultWideningThreshold`)
- the maximum number of lines to read from external files (`MaxReadLines`)

PR #1772, opened 2025-06-21
Support loading data frames from external files

OliverGerstl:


src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -23,15 +28,20 @@ const DataFrameFunctionMapper = {
 	'read.csv2':     mapDataFrameRead,
 	'read.delim':    mapDataFrameRead,
 	'read.delim2':   mapDataFrameRead,
+	'read_table':    mapDataFrameRead,
+	'read_csv':      mapDataFrameRead,
```
EagleoutIce, line 33: Could we maybe explicitly flag functions for which we know that they product a tibble instead of a dataframe? just for documentation and communication purposes

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -2,17 +2,22 @@ import type { ResolveInfo } from '../../../dataflow/eval/resolve/alias-tracking'
...
+import type { RParseRequest } from '../../../r-bridge/retriever';
+import { requestFromInput } from '../../../r-bridge/retriever';
+import { readLineByLineSync } from '../../../util/files';
 import type { AbstractInterpretationInfo, DataFrameInfo, DataFrameOperations } from '../absint-info';
 import { resolveIdToAbstractValue } from '../absint-visitor';
 import { DataFrameTop } from '../domain';
-import { resolveIdToArgName, resolveIdToArgValue, resolveIdToArgValueSymbolName, resolveIdToArgVectorLength, unescapeArgument } from '../resolve-args';
+import { resolveIdToArgName, resolveIdToArgValue, resolveIdToArgValueSymbolName, resolveIdToArgVectorLength, unquoteArgument, unescapeSpecialChars } from '../resolve-args';
 
+const MaxReadLines = 1e7;
```
EagleoutIce, line 21: Why not add this to the flowr configuration?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -44,20 +54,25 @@ const DataFrameFunctionMapper = {
...
-	'read.csv2':     { fileName: { pos: 0, name: 'file' }, header: { pos: 1, name: 'header' }, separator: { pos: 2, name: 'sep' } },
-	'read.delim':    { fileName: { pos: 0, name: 'file' }, header: { pos: 1, name: 'header' }, separator: { pos: 2, name: 'sep' } },
-	'read.delim2':   { fileName: { pos: 0, name: 'file' }, header: { pos: 1, name: 'header' }, separator: { pos: 2, name: 'sep' } },
+	'read.table':    { fileName: { pos: 0, name: 'file' }, header: { pos: 1, name: 'header', default: false }, separator: { pos: 2, name: 'sep', default: '\\s' }, quote: { pos: 3, name: 'quote', default: '"\'' }, skipLines: { pos: 12, name: 'skip', default: 0 }, comment: { pos: 17, name: 'comment.char', default: '#' }, text: { pos: 23, name: 'text' } },
+	'read.csv':      { fileName: { pos: 0, name: 'file' }, header: { pos: 1, name: 'header', default: true }, separator: { pos: 2, name: 'sep', default: ',' }, quote: { pos: 3, name: 'quote', default: '"' }, skipLines: { pos: 12, name: 'skip', default: 0 }, comment: { pos: 17, name: 'comment.char', default: '' }, text: { pos: 23, name: 'text' } },
```
EagleoutIce, line 59: given that some of these call each other we may want to consider future support for multiple arguments that may contribute to a given type. In general we should discuss a consistent way to document the semantics/characteristics of arguments

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -44,20 +54,25 @@ const DataFrameFunctionMapper = {
...
+	'read_delim':    { fileName: { pos: 0, name: 'file' }, separator: { pos: 1, name: 'delim', default: '\t' }, quote: { pos: 2, name: 'quote', default: '"' }, header: { pos: 5, name: 'col_names', default: true }, comment: { pos: 12, name: 'comment', default: '' }, skipLines: { pos: 14, name: 'skip', default: 0 } },
 	'cbind':         { special: ['deparse.level', 'make.row.names', 'stringsAsFactors', 'factor.exclude'] },
 	'rbind':         { special: ['deparse.level', 'make.row.names', 'stringsAsFactors', 'factor.exclude'] },
-	'head':          { dataFrame: { pos: 0, name: 'x' }, amount: { pos: 1, name: 'n' } },
-	'tail':          { dataFrame: { pos: 0, name: 'x' }, amount: { pos: 1, name: 'n' } },
-	'subset':        { dataFrame: { pos: 0, name: 'x' }, subset: { pos: 1, name: 'subset' }, select: { pos: 2, name: 'select' }, drop: { pos: 3, name: 'drop' } },
+	'head':          { dataFrame: { pos: 0, name: 'x' }, amount: { pos: 1, name: 'n', default: 6 } },
```
EagleoutIce, line 70: i just looked it up because i thought they used a global option.. this feels so wrong :eyes: In general we may want to use an `RValue` here? But this is  not necessary for now

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+			request = requestFromInput(unescapeSpecialChars(text));
+		}
+	}
+	if(request === undefined) {
+		return [{
+			operation: 'read',
+			operand:   undefined,
+			args:      { source: source, colnames: undefined, rows: undefined }
+		}];
+	}
```
EagleoutIce: Ist there a reason for why you do not outsource this into a obtainRequestFromRead?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+	const quoteArg = getFunctionArgument(args, params.quote, info);
+	const commentArg = getFunctionArgument(args, params.comment, info);
+	const skipLinesArg = getFunctionArgument(args, params.skipLines, info);
+	const header = headerArg ? resolveIdToArgValue(headerArg, info) : params.header.default;
+	const separator = separatorArg ? resolveIdToArgValue(separatorArg, info) : params.separator.default;
+	const quote = quoteArg ? resolveIdToArgValue(quoteArg, info) : params.quote.default;
+	const comment = commentArg ? resolveIdToArgValue(commentArg, info) : params.comment.default;
+	const skipLines = skipLinesArg ? resolveIdToArgValue(skipLinesArg, info) : params.skipLines.default;
+
+	if(typeof header !== 'boolean' || typeof separator !== 'string' || typeof quote !== 'string' || typeof comment !== 'string' || typeof skipLines !== 'number') {
```
EagleoutIce, line 195: This may be a `guardTypes`?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+		return [{
+			operation: 'read',
+			operand:   undefined,
+			args:      { source: source, colnames: undefined, rows: undefined }
```
EagleoutIce: Is there a reason you do not use the shorthand?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+		}];
+	}
+	const LineCommentRegex = new RegExp(`\\s*[${comment}].*`);
```
EagleoutIce: Why do we drop leading spaces? I thought `read_table` only drops after

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+	let firstLine = undefined as (string | undefined)[] | undefined;
```
EagleoutIce, line 203: cursed :D

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -141,13 +157,110 @@ function mapDataFrameConvert(
...
+				rowCount++;
+			}
+		}
+	};
+
+	if(request.request === 'text') {
+		request.content.split('\n').forEach(parseLine);
+	} else if(request.request === 'file') {
+		allLines = readLineByLineSync(request.content, parseLine, MaxReadLines);
+	}
```
EagleoutIce: I would again outsource this into a function that acecpts a parser for content consumptions

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -698,7 +811,23 @@ function getEffectiveArgs(
...
+}
+
+function getEntriesFromCsvLine(line: string, sep: string = ',', quote: string = '"', comment = ''): (string | undefined)[] {
+	sep = escapeRegExp(sep, true);  // only allow tokens like `\s`, `\t`, or `\n` in separator, quote, and comment chars
+	quote = escapeRegExp(quote, true);
+	comment = escapeRegExp(comment, true);
+	const quantifier = sep === '\\s' ? '+' : '*';  // do not allow unquoted empty entries in whitespace-sparated files
+
+	const LineCommentRegex = new RegExp(`\\s*[${comment}].*`);
+	const CsvEntryRegex = new RegExp(`(?<=^|[${sep}])(?:[${quote}]((?:[^${quote}]|[${quote}]{2})*)[${quote}]|([^${sep}]${quantifier}))`, 'g');
```
EagleoutIce, line 847: Is this the regex they use as well? i thought readr relied on a state machine. If so we may be better of using the same behavior?

src/util/files.ts
```diff
@@ -95,31 +95,41 @@ export function writeTableAsCsv(table: Table, file: string, sep = ',', newline =
...
 	}
+	return true;
 }
 
 /**
  * Reads a file line by line and calls the given function for each line.
  * The `lineNumber` starts at `0`.
+ * The `maxLines` option limits the maximum number of read lines and is `Infinity` by default.
+ *
+ * @returns Whether all lines have been successfully read (`false` if `maxLines` was reached)
```
EagleoutIce: or file does not exist

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -852,11 +1024,11 @@ df <- transform(df, name = c(letters[1:5]))
 		]
 	);
 
-	describe.skipIf(skipDplyr)('dplyr Functions', () => {
+	describe.skipIf(skipLibraries)('dplyr Functions', () => {
 		testDataFrameDomain(
 			`
 df <- data.frame(id = 1:5, score = c(80, 75, 90, 70, 85))
-df <- dplyr::group_by(df, id) |> as.data.frame()
+df <- dplyr::group_by(df, id)
```
EagleoutIce, line 1031: do we drop this because we do not handle the cast? or because we dropped the separation?
I mean one safe way for `as.data.frame` to work would be to use identity if we already have df shape info

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -959,7 +1128,7 @@ df <- aggregate(df, list(group = df$id), mean)
 		]
 	);
 
-	describe.skipIf(skipDplyr)('dplyr Functions', () => {
+	describe.skipIf(skipLibraries)('dplyr Functions', () => {
```
EagleoutIce, line 1131: is this for debugging? Or for your thesis?  in any way i would document the flag

EagleoutIce (approved):

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Yes, I added issue #1786 for this

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Yes, I added this to #1597

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Yes, you are right. That was just a simple solution to define the arguments flexibly for now. We can of course discuss other solutions for the future

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: I normally prefer to state this explicitly, but I can also use the shorthand :)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: `read.csv` and `read_csv` ignore whitespaces around CSV entries by default

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: I just use this regex to parse a single CSV line with the default CSV behavior as minimal CSV support. If the CSV behavior is changed via the function parameters, we have to over-approximate the shape

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: Yes, `as.data.frame` is already mapped to identity. I just omit it here, since I want to test the `as.data.frame` method separately and ensure that we infer the correct shapes for `group_by` without additional functions

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: This is due to the fact that the CI pipeline fails when importing packages like `readr` or `dplyr` (e.g. [here](https://github.com/flowr-analysis/flowr/actions/runs/14551782269/job/40823827939)). Therefore, I am running all tests locally and have to omit the tests with library calls via `library` or `::` in the CI (see also [1777-add-more-tests-to-verify-capabilities-of-shape-inference](https://github.com/flowr-analysis/flowr/blob/dac990a15bdf799d5188f143bc1bb5cd9771b972/test/functionality/abstract-interpretation/data-frame/inference.test.ts#L8))

EagleoutIce: We can/should add lines to the execution which tries to install these packages or Skip these tests based on the availability on the system

OliverGerstl: Ok, I will add this to #1777

PR #1780, opened 2025-06-24
Project Builder API

MaxAtoms:


MaxAtoms:
The crude prototype can be executed via: `npx ts-node src/project/flowr-analyzer.ts`

src/project/flowr-analyzer-builder.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { RParseRequests } from '../r-bridge/retriever';
+import { FlowrAnalyzer } from './flowr-analyzer';
+import { retrieveEngineInstances } from '../engines';
+import type { KnownParser } from '../r-bridge/parser';
+
+export class FlowrAnalyzerBuilder {
+	private flowrConfig:      DeepWritable<FlowrConfigOptions> = cloneConfig(defaultConfigOptions);
+	private readonly request: RParseRequests;
+
+	public amendConfig(func: (config: DeepWritable<FlowrConfigOptions>) => FlowrConfigOptions) : FlowrAnalyzerBuilder {
```
EagleoutIce: You can give `this` as a return type which harmonizes with builder extensions :)

src/project/flowr-analyzer-builder.ts
```diff
@@ -0,0 +1,37 @@
...
+		this.flowrConfig = amendConfig(this.flowrConfig, func);
+		return this;
+	}
+
+	public setEngine(engine : 'tree-sitter' | 'r-shell') {
```
EagleoutIce: This should be the `KnownParser` type which... Should be renamed to known engine :p

src/project/flowr-analyzer.ts
```diff
@@ -0,0 +1,79 @@
...
+import { graphToMermaidUrl } from '../util/mermaid/dfg';
+import type { KnownParser } from '../r-bridge/parser';
+import type { SlicingCriteria } from '../slicing/criterion/parse';
+import type { Queries, SupportedQueryTypes } from '../queries/query';
+import { executeQueries } from '../queries/query';
+import type { DataflowInformation } from '../dataflow/info';
+import type { NormalizedAst } from '../r-bridge/lang-4.x/ast/model/processing/decorate';
+
+export class FlowrAnalyzer {
+	private readonly flowrConfig: FlowrConfigOptions = cloneConfig(defaultConfigOptions);
```
EagleoutIce: Why this initialization if it is overwritten in the ctor?

src/project/flowr-analyzer.ts
```diff
@@ -0,0 +1,79 @@
...
+	private ast = undefined as unknown as NormalizedAst;
+	private dataflowInfo = undefined as unknown as DataflowInformation;
+
+	constructor(config: FlowrConfigOptions, parser: KnownParser, request: RParseRequests) {
+		this.flowrConfig = config;
+		this.request = request;
+		this.parser = parser;
+	}
+
+	public async dataflow() : Promise<PipelineOutput<typeof DEFAULT_DATAFLOW_PIPELINE>> {
```
EagleoutIce: Using default data flow pipeline here is... Wrong in general we should select the type based on the engine :D in general we maybe want to embed a cache?

src/project/flowr-analyzer.ts
```diff
@@ -0,0 +1,79 @@
...
+
+	public async query(query: Queries<SupportedQueryTypes>) {
+		if(!this.dataflowInfo) {
+			await this.dataflow();
+		}
+		return executeQueries({ ast: this.ast, dataflow: this.dataflowInfo, config: this.flowrConfig }, query);
+	}
+}
+
+async function main() {
```
EagleoutIce: To toy this is fine :D in general we should at least do the classical main module check (see the wiki page printers)

EagleoutIce (commented):
I really like the basic structure :)

src/project/flowr-analyzer-builder.ts
MaxAtoms: So we want to have an extra field for the engine that has precedence over the options?
It gets a bit confusing when doing something like this:

```
await new FlowrAnalyzerBuilder(require("x <- 1"))
	.setEngine(new RShell())
	.amendConfig(c => {
		c.defaultEngine = 'tree-sitter'
		return c;
	}).build();
```

src/project/flowr-analyzer.ts
MaxAtoms: > in general we maybe want to embed a cache?

Yes, I think that would make sense. We need to have the dataflow analysis results for the query execution anyways. In theory, the slicing pipeline could re-use the dataflow step results, but I'm not sure whether we should enable this right now

src/project/flowr-analyzer-builder.ts
MaxAtoms: Correction: Take `name` property from `KnownParser`

MaxAtoms: `amendConfig` should accept a function that can return `FlowrConfigOptions | void`.
Disabling the eslint warning is fine

src/project/flowr-analyzer-builder.ts
```diff
@@ -0,0 +1,56 @@
...
+		this.flowrConfig.defaultEngine = engine;
+		return this;
+	}
+
+	constructor(request: RParseRequests, plugins?: AnyFlowrAnalyzerPlugin[]) {
+		this.request = request;
+		this.plugins = plugins ? plugins : [];
+	}
+
+	public registerPlugin(plugin: AnyFlowrAnalyzerPlugin | AnyFlowrAnalyzerPlugin[]): void {
```
MaxAtoms: Use spread operator

MaxAtoms: Should return `this`

src/project/flowr-analyzer-builder.ts
```diff
@@ -0,0 +1,56 @@
...
+		if(Array.isArray(plugin)){
+			this.plugins.push(...plugin);
+		} else {
+			this.plugins.push(plugin);
+		}
+	}
+
+	public unregisterPlugin(plugin: AnyFlowrAnalyzerPlugin | AnyFlowrAnalyzerPlugin[]): void {
```
MaxAtoms: Same here: Spread operator and `this`

PR #1781, opened 2025-06-24
doc-fix: don't use subdirectories for linting rules

Ellpeck:


EagleoutIce (approved):
Let's hope for the best!

PR #1785, opened 2025-06-27
[Linter] Basic Dead Code Detection

Ellpeck:


Ellpeck:
I don't think I'm happy with the solution in 77285c2930cdf337a5be6fd9bb7c4c46c09b689b since, right now, it can't be reused for multiple distinct searches. I do like storing additional information in the search elements object like this though, so I don't know. What do you think abt this, @EagleoutIce?

src/dataflow/graph/dataflowgraph-builder.ts
```diff
@@ -218,7 +218,7 @@ export class DataflowGraphBuilder extends DataflowGraph {
 		if('target' in to) {
 			toIds = to.target;
 		} else {
-			const result = runSearch(to.query, data);
+			const result = runSearch(to.query, data).getElements();
```
EagleoutIce, line 221: maybe we should call this `unpack` in the future?

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+	range: SourceRange
+}
+
+export interface DeadCodeConfig extends MergeableRecord {
+	analyzeDeadCode: boolean
+}
+
+export const DEAD_CODE = {
+	createSearch: (config) => Q.all().with(Enrichment.CfgInformation, {
+		simplificationPasses: [...DefaultCfgSimplificationOrder, ...(config.analyzeDeadCode ? ['analyze-dead-code' as CfgSimplificationPassName] : [])]
```
EagleoutIce: as const should be more robust

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+			results: combineRanges(
+				elements.getElements()
+					.filter(element => {
+						const cfgInformation = enrichmentContent(element, Enrichment.CfgInformation);
+						return !cfgInformation.isReachable;
+					})
+					.map(element => element.node.info.fullRange ?? element.node.location)
+					.filter(isNotUndefined))
+				.map(range => ({
+					certainty: LintingCertainty.Definitely,
```
EagleoutIce, line 48: How do we define definitely here

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+					range
+				})),
+			'.meta': undefined as never
```
EagleoutIce: No? Not even the considered nodes?

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+		};
+	},
+	prettyPrint: {
+		[LintingPrettyPrintContext.Query]: result => `Code at ${formatRange(result.range)}`,
+		[LintingPrettyPrintContext.Full]:  result => `Code at ${formatRange(result.range)} will never be executed`,
```
EagleoutIce: can vs. will?

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+	},
+	info: {
+		name:          'Dead Code',
+		tags:          [LintingRuleTag.Smell, LintingRuleTag.Usability, LintingRuleTag.Reproducibility],
+		description:   'Marks areas of code that are never reached or whose invocation has no effect.',
```
EagleoutIce: Ich würde die mit no effect in ne andre linting Regel packen

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+		defaultConfig: {
+			analyzeDeadCode: true
+		}
+	}
+} as const satisfies LintingRule<DeadCodeResult, never, DeadCodeConfig>;
+
+function combineRanges(ranges: SourceRange[]): SourceRange[] {
```
EagleoutIce: Das ist doch was generell hilfreiches

src/linter/rules/file-path-validity.ts
```diff
@@ -56,9 +53,9 @@ export const FILE_PATH_VALIDITY = {
 		ignoreDefaultFunctions: true,
 		readFunctions:          ReadFunctions.concat(config.additionalReadFunctions),
 		writeFunctions:         WriteFunctions.concat(config.additionalWriteFunctions)
-	}),
+	}).with(Enrichment.CfgInformation, { analyzeDeadCode: false, checkReachable: false }),
```
EagleoutIce: Shouldn't all false be the default?

EagleoutIce: \+ remember to discuss how updates in later rules changed previous ones

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
+import { LintingCertainty, LintingPrettyPrintContext, type LintingResult, type LintingRule } from '../linter-format';
+import type { SourceRange } from '../../util/range';
+import {    rangeIsSubsetOf } from '../../util/range';
+
```
EagleoutIce: :notes: I'm spacing in, I'm spacing out....

src/linter/rules/dead-code.ts
```diff
@@ -0,0 +1,60 @@
...
+import { isNotUndefined } from '../../util/assert';
+import type { CfgSimplificationPassName } from '../../control-flow/cfg-simplification';
+import { DefaultCfgSimplificationOrder } from '../../control-flow/cfg-simplification';
+
+export interface DeadCodeResult extends LintingResult {
+	range: SourceRange
+}
+
+export interface DeadCodeConfig extends MergeableRecord {
+	analyzeDeadCode: boolean
```
EagleoutIce: maybe just allow the same simplification passes as `extractCFG`?

src/search/flowr-search.ts
```diff
@@ -71,7 +64,9 @@ export type FlowrSearchInput<
 
 /** Intentionally, we abstract away from an array to avoid the use of conventional typescript operations */
 export class FlowrSearchElements<Info = NoInfo, Elements extends FlowrSearchElement<Info>[] = FlowrSearchElement<Info>[]> {
-	private elements: Elements = [] as unknown as Elements;
+
```
EagleoutIce: Space me one more time

src/search/search-executor/search-enrichers.ts
```diff
@@ -51,17 +54,58 @@ export interface CallTargetsContent extends MergeableRecord {
...
 	linkedIds: FlowrSearchElement<ParentInformation>[]
 }
 
+export interface CfgInformationElementContent extends MergeableRecord {
+	/**
+	 * Whether the current node is a root node in the CFG.
+	 */
+	isRoot:       boolean
+	/**
+	 * Whether the current node is reachable from the root of the CFG.
```
EagleoutIce, line 68: not root -> entry point

src/search/search-executor/search-enrichers.ts
```diff
@@ -51,17 +54,58 @@ export interface CallTargetsContent extends MergeableRecord {
...
 	targets: (FlowrSearchElement<ParentInformation> | Identifier)[];
 }
+
 export interface LastCallContent extends MergeableRecord {
 	linkedIds: FlowrSearchElement<ParentInformation>[]
 }
 
+export interface CfgInformationElementContent extends MergeableRecord {
+	/**
+	 * Whether the current node is a root node in the CFG.
```
EagleoutIce: more not in a function definition

src/search/search-executor/search-enrichers.ts
```diff
@@ -126,6 +170,47 @@ export const Enrichments = {
...
+				return prev;
+			}
+
+			const content: CfgInformationSearchContent = {
+				...prev,
+				cfg: extractCfg(data.normalize, data.config, data.dataflow.graph, args.simplificationPasses),
+			};
+			if(args.checkReachable) {
+				const reachable = new Set<NodeId>();
+				visitCfgInOrder(content.cfg.graph, content.cfg.entryPoints, node => {
```
EagleoutIce: Please outsource into a general purpose function this should be used in the dead code removal

src/search/search-executor/search-generators.ts
```diff
@@ -16,6 +15,7 @@ import type { Query, SupportedQuery } from '../../queries/query';
 import { executeQueries, SupportedQueries } from '../../queries/query';
 import type { BaseQueryResult } from '../../queries/base-query-format';
 import type { RNode } from '../../r-bridge/lang-4.x/ast/model/model';
+import { enrichElement ,   Enrichment } from './search-enrichers';
```
EagleoutIce: :music: everybody step to the right

src/search/search-executor/search-generators.ts
```diff
@@ -90,23 +90,29 @@ function generateFrom(data: FlowrSearchInput<Pipeline>, args: { from: FlowrSearc
...
+			nodes.add({ node: data.normalize.idMap.get(node) as RNode<ParentInformation> });
 		}
+		nodesByQuery.set(query as Query['type'], nodes);
 	}
-	return new FlowrSearchElements([...nodes]);
+
+	// enrich elements with query data
+	const elements = new FlowrSearchElements([...nodesByQuery].flatMap(([_, nodes]) => [...nodes])).enrich(data, Enrichment.QueryData, { queries: result });
+	return elements.mutate(s => s.map(e => {
+		const query = [...nodesByQuery].find(([_, nodes]) => nodes.has(e))?.[0] as Query['type'];
```
EagleoutIce: maybe something more readable than `?.[0]` ?

src/util/range.ts
```diff
@@ -107,5 +107,5 @@ export function rangeCompare([r1sl,r1sc,,]: SourceRange, [r2sl,r2sc,,]: SourceRa
  * Checks if the first range is a subset of the second range.
  */
 export function rangeIsSubsetOf([r1sl,r1sc,r1el,r1ec]: SourceRange, [r2sl,r2sc,r2el,r2ec]: SourceRange): boolean {
-	return r1sl >= r2sl && r1sc >= r2sc && r1el <= r2el && r1ec <= r2ec;
-}
\ No newline at end of file
+	return (r1sl > r2sl || r1sl === r2sl && r1sc >= r2sc) && (r1el < r2el || r1sl === r2sl && r1ec <= r2ec);
```
EagleoutIce, line 110: performance will kill us - why

test/functionality/util/range.test.ts
```diff
@@ -2,11 +2,14 @@ import type {
 	SourceRange
 } from '../../../src/util/range';
 import {
+	rangeIsSubsetOf
+	,
```
EagleoutIce: :notes: down down shake it up

src/util/range.ts
Ellpeck: or to be much clearer about it: yes, > and === could just be combined into >= here

src/linter/rules/dead-code.ts
Ellpeck: what do u mean

EagleoutIce: why is the certainty definitly for an approximative dead code analysis

src/search/search-executor/search-enrichers.ts
Ellpeck: Wouldn't an entry point check check for `cfg.entryPoints` rather than `cfg.graph.rootIds()`? Or are those the same? I'm confused

EagleoutIce: if we determine whether it is reachable by a *root* id this would effectively determine anyhting as reachable, you start from the entry points: 

https://github.com/flowr-analysis/flowr/pull/1785/files#diff-eb1cfba6ab3786fadfcfcde1742894626f94dcbff063b38665ab7a9e4a6b52c7R201

src/util/range.ts
Ellpeck: Wait, can this be simplified? Because if the start line is *equal*, it has to also match the column, but if it's *greater*, it doesn't have to match the column :(

src/linter/rules/dead-code.ts
Ellpeck: should it just always be maybe then? i dont rly know ngl

EagleoutIce: same - that was why I asked for the definitely property I think we should use definitely only when we are sure that the automatic application of the quickfix cannot change the program semantics (if a quickfix is there) and if there is no quickfix then whenever the linting result is not affected by unknown side effects etc. Hence the "maybe" case should be the default and anything else should only be there if we know more - how do you like that?

Ellpeck: that sounds good tbh. i'll still make a note to discuss this more (both us & in the thesis) though :)

src/control-flow/cfg-simplification.ts
```diff
@@ -76,3 +73,11 @@ function uniqueControlFlowSets(cfg: ControlFlowInformation, _info?: CfgPassInfo)
 function toBasicBlocks(cfg: ControlFlowInformation, _info?: CfgPassInfo): ControlFlowInformation {
 	return convertCfgToBasicBlocks(cfg);
 }
+
+export function cfgFindAllReachable(cfg: ControlFlowInformation): Set<NodeId> {
```
EagleoutIce, line 81: Maybe some tsdoc?

src/search/flowr-search.ts
```diff
@@ -97,4 +91,19 @@ export class FlowrSearchElements<Info = NoInfo, Elements extends FlowrSearchElem
 		this.elements = mutator(this.elements);
 		return this;
 	}
+
+	public enrich<E extends Enrichment>(data: FlowrSearchInput<Pipeline>, enrichment: E, args?: EnrichmentSearchArguments<E>): this {
```
EagleoutIce, line 101: I think some documentation here would be nice as well

src/util/range.ts
EagleoutIce: no it is fine this way

test/functionality/linter/lint-dead-code.test.ts
```diff
@@ -0,0 +1,44 @@
...
+			assertLinter('TRUE FALSE', parser, 'if(TRUE) 1 else if (FALSE) 2 else 3', 'dead-code', [
+				{ certainty: LintingCertainty.Definitely, range: [1, 17, 1, 35] }
+			]);
+			assertLinter('FALSE FALSE', parser, 'if(FALSE) 1 else if (FALSE) 2 else 3', 'dead-code', [
+				{ certainty: LintingCertainty.Definitely, range: [1, 11, 1, 11] },
+				{ certainty: LintingCertainty.Definitely, range: [1, 29, 1, 29] }
+			]);
+			assertLinter('FALSE TRUE', parser, 'if(FALSE) 1 else if (TRUE) 2 else 3', 'dead-code', [
+				{ certainty: LintingCertainty.Definitely, range: [1, 11, 1, 11] },
+				{ certainty: LintingCertainty.Definitely, range: [1, 35, 1, 35] }
```
EagleoutIce, line 48: could you please add another test with an infinite repeat loop? similarly with while?

test/functionality/search/search-line.test.ts
```diff
@@ -89,5 +91,35 @@ describe('flowR search', withTreeSitter(parser => {
...
+			assertSearch('reachable never', parser, 'if(FALSE) 1 else 2', ['1@if', '1@FALSE', '1@2', '$4', '$6'], Q.all().with(Enrichment.CfgInformation, cfgArgs).filter({
+				name: FlowrFilter.MatchesEnrichment, args: {
+					enrichment: Enrichment.CfgInformation,
+					test:       /"isReachable":true/
+				}
+			}));
+			assertSearch('reachable no dead code', parser, 'if(FALSE) 1 else 2', [], Q.all().with(Enrichment.CfgInformation).filter({
+				name: FlowrFilter.MatchesEnrichment, args: {
+					enrichment: Enrichment.CfgInformation,
+					test:       /"isReachable":false/
```
EagleoutIce, line 114: wild xD

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.4.0 (see [Release v2.4.0 (Forward-Slicing, Minor Fixes)](https://github.com/flowr-analysis/flowr/releases/tag/v2.4.0))._

PR #1787, opened 2025-06-28
Add more tests to verify capabilities of shape inference

OliverGerstl:


src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -16,6 +16,10 @@ interface DataFrameInfoBase {
 	domain?: DataFrameStateDomain
 }
 
+export interface DataFrameUnassignedInfo {
```
EagleoutIce, line 22: We should add a comment to clarify whether this refers to completely unbound data frames, anonymous data frames, etc.

src/abstract-interpretation/data-frame/absint-info.ts
```diff
@@ -16,6 +16,10 @@ interface DataFrameInfoBase {
...
+	type: 'unassigned'
+}
+
 export interface DataFrameAssignmentInfo {
 	type:       'assignment',
```
EagleoutIce, line 30: In general a short overview for the scenarios in which the respective infos apply

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -57,22 +57,25 @@ class DataFrameAbsintVisitor<
...
-			this.newDomain = wideningDataFrameStates(this.oldDomain, this.newDomain);
-		}
 		return visitedCount === 0 || !equalDataFrameState(this.oldDomain, this.newDomain);
 	}
-	
+
 	protected override visitDataflowNode(vertex: Exclude<CfgSimpleVertex, CfgBasicBlockVertex>): void {
 		const node = this.getNormalizedAst(isMarkerVertex(vertex) ? vertex.root : vertex.id);
 
 		if(node !== undefined) {
```
EagleoutIce: We could invert the guard to avoid the added nesting

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -57,22 +57,25 @@ class DataFrameAbsintVisitor<
...
 			this.oldDomain = node.info.dataFrame?.domain ?? new Map<NodeId, DataFrameDomain>();
 			super.visitDataflowNode(vertex);
-			node.info.dataFrame ??= {};
+
+			if((this.visited.get(vertex.id) ?? 0) >= (this.config.wideningThreshold ?? DefaultWideningThreshold)) {
```
EagleoutIce: maybe outsource as a `shouldWiden()`?

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -57,22 +57,25 @@ class DataFrameAbsintVisitor<
...
+				this.newDomain = wideningDataFrameStates(this.oldDomain, this.newDomain);
+			}
+			// mark variable definitions as "unassigned", since the evaluation of the assigned value is delayed until processing the assignment
+			node.info.dataFrame ??= isVariableDefinitionVertex(this.getDataflowGraph(vertex.id)) ? { type: 'unassigned' } : {};
```
EagleoutIce: Conceptually, shouldn't we stop here if widened? I mean yes we can continue in the future, but it feels, weird to make a default unassigned set if widening applies

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -125,7 +128,8 @@ class DataFrameAbsintVisitor<
 				const identifier = this.getNormalizedAst(node.info.dataFrame.identifier);
 
 				if(identifier !== undefined) {
-					identifier.info.dataFrame ??= {};
+					// remove "unassigned" markers from variable definitons and update domain
+					identifier.info.dataFrame = {};
```
EagleoutIce: I mean so far this is fine, but in general overwriting the info feels... wrong? Maybe a `if(hasUnassignedInfo(identifier)) clearExistingDomainInfo(identifier)`? In that case we can also clear the comments

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -142,11 +146,10 @@ class DataFrameAbsintVisitor<
 					getOriginInDfg(this.config.dfg, operation.operand)
 						?.filter(origin => origin.type === OriginType.ReadVariableOrigin)
 						.forEach(origin => this.newDomain.set(origin.id, value));
+				} else if(getConstraintType(operation.operation) === ConstraintType.ResultPostcondition) {
+					this.newDomain.set(node.info.id, value);
```
EagleoutIce, line 152: Overwriting on a post condition is fine, but we should discuss cases in which the assignment is, e.g., delayed

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -214,9 +217,11 @@ export function resolveIdToAbstractValue(
 	if(node.type === RType.Symbol) {
 		const values = getOriginInDfg(dfg, node.info.id)
 			?.filter(entry => entry.type === OriginType.ReadVariableOrigin)
-			.map(entry => resolveIdToAbstractValue(entry.id, dfg, domain));
+			.map<RNode<ParentInformation & AbstractInterpretationInfo> | undefined>(entry => dfg.idMap?.get(entry.id))
+			.filter(entry => entry?.info.dataFrame?.domain !== undefined && entry.info.dataFrame.type !== 'unassigned')
```
EagleoutIce: again, maybe outsource this using a semantic name?

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -214,9 +217,11 @@ export function resolveIdToAbstractValue(
...
+			.map(entry => resolveIdToAbstractValue(entry, dfg, domain));
 
-		if(values?.every(isNotUndefined)) {
+		if(values !== undefined && values.length > 0 && values.every(isNotUndefined)) {
```
EagleoutIce: why is the `> 0` logic not part of `join`?

test/functionality/abstract-interpretation/data-frame/data-frame.ts
```diff
@@ -28,47 +30,121 @@ export enum DomainMatchingType {
...
+ *
+ * @param shell    - The R shell to use to run the code
+ * @param code     - The code to test
+ * @param criteria - The slicing criteria to test including the expected shape constraints and the {@link DataFrameTestOptions} for each criterion (defaults to {@link DataFrameTestExact})
+ * @param skipRun  - Whether the real test with the execution of the R code should be skipped (defaults to `false`)
+ * @param parser   - The parser to use for the data flow graph creation (defaults to the R shell)
+ * @param name     - An optional name or test label for the test (defaults to the code)
+ * @param config   - The config to use for the test (defaults to {@link defaultConfigOptions})
+ */
+export function testDataFrameDomain(
```
EagleoutIce, line 87: Suggestion: what do you think of a factory class? Starting with your data frame tests but then for... all tests effectively. 
In the tests we then would write something like:

`withDomainDataFrame(tester => ...)`

and then within we would write `tester(code...)`

this 1) would abstract away from the `RShell`/Known Parser in the background
2) this could incporporate the required `describe.sequential` within the `tester` logic, 
3) `tester.<autocomplete>` provides you with all sensible testing methods
4) `withDomainDataFrame` (or whatever we name it) could accept default configs, etc.

test/functionality/abstract-interpretation/data-frame/data-frame.ts
```diff
@@ -80,150 +156,231 @@ export function assertDataFrameDomain(
...
-			return value as DataFrameDomain[K];
-		}
-		case 'cols':
-		case 'rows': {
-			const value = Number.parseInt(line);
-			return [value, value] as DataFrameDomain[K];
-		}
-		default:
-			assertUnreachable(type);
+	const OutputRegex = /(?<=^|,)(?:\[([^\]]*)\]|([^,]*))/g;
```
EagleoutIce: is this copy pasted from somewhere? it looks fragile

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -58,237 +41,124 @@ describe.sequential('Data Frame Abstract Interpretation', withShell(shell => {
...
-			source => `df <- readr::read_csv(${source}, comment = "#")`,
-			[['1@df', { colnames: ColNamesTop, cols: [3, 3], rows: [5, 5] }, { colnames: DomainMatchingType.Overapproximation }]]
+		testDataFrameDomain(
+			shell,
+			`
+\`df1\` <- data.frame(id = 1:5)
+'df2' <- data.frame(id = 1:5)
+"df3" <- data.frame(id = 1:5)
+df <- cbind(df1, df2, df3)
+			`.trim(),
```
EagleoutIce, line 79: coming back to the tester etc. I think trimming is something that should happen automatically (configurable)

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -58,237 +41,124 @@ describe.sequential('Data Frame Abstract Interpretation', withShell(shell => {
...
-			source => `df <- readr::read_csv(${source}, col_names = FALSE, skip = 4)`,
-			[['1@df', { colnames: ColNamesTop, cols: [3, 3], rows: [5, 5] }, { colnames: DomainMatchingType.Overapproximation }]]
+		testDataFrameDomain(
+			shell,
+			`
+df1 <- data.frame(id = 1:5)
+df2 <- df1
+			`.trim(),
+			[
+				['1@df1', { colnames: ['id'], cols: [1, 1], rows: [5, 5] }],
```
EagleoutIce, line 94: why isn't this another object? e.g.:

`{ '1@df1': { colnames....} }` this way, we would get an automatic check for testing the same criteria twice (at least if given in the same format)

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -326,838 +214,3432 @@ while (TRUE) {
...
+			shell,
+			'df <- as.data.frame(optional = TRUE, fix.empty.names = FALSE, x = data.frame(id = 1:3, name = 4:6))',
+			[['1@df', { colnames: ['id', 'name'], cols: [2, 2], rows: [3, 3] }]]
+		);
+	});
+
+	describe('Read', () => {
+		testDataFrameDomainWithSource(
+			shell,
+			'"a.csv"', `text = "${getFileContent('a.csv')}"`,
```
EagleoutIce, line 407: the tester object could take in the `getFileContent` function as higher-order and then abstract this away.

It fels kind of weird to: 1) pass nested quotes for the name 2) assume that the file does not contain nested `"` symbols (`JSON.stringify` would be safer here)

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -326,838 +214,3432 @@ while (TRUE) {
...
+			'"e.csv"', `text = "${getFileContent('e.csv')}"`,
+			source => `df <- read.table(${source}, header = TRUE)`,
+			[['1@df', { colnames: ['first', 'last', 'state', 'phone'], cols: [4, 4], rows: [3, 3] }]]
+		);
+
+		testDataFrameDomainWithSource(
+			shell,
+			'"f.csv"', `text = "${getFileContent('f.csv')}"`,
+			source => `df <- read.delim(${source})`,
+			[['1@df', { colnames: ColNamesTop, cols: [4, 4], rows: [3, 3] }, { colnames: DomainMatchingType.Overapproximation }]]
```
EagleoutIce, line 479: i tend to prefer `OverapproximateColumns` or something as object here that can be used repeatedly. The `DomainMatchingType` enum for this is nice but it feels like a break of reading flow

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -326,838 +214,3432 @@ while (TRUE) {
...
+			testDataFrameDomain(
+				shell,
+				`
+df1 <- data.frame(id = 1:4, name = "A", score = c(80, 75, 90, 70))
+df2 <- data.frame(id = 1:6, name = "B", category = c("A", "B", "B", "A", "C", "B"))
+df <- merge(df1, df2, "id", no.dups = FALSE)
+				`.trim(),
+				[
+					['3@df', { colnames: ColNamesTop, cols: [5, 5], rows: [4, 4] }, { colnames: DomainMatchingType.Overapproximation }]
+				]
```
EagleoutIce, line 3567: Why, aren't these tests, e.g., in a for loop to highlight that the only things that change are the arguments, and the ranges? repeating the code over and over (with slight permutations) - at least imho - makes this much harder to read

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: The "unassigned" marker is only set if there was no data frame information before (`??=`), i.e. the vertex has not been visited yet. When we have widened, the condition `equalDataFrameState` in `visitNode` will be true in the next iteration and we will stop

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: We are just setting the abstract value for the operation itself, assignments are handled separately

src/abstract-interpretation/data-frame/absint-visitor.ts
OliverGerstl: The `join` always returns an abstract data frame value (`Top` if no arguments provided). But if there is no origin that represents a data frame, we do not want to infer an abstract data frame value, i.e. we return `undefined`

test/functionality/abstract-interpretation/data-frame/data-frame.ts
OliverGerstl: Yes, that is something that we can discuss. But this is probably out of scope for the thesis

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: The quotes are escaped inside the `getFileContent` function

test/functionality/abstract-interpretation/data-frame/data-frame.ts
OliverGerstl: I added this to #1790

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: I added this to #1790

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: I added this to #1790

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: I added this to #1790

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: I added this to #1790

test/functionality/abstract-interpretation/data-frame/data-frame.ts
OliverGerstl: This is the regex used for parsing CSV entries. But I changed it to the regex that is used in the evaluation

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -88,6 +90,7 @@ class DataFrameAbsintVisitor<
 		if(node !== undefined && (targetNode?.type === RType.Symbol || targetNode?.type === RType.String) && sourceNode !== undefined) {
```
EagleoutIce: `assignsToSymbol`?

src/abstract-interpretation/data-frame/absint-visitor.ts
EagleoutIce: Hmmm ok, in that case i would again use a helper like `hasValues`?

PR #1795, opened 2025-07-04
Fix remaining special edge cases of wrong over-approximation

OliverGerstl:


src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,261 @@
...
+export function escapeRegExp(text: string, allowTokens: boolean = false): string {
+	if(allowTokens) {
+		return text.replace(/[.*+?^${}()|[\]]/g, '\\$&');
+	} else {
+		return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+	}
+}
+
+/**
+ * Maps all invalid, duplictae or empty column names to top depending on the provided arguments.
```
EagleoutIce:
```suggestion
 * Maps all invalid, duplicate or empty column names to top depending on the provided arguments.
```

src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -90,47 +85,72 @@ function mapDataFrameIndexColRowAccess(
...
+			}
 			operand = undefined;
 		}
 		if(colArg !== undefined && colArg !== EmptyArgument) {
-			result.push({
-				operation: columns === undefined || columns.every(col => typeof col === 'string' || col >= 0) ? 'subsetCols' : 'removeCols',
-				operand:   operand?.info.id,
-				args:      { colnames: columns?.map(col => typeof col === 'string' ? col : undefined) }
-			});
+			if(columns === undefined || columns.every(col => typeof col === 'string' || col >= 0)) {
```
EagleoutIce: maybe an "allColumnsKnown"?

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,261 @@
...
+ * @param argument - The specification of the argument to get
+ * @param info     - Argument resolve information
+ * @returns An argument matching the specified `argument` or `undefined`
+ */
+export function getFunctionArgument(
+	args: readonly RFunctionArgument<ParentInformation>[],
+	argument: FunctionParameterLocation<unknown> | string,
+	info: ResolveInfo
+): RFunctionArgument<ParentInformation> | undefined {
+	const pos = typeof argument !== 'string' ? argument.pos : -1;
```
EagleoutIce, line 113: please check against this being the non-empty argument

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -0,0 +1,261 @@
...
+	const name = typeof argument !== 'string' ? argument.name : argument;
+	let arg = undefined;
+
+	if(name !== undefined) {
+		arg = args.find(arg => resolveIdToArgName(arg, info) === name);
+	}
+	if(arg === undefined && pos >= 0 && pos < args.length && args[pos] !== EmptyArgument && args[pos].name === undefined) {
```
EagleoutIce: This check should be outsourced with a semantic name

EagleoutIce (approved):
Very hard to check for correctness, but let's see how this goes

src/abstract-interpretation/data-frame/mappers/arguments.ts
OliverGerstl: The `argument` here is no actual function call argument but the location/name specification for a function argument/parameter

PR #1798, opened 2025-07-06
Seed CLI flag for limiting the number of benchmarked files

ydq23:


src/cli/benchmark-app.ts
```diff
@@ -84,8 +86,9 @@ async function benchmark() {
 
 	if(options.limit) {
 		log.info(`limiting to ${options.limit} files`);
+		const random = options.seed ? seedrandom(options.seed) : Math.random;
 		// shuffle and limit
-		files.sort(() => Math.random() - 0.5);
+		files.sort(() => random() - 0.5);
```
EagleoutIce, line 91: There is another level of random that this should respect - within the benchmark slicer we sample for slicing points in case of a limit and no equidistant sampling. Can you use the same sampling there?

OliverGerstl: Done :)

EagleoutIce (approved):

EagleoutIce:
Recreating a seed generator feels a little bit like cheating but hey :D

Issue #1752, opened 2025-06-15
Restructure the abstract domains for data frame shapes


OliverGerstl:
Restructure the abstract domains for the data frame shape inference by

- generalizing the column names domain as constant string set domain
- renaming the current interval domain to positive interval domain
- using classes for abstract domains to prevent many join, meet, and widening functions

PR #1801, opened 2025-07-07
Add config options for data frame shape inference

OliverGerstl:
Adding configuration options for the maximum number of inferred column names, the widening threshold, disabling reading loaded data files, and the maximum number of lines to read from loaded data files


src/abstract-interpretation/data-frame/domain.ts
```diff
@@ -1,7 +1,8 @@
+import { defaultConfigOptions } from '../../config';
 import type { NodeId } from '../../r-bridge/lang-4.x/ast/model/processing/node-id';
 import { setEquals } from '../../util/collections/set';
 
-const MaxColNames = 50;
+const MaxColNames = defaultConfigOptions.abstractInterpretation.dataFrame.maxColNames;
```
EagleoutIce, line 5: this should not happen "on module load", as this, for example, will ignore requested config overwrites, but honestly, for a default.. it is fine i guess

src/config.ts
```diff
@@ -261,7 +295,15 @@ export const flowrConfigFileSchema = Joi.object({
 		slicer: Joi.object({
 			threshold: Joi.number().optional().description('The maximum number of iterations to perform on a single function call during slicing.')
 		}).optional().description('The configuration for the slicer.')
-	}).description('How to resolve constants, constraints, cells, ...')
+	}).description('How to resolve constants, constraints, cells, ...'),
+	abstractInterpretation: Joi.object({
+		dataFrame: Joi.object({
+			maxColNames:       Joi.number().min(1).description('The maximum number of columns names to infer for data frames before over-approximating the column names to top.'),
```
EagleoutIce: Why not allow 0 to "disable"?

src/config.ts
```diff
@@ -155,6 +155,32 @@ export interface FlowrConfigOptions extends MergeableRecord {
...
+			 */
+			readonly maxColNames:       number;
+			/**
+			 * The threshold for the number of visitations of a node at which widening should be performed to ensure the termination of the fixpoint iteration
+			 */
+			readonly wideningThreshold: number;
+			/**
+			 * Whether data frame shapes should be extracted from loaded external files, such as CSV files
+			 */
+			readonly readLoadedData:    boolean;
```
EagleoutIce: "includeExternalDatafiles"?

src/config.ts
```diff
@@ -155,6 +155,32 @@ export interface FlowrConfigOptions extends MergeableRecord {
...
+			/**
+			 * The maximum number of lines to read when extracting data frame shapes from loaded files, such as CSV files
+			 */
+			readonly maxReadLines:      number;
```
EagleoutIce: shouldn't this be more of a "child property" of `readLoadedData`?

src/abstract-interpretation/data-frame/domain.ts
OliverGerstl: Yes, this is just a temporary solution until the new structure for the abstract domains of #1752 is applied that already includes support for these config options

src/config.ts
OliverGerstl: Maybe rather something like "readLoadedDataFiles", as this should indicate that we are *reading* the data frame shapes from *loaded* data files

PR #1809, opened 2025-07-16
test(slicing): move params

MaxAtoms:
We wanted to consolidate the parameters in `assertSliced`, right?
Let me know if you see other quick-win refactoring opportunities here 🙂


test/functionality/_helper/shell.ts
```diff
@@ -453,19 +453,18 @@ export function assertSliced(
 	input: string,
 	criteria: SlicingCriteria,
 	expected: string,
-	userConfig?: Partial<TestConfigurationWithOutput> & { autoSelectIf?: AutoSelectPredicate, skipTreeSitter?: boolean, skipCompare?: boolean, cfgExcludeProperties?: readonly CfgProperty[], flowrConfig?: FlowrConfigOptions },
-	testCaseFailType?: TestCaseFailType,
-	getId: () => IdGenerator<NoInfo> = () => deterministicCountingIdGenerator(0)
+	userConfig?: Partial<TestConfigurationWithOutput> & { autoSelectIf?: AutoSelectPredicate, skipTreeSitter?: boolean, skipCompare?: boolean, cfgExcludeProperties?: readonly CfgProperty[], flowrConfig?: FlowrConfigOptions, testCaseFailType?: TestCaseFailType, getId?: () => IdGenerator<NoInfo> },
```
EagleoutIce: Could we additionally take all the stuff from the intersection (`{...}`) and move it into an interface so that we can comment on the meaning of each individual property? Ideally the interface does not use `|undefined` and then we wrap it within a `Partial` similar to the `TestConfigurationWithOutput`

EagleoutIce (commented):
Thank you, this makes it much more readable imho

test/functionality/_helper/shell.ts
MaxAtoms: Done ✅ Feel free to adjust naming and comments as needed

test/functionality/_helper/shell.ts
```diff
@@ -447,60 +447,76 @@ function testWrapper(skip: boolean | undefined, shouldFail: boolean, testName: s
 
 export type TestCaseFailType = 'fail-shell' | 'fail-tree-sitter' | 'fail-both' | undefined;
```
MaxAtoms, line 448: Not to become too yak shavy, but we only seem to use `TestCaseFailType` `fail-both`. Do you suspect that we will ever need the other two cases or could we simplify this to a boolean, @EagleoutIce?

EagleoutIce: No addressing this is perfectly fine :D 

For now we only have `fail-both` and in ideal future this is what probably suffices. However, especially during the first integration of tree-sitter but even now from time to time we find some bugs in the tree-sitter-r grammar (or vice versa) and henceforth encounter cases that do not work with a specific engine (once the bug was on our side but the fix was more complicated so I used the flag to defer the problem to another PR). So for now I would leave them in if this is fine for you too

PR #1815, opened 2025-07-18
Prepare the second evaluation of data frame shape inference

OliverGerstl:
This fixes some issues found during the second evaluation of the data frame shape inference

### DIFF
--- a/src/abstract-interpretation/data-frame/mappers/function-mapper.ts
+++ b/src/abstract-interpretation/data-frame/mappers/function-mapper.ts
@@ -582,11 +582,10 @@ export function mapDataFrameFunctionCall<Name extends DataFrameFunction>(
 		if(isDataFrameFunction(node.functionName.content)) {
 			const functionName = node.functionName.content as Name;
 			const mapper = DataFrameFunctionMapper[functionName].mapper as DataFrameFunctionMapping<DataFrameFunctionParams<Name>>;
-			const params = DataFrameFunctionParamsMapper[functionName] as DataFrameFunctionParams<Name>;
+			const params = DataFrameFunctionParamsMapper[functionName] as DataFrameFunctionParams<Name> & { critical?: FunctionParameterLocation<unknown>[] };
 			const args = getFunctionArguments(node, dfg);
-			const critical = (params as { critical?: FunctionParameterLocation<unknown>[] }).critical;
 
-			if(hasCriticalArgument(args, critical, resolveInfo)) {
+			if(hasCriticalArgument(args, params.critical, resolveInfo)) {
 				operations = [{ operation: 'unknown', operand: undefined }];
 			} else {
 				operations = mapper(args, params, resolveInfo, config);


src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -38,7 +38,7 @@ export interface FunctionParameterLocation<T = undefined> {
  */
 export function escapeRegExp(text: string, allowTokens: boolean = false): string {
 	if(allowTokens) {
-		return text.replace(/[.*+?^${}()|[\]]/g, '\\$&');
+		return text.replace(/[.*+?^${}()|[\]]/g, '\\$&').replace(/\\[^stn]/g, '\\$&');
```
EagleoutIce: This should become a "global helper function" in the future - replacing twice feels so painful

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -67,6 +67,11 @@ const DataFrameFunctionMapper = {
 	'arrange':       { mapper: mapDataFrameIdentity, library: 'dplyr' }
 } as const satisfies Record<string, DataFrameFunctionMapperInfo<never>>;
 
+/**
+ * List of data frame functions that are not explicitly supported but return data frames.
+ */
+const UnsupportedDataFrameFunctions = ['anova', 'lmer', 'Anova', 'Manova', 'AIC', 'BIC'] as const satisfies string[];
```
EagleoutIce: Why not move this into a flowR config so that it can be configured and changed if someone 1) finds new functions or 2) wants to try anyway

src/util/files.ts
```diff
@@ -127,7 +127,7 @@ export async function readLineByLine(filePath: string, onLine: (line: Buffer, li
  * See {@link readLineByLine} for an asynchronous version.
  */
 export function readLineByLineSync(filePath: string, onLine: (line: Buffer, lineNumber: number) => void, maxLines: number = Infinity): boolean {
-	if(!fs.existsSync(filePath)) {
+	if(!fs.existsSync(filePath) || !fs.statSync(filePath).isFile()) {
```
EagleoutIce: `statSync` should include `existsSync` (it is just slightly more expensive) so we can drop the first part!
```suggestion
	if(!fs.statSync(filePath).isFile()) {
```

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: Additionally, we may want to add package information for the future

src/abstract-interpretation/data-frame/mappers/arguments.ts
OliverGerstl: This is just to ensure that only the tokens `\s`, `\t`, and `\n` should be kept when escaping the regex with the option `allowTokens` (for sanitization)

src/util/files.ts
OliverGerstl: The `fs.statSync` function throws an error "ENOENT: no such file or directory" if the file does not exist. But I could, of course, wrap the function in a try-catch block

EagleoutIce: I would prefer the try catch to avoid race conditions if the file vanishes

src/abstract-interpretation/data-frame/mappers/arguments.ts
EagleoutIce: Yes i know but is there a reason this does not happen within the first regex?

OliverGerstl: No, I included the second replacement into the first regex :)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: I restructured the list of unsupported data frame functions to make them more flexible and document the packages of each function, and added some more unsupported data frame functions.
I also created an issue #1823 for adding this list to the flowR config for customization :)

src/util/files.ts
OliverGerstl: Oh yes, you are right. I found an option to disable throwing an error in this case and simplified the check to a single call to `statSync`

PR #1922, opened 2025-08-29
Fixing cds and single loop detection

EagleoutIce:


EagleoutIce:
this spawns https://github.com/flowr-analysis/flowr/issues/1923

PR #1940, opened 2025-09-12
Add analyzer to Wiki

MaxAtoms:


src/documentation/print-dataflow-graph-wiki.ts
```diff
@@ -862,14 +862,9 @@ ${details('Example: While-Loop Body', await printDfGraphForCode(shell, 'while(TR
...
-	const shell = new RShell();
-	const result = await new PipelineExecutor(DEFAULT_DATAFLOW_PIPELINE, {
-		parser:  shell,
-		request: requestFromInput('x <- 1\nx + 1')
-	}, defaultConfigOptions).allRemainingSteps();
-	shell.close();
-	return result;
+async function dummyDataflow(): Promise<DataflowInformation> {
+	const analyzer = await new FlowrAnalyzerBuilder(requestFromInput('x <- 1\nx + 1')).build();
+	return await analyzer.dataflow();
```
MaxAtoms: As discussed earlier, the script will not terminate if we do not explicitly create and close a RShell here, @EagleoutIce

PR #1941, opened 2025-09-16
Support for rmd notebooks 

gigalasr:


src/r-bridge/retriever.ts
```diff
@@ -135,11 +136,19 @@ export function retrieveParseDataFromRCode(request: RParseRequest, shell: RShell
...
-	/* call the function with the request */
-	const command =`flowr_get_ast(${request.request}=${JSON.stringify(
-		request.content
-	)}${suffix})`;
+
+	let command = '';
+	if(request.request === 'file' && !isNormalRFile(request.content)) {
+		command =`flowr_get_ast(text=${JSON.stringify(
+			readFileWithAdapter(request.content).code
+		)})`;
```
EagleoutIce: remember the suffix

src/util/formats/adapter.ts
```diff
@@ -0,0 +1,37 @@
+import type { RParseRequest } from '../../r-bridge/retriever';
+import { assertUnreachable } from '../assert';
+import type { FileAdapter, FileData } from './adapter-format';
+import { RAdapter } from './adapters/r-adapter';
+import { RmdAdapter } from './adapters/rmd-adapter';
+import path from 'path';
+
+export const FileAdapters = {
+	'.R':   RAdapter,
```
EagleoutIce: I would drop the `.` prefix to make it clear that we do not focus on the ending (i.e., if someone registers `Rxx` in the future we can map it to an existing adater

src/util/formats/adapters/r-adapter.ts
```diff
@@ -0,0 +1,11 @@
+import type { FileAdapter } from '../adapter-format';
+import fs from 'fs';
+
+export const RAdapter = {
+	readFile: (p: string) => {
+		return {
+			'type': '.R',
+			'code': fs.readFileSync(p, 'utf8')
```
EagleoutIce: I do not like that this already loads the file. instead of a `code` field, they should either offer a new request or at least only store the file path. For bigger files, serializing them this way breaks the R shell, for those we need `file=...`

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+		const frontmatter = matter(raw);
+
+		// Parse Codeblocks
+		const walker = ast.walker();
+		const blocks: CodeBlockEx[] = [];
+		let e;
+		while((e = walker.next())) {
+			const node = e.node;
+			if(node.type === 'code_block') {
+				if(!node.info ||
```
EagleoutIce: please outsource this guard into a helper function -> also all of those individual negations feel weird to me, especially because the `!node.info` is implicit with optional chaining with `!node.info?.startsWith...`

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+		// Parse Codeblocks
+		const walker = ast.walker();
+		const blocks: CodeBlockEx[] = [];
+		let e;
+		while((e = walker.next())) {
+			const node = e.node;
+			if(node.type === 'code_block') {
+				if(!node.info || 
+				   !node.literal ||
+				   !node.info.startsWith('{r') ||
```
EagleoutIce: isn't capital R allowed as well? 
additionally: this will trigger for: `{rust` won't it?

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+					continue;
+				}
+				
+				blocks.push({
+					code:     node.literal,
+					options:  parseOptions(node.info, node.literal),
+					startpos: { line: node.sourcepos[0][0] + 1, col: 0 }
+				});
+			} 
+			// else if(node.type === 'code') {
```
EagleoutIce: please remove it completely if not needed

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+	const goToLine = (n: number) => {
+		const diff = n - line;
+		guard(diff >= 0);
+		line += diff;
+		output += '\n'.repeat(diff);
+	};
+
+
+	const countNewlines = (str: string) => {
+		return str.split(/\r\n|\r|\n/).length - 1;
```
EagleoutIce: i think it is more performant to have the regex otsourced as a variable that is built once

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+	
+	const lines = content.split('\n');
+	for(const line of lines) {
+		if(!line.trim().startsWith('#|')) {
+			break;
+		}
+
+		const opt = line.substring(3);
+
+		opts += opts.length === 0 ? `${opt}` : `, ${opt}`;
```
EagleoutIce: the first `` `${opt}` `` should be just `opt`

EagleoutIce (commented):
very nice :)

src/r-bridge/retriever.ts
gigalasr: but for text there is no suffix, no?

src/util/formats/adapters/r-adapter.ts
gigalasr: Hmm yeah I see your point here. 
The way it is currently written, this path is only used by tree sitter. The RShell Parser is skipping the adapter, if the file is an R file.

Not really the cleanest design right now, but I heard we want to drop the RShell Parser anyways. What is your opinion?

Alternativly, I could move all the tree sitter / r shell parser stuff to this adapter. And reverse how stuff is called right now?

src/util/formats/adapter.ts
gigalasr: But we are using this to check the file ending?

src/util/formats/adapters/rmd-adapter.ts
gigalasr: We do need this for inline code blocks. But I wasn't sure how to implement them because they don't provide a sourcepos :/

src/r-bridge/retriever.ts
EagleoutIce: fair, but this is only for now, i weakly remember the necessity in the future so maybe still have the suffix toggle?

src/util/formats/adapter.ts
EagleoutIce: i think the `FileAdapters` object should link a document type to its adapter, not the file ending per-se because these can also be configured via MIME and we would want to respect this imho

src/util/formats/adapters/r-adapter.ts
EagleoutIce: We do not want to drop the `RShell` parser. It is not the default but it still has its merits. Additionally, especially with the planned addition of incremental analysis capabilities with the project frontend we need to keep the "file-pointer" as long as possible because otherwise this breaks

src/util/formats/adapters/rmd-adapter.ts
EagleoutIce: ok i think it is better to discuss this

src/util/formats/adapters/r-adapter.ts
gigalasr: So how should I proceed with this? 
I think it is also weird if it does not load the file, because 
a) The TreeSitter Parser needs the file directly
b) The RmdAdapter also loads the file directly

src/util/formats/adapter.ts
gigalasr: soo you want these keys to be mime types like `text/R` and `text/Rmd`?

EagleoutIce: i dont think we need the `text` prefix, but essentially yes - they more or less should indicate the semantics contained in the file and not necessarily anything dependent on the file name (we might also add guessing layer in the future that tries to detect file content)!

src/util/formats/adapters/r-adapter.ts
EagleoutIce: a) any parser has to *load* the file :D - this is the latest stage it can be loaded at and we try to move it to this as far as possible
b) ideally the adapter would be invoked right before parsing, but we cannot guaranee this (e.g., because after analyzing a single file we discover new files that we have to check first, ...) - so it is more of a "bad" thing that the RmD adapter loads the file instead of it being a good thing for the R adapter to not load it. For RmD this, in a way, cannot be really helped (for now) - but one advatage of tree-sitter syntaxes etc. is the incrementally which we really care about here -> if a file changes, the less we have to re-do the better (and with the Adapter loading the file we either have to replicate incremental re-loading there, or we have to always reload the complete file, then reidentify events, and then give these events to the incremental parser)

src/util/formats/adapters/rmd-adapter.ts
EagleoutIce: i understand that they do not contain a *sourcepos* but they should still contain a ref to their textual location shoudnt they?

EagleoutIce:
one additional thing i noticed, a simple step that we could also make configurable for this stage (or as a subsequent issue) is that each cell becomes its "own file" for flowR - this way we do not have to add additional logic to handle notebook specific execution semantics

src/benchmark/slicer.ts
```diff
@@ -180,7 +179,8 @@ export class BenchmarkSlicer {
 		await this.calculateStatsAfterInit(request);
 	}
 
-	private async calculateStatsAfterInit(request: RParseRequestFromFile | RParseRequestFromText) {
+	private async calculateStatsAfterInit(r: RParseRequestFromFile | RParseRequestFromText) {
+		const request = convertRequestWithAdapter(r);
```
EagleoutIce: okay maybe i should have looked at the code first. This should *not* do the conversion here - we should make sure that every request that is created is created by the use of an adapter.

src/cli/slicer-app.ts
```diff
@@ -99,7 +100,11 @@ async function getSlice() {
 		console.log(JSON.stringify(output, jsonReplacer));
 	} else {
 		if(doSlicing && options.diff) {
-			const originalCode = options['input-is-text'] ? options.input : fs.readFileSync(options.input).toString();
+			let originalCode = options.input;
+			if(!options['input-is-text']) {
+				const request = convertRequestWithAdapter({ request: 'file', content: options.input });
```
EagleoutIce: ok so from a usage perspective, I would prefer a function named:

`requestFromFile`(s) - similar to `requestFromInput`. This request creator should then be used whenever we need a request (similar to `requestFromInput`)

src/cli/slicer-app.ts
```diff
@@ -99,7 +100,11 @@ async function getSlice() {
...
+				originalCode = request.request === 'text' ? request.content : fs.readFileSync(request.content).toString();
```
EagleoutIce, line 106: and just to clarify, the reason we do *this* and read it here because the diff presentation is a cheap hack, not meant to be used in any productive setup but just for visualization in the REPL

src/r-bridge/lang-4.x/tree-sitter/tree-sitter-executor.ts
```diff
@@ -59,6 +60,8 @@ export class TreeSitterExecutor implements SyncParser<Parser.Tree> {
 	}
 
 	public parse(request: RParseRequest): Parser.Tree {
+		request = convertRequestWithAdapter(request);
```
EagleoutIce: I repeat myself here but i think this way it is clearer -> once we have a request no adapter should trigger anymore, the adapters *should* be used when the request is created!

src/util/formats/adapter.ts
```diff
@@ -0,0 +1,32 @@
...
+	if(request.request === 'text') {
+		// For now we don't know what type the request is 
+		// and have to assume it is normal R Code
+		// In the future we could add a heuristic to guess the type
+		return 'R';
+	}
+
+	switch(path.extname(request.content).toLowerCase()) {
+		case '.r':   return 'R';
+		case '.rmd': return 'Rmd';
```
EagleoutIce: please use an object/map for this so that we can simply lookup the extname :) (and, e.g., really integrate system mimetypes etc. in the future)

src/util/formats/adapters/rmd-adapter.ts
```diff
@@ -0,0 +1,117 @@
...
+		goToLine(block.startpos.line);
+		output += block.code;
+		line += countNewlines(block.code);
+	}
+
+	return output;	
+}
+
+export function parseCodeBlockOptions(header: string, content: string): string {
+	let opts = header.length === 3
```
EagleoutIce: can you add documentation for this magic number?

test/testfiles/notebook/example.Rmd
```diff
@@ -0,0 +1,37 @@
...
+
+
+With Body Opts
+```{r, echo=FALSE}
+#| cache=FALSE
+cat(test)
+```
+
+
+Unrelated Block `right here with unrelated code`
```
EagleoutIce, line 33: Just for the test, could we have another *r* block after the unrelated code? :)

PR #1944, opened 2025-09-18
Integrate tree-sitter queries into the search

Ellpeck:


src/project/flowr-analyzer.ts
```diff
@@ -14,25 +18,21 @@ import type { GetSearchElements } from '../search/flowr-search-executor';
...
-	 */
-    parserInformation():   Promise<TreeSitterInformation | RShellInformation>;
-	/**
-	 * Sends a command to the underlying R engine and collects the output.
-	 * @param command     - The command to send to the R engine.
-	 * @param addonConfig - Additional configuration for the output collector.
-	 */
-	sendCommandWithOutput(command: string, addonConfig?: Partial<OutputCollectorConfiguration>): Promise<string[]>;
+    /**
+     * Returns a set of additional data and helper functions exposed by the underlying {@link KnownParser}.
```
EagleoutIce: I think, just for compatibility this should *also* hint at the `.metadata()` option

src/project/flowr-analyzer.ts
```diff
@@ -126,15 +127,23 @@ export class FlowrAnalyzer<Parser extends KnownParser = KnownParser> implements
...
-		guard(this.parser instanceof RShell, 'sendCommandWithOutput can only be used with RShell parsers!');
-		return this.parser.sendCommandWithOutput(command, addonConfig);
+	public parserInformation(): ParserInformation {
+		this.parserInfo ??= {
+			metadata: async() => {
+				return this.parser.name === 'r-shell' ?
+					{ name: 'r-shell', rVersion: await (this.parser as RShell).rVersion() }
+					: { name: 'tree-sitter' };
+			},
+			sendCommandWithOutput: (command: string, addonConfig?: Partial<OutputCollectorConfiguration>) => {
```
EagleoutIce: Why not instantiate this from the type? I.e. return a `ParserInformation<KnownParser>` or `KnownParser["ParseInfo"]` which allows us to delegate this things on a per-engine level. It feels weird to provide this for either engines.

src/project/flowr-analyzer.ts
```diff
@@ -126,15 +127,23 @@ export class FlowrAnalyzer<Parser extends KnownParser = KnownParser> implements
...
-
-	public async sendCommandWithOutput(command: string, addonConfig?: Partial<OutputCollectorConfiguration>): Promise<string[]> {
-		guard(this.parser instanceof RShell, 'sendCommandWithOutput can only be used with RShell parsers!');
-		return this.parser.sendCommandWithOutput(command, addonConfig);
+	public parserInformation(): ParserInformation {
+		this.parserInfo ??= {
+			metadata: async() => {
+				return this.parser.name === 'r-shell' ?
+					{ name: 'r-shell', rVersion: await (this.parser as RShell).rVersion() }
+					: { name: 'tree-sitter' };
```
EagleoutIce: i think this is a good point to also add the tree sitter grammar version? (for implicit documentation)

src/r-bridge/lang-4.x/tree-sitter/tree-sitter-executor.ts
```diff
@@ -68,6 +69,12 @@ export class TreeSitterExecutor implements SyncParser<Parser.Tree> {
 		return this.parser.parse(sourceCode);
 	}
 
+	public query(source: string, tree: Parser.Tree): QueryCapture[] {
+		const query = this.parser.getLanguage().query(source);
+		const matches = query.matches(tree.rootNode);
```
EagleoutIce, line 89: I would love for the top-level interface to also accept a `query` directly (i.e. provide a `makeQuery` and a `runQuery` interface. Because the first does (afaik) the optimization etc. and if we want to run the same syntax query against a 100 trees or so this should help

src/r-bridge/lang-4.x/tree-sitter/tree-sitter-normalize.ts
```diff
@@ -266,7 +282,8 @@ function convertTreeNode(node: SyntaxNode): RNode {
 				info:     {
 					fullRange:        range,
 					additionalTokens: [...variableComments, ...sequenceComments].map(c => c[1]),
```
EagleoutIce: uh, if we are already there, why not make this a concat? :D

src/search/flowr-search-builder.ts
```diff
@@ -41,6 +41,13 @@ export const FlowrSearchGenerator = {
 	fromQuery(...from: readonly Query[]): FlowrSearchBuilder<'from-query', [], ParentInformation, FlowrSearchElements<ParentInformation, FlowrSearchElement<ParentInformation>[]>> {
 		return new FlowrSearchBuilder({ type: 'generator', name: 'from-query', args: { from } });
 	},
+	/**
+	 * Initializes a new search query based on the results of the given tree-sitter query.
+	 * Please note that this search generator is incompatible with the {@link RShell} parser and only works when using flowR with the {@link TreeSitterExecutor}.
+	 */
+	fromTreeSitterQuery(source: string, ...captures: string[]): FlowrSearchBuilder<'from-query', [], ParentInformation, FlowrSearchElements<ParentInformation, FlowrSearchElement<ParentInformation>[]>> {
```
EagleoutIce: similarly, i want this to *also* accept a pre-build tree-sitter query 
additionally, `captures` should be readonly as an array, no? (I forgot, but maybe we also have a unique-array type to avoid duplications, maybe i build this for the labels)

src/search/search-executor/search-generators.ts
```diff
@@ -121,6 +130,46 @@ async function generateFromQuery(input: FlowrAnalysisProvider, args: {
...
+	if(!args.captures?.length) {
+		const defaultCaptureName = 'defaultCapture';
+		args.source += ` @${defaultCaptureName}`;
+		args.captures = [defaultCaptureName];
+	}
+	// allow specifying capture names with or without the @ in front :)
+	const captures = new Set<string>(args.captures.map(c => c.startsWith('@') ? c.substring(1) : c));
+
+	const result = await input.parserInformation().treeSitterQuery(args.source);
+	const relevant = result.filter(c => captures.has(c.name));
```
EagleoutIce, line 148: is there a way to retrieve the available captures in the query and warn if they do not match?

test/functionality/search/search-line.test.ts
```diff
@@ -91,6 +91,30 @@ describe('flowR search', withTreeSitter(parser => {
...
+			});
+		});
+
+		describe('custom capture', () => {
+			assertSearch('correct capture', parser, 'x <- "hello"', ['1@"hello"'], Q.fromTreeSitterQuery('(string) @s', 's'));
+			assertSearch('capture with @', parser, 'x <- "hello"', ['1@"hello"'], Q.fromTreeSitterQuery('(string) @s', '@s'));
+			assertSearch('incorrect capture', parser, 'x <- "hello"', [], Q.fromTreeSitterQuery('(string) @s', 'k'));
+
+			describe('multiple', () => {
+				assertSearch('binary op', parser, 'x <- 2\ny <- 17\ncat(y)', ['1@<-', '2@<-', '1@x', '2@y'], Q.fromTreeSitterQuery('(binary_operator lhs: (identifier) @id) @op', 'op', 'id'));
```
EagleoutIce: i love the tests, but could we have a couple tests that do something additionally on the query? just to show that you can also query dataflow information (e.g. only find `<-` which are function calls, or really built-in assignemnts etc.).

src/search/flowr-search-builder.ts
EagleoutIce: Besides: do you not like the `.query` proposal? or is this still wip? 

For the automatically generated documentation i think this would benefit from an `@example` block, what do you think?

src/cli/repl/commands/repl-execute.ts
```diff
@@ -16,7 +17,7 @@ If you want to do so, please restart flowR with the ${output.formatter.format('-
 
 async function executeRShellCommand(output: ReplOutput, analyzer: ReadonlyFlowrAnalysisProvider, statement: string) {
 	try {
-		const result = await analyzer.sendCommandWithOutput(statement, {
+		const result = await (analyzer.parserInformation() as RShellInformation).sendCommandWithOutput(statement, {
```
EagleoutIce, line 20: My original hope was that the type would carry the known parser in its generic and use this to provide adequate type information here. Did this not work?

PR #1948, opened 2025-09-24
Mass Eval Query Updates

EagleoutIce:


EagleoutIce:
// 1. find function definitions, where how; find nested helper function definitions too
// 2.  higher order this
// 3. for/repeat/while; maybe also parallel loops
// 4. library load , percentage in the code, whether nested or not
// 5. unique file loads and stores
// 6. use package names or first; maybe find where they are mixed
// 7. size of notebooks not really from flowr
// 8. vectorization ore like lapply etc. in better ways
// 9. do use query: function call query, maybe binary with left value right funktion
// 10. built-ins by function call name
// 11. how wide are dataframes => dataframe shape query


// partial matching on function names
// X. to people assert types
// Y. strictness of functions

PR #1950, opened 2025-09-26
perf(analyzer): re-use cfgQuick if available

### LINKED ISSUE #1933: [Projects] Optimize performance of the DFG extractor
If the dfg extractor requires a CFG (see [here](https://github.com/flowr-analysis/flowr/blob/baa235769e577bcac0b4735c06d36616fe56564a/src/dataflow/extractor.ts#L69)) make it so that it describes the initial CFG cache within the project (e.g., just pass it along or assign it in the object) - note, that the cfg is simple, hence, without any CFG processor active (currently for speed reasons but this shoud be configurable in the future).

MaxAtoms:
Closes #1933

### DIFF
--- a/src/project/cache/flowr-analyzer-cache.ts
+++ b/src/project/cache/flowr-analyzer-cache.ts
@@ -1,13 +1,11 @@
 import type { KnownParser } from '../../r-bridge/parser';
 import type { CacheInvalidationEvent } from './flowr-cache';
-import { FlowrCache , CacheInvalidationEventType } from './flowr-cache';
+import { CacheInvalidationEventType, FlowrCache } from './flowr-cache';
 import type {
 	DEFAULT_DATAFLOW_PIPELINE,
 	TREE_SITTER_DATAFLOW_PIPELINE
 } from '../../core/steps/pipeline/default-pipelines';
-import {
-	createDataflowPipeline
-} from '../../core/steps/pipeline/default-pipelines';
+import { createDataflowPipeline } from '../../core/steps/pipeline/default-pipelines';
 import type { PipelineExecutor } from '../../core/pipeline-executor';
 import type { FlowrConfigOptions } from '../../config';
 import { defaultConfigOptions } from '../../config';
@@ -182,10 +180,19 @@ export class FlowrAnalyzerCache<Parser extends KnownParser> extends FlowrCache<A
 		if(useDataflow) {
 			/* if force is active, it will have triggered with normalize */
 			dataflow = await this.dataflow();
+		} else {
+			dataflow = this.peekDataflow();
 		}
 
-		const result = simplifications.length === 0 && !useDataflow ? extractCfgQuick(normalized) :
-			extractCfg(normalized, this.args.config ?? defaultConfigOptions, dataflow?.graph, simplifications);
+		let result: ControlFlowInformation;
+		if(useDataflow || simplifications.length > 0) {
+			result = extractCfg(normalized, this.args.config ?? defaultConfigOptions, dataflow?.graph, simplifications);
+		} else if(dataflow !== undefined) {
+			result = dataflow.cfgQuick;
+		} else {
+			result = extractCfgQuick(normalized);
+		}
+
 		this.controlFlowCache.simplified.set([simplifications, useDataflow], result);
 		return result;
 	}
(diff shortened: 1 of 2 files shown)


src/project/cache/flowr-analyzer-cache.ts
```diff
@@ -185,10 +185,19 @@ export class FlowrAnalyzerCache<Parser extends KnownParser> extends FlowrCache<A
 		if(useDataflow) {
 			/* if force is active, it will have triggered with normalize */
 			dataflow = await this.dataflow();
+		} else {
+			dataflow = this.peekDataflow();
```
EagleoutIce: peeks should not be automatic but requested, we should update the `useDataflow` field to an enum in that case. Because in general fif i request the cfg without call information i may have certain expectations (e.g. on the existence of resolved call vertices) - so this should not happen just because we can!

src/project/cache/flowr-analyzer-cache.ts
```diff
@@ -185,10 +185,19 @@ export class FlowrAnalyzerCache<Parser extends KnownParser> extends FlowrCache<A
...
+		}
+
+		let result: ControlFlowInformation;
+		if(useDataflow || simplifications.length > 0) {
+			result = extractCfg(normalized, this.args.config, dataflow?.graph, simplifications);
+		} else if(dataflow && dataflow.cfgQuick) {
+			result = dataflow.cfgQuick;
```
EagleoutIce: can we mirror this fact into the cache for this? Otherwise it feels harder to write the bootstrapping for the cfgs with additional passing.

EagleoutIce: Extra PR, we do this once we start with the incremental cfg system

src/project/cache/flowr-analyzer-cache.ts
MaxAtoms: I hope the current state now reflects the outcome of our discussion last week

src/documentation/print-dataflow-graph-wiki.ts
```diff
@@ -1056,7 +1056,7 @@ Let's start by looking at the properties of the dataflow information object: ${O
 
 ${ (() => {
 			/* this includes the meta field for timing */
-			guard(Object.keys(result).length === 8, () => 'Update Dataflow Documentation!'); return ''; 
+			guard(Object.keys(result).length === 9, () => 'Update Dataflow Documentation!'); return '';
```
EagleoutIce, line 1059: this should indicate that we add at least a sentence explaining why the quickCfg may be there :D

src/project/cache/flowr-analyzer-cache.ts
```diff
@@ -168,39 +167,46 @@ export class FlowrAnalyzerCache<Parser extends KnownParser> extends FlowrCache<A
...
 
 		const normalized = await this.normalize(force);
-		let dataflow: NonNullable<AnalyzerCacheType<Parser>>['dataflow'] | undefined = undefined;
-		if(useDataflow) {
-			/* if force is active, it will have triggered with normalize */
-			dataflow = await this.dataflow();
+
+		let result: ControlFlowInformation;
+		switch(kind) {
+			case CfgKind.WithDataflow: {
```
EagleoutIce: do we need the braces here?

src/project/cfg-kind.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Denotes the kind of CFG.
+ */
+export enum CfgKind {
+	WithDataflow,
+	NoDataflow,
+	Quick
```
EagleoutIce, line 16: Maybe we should move the documentation of `quick` here :D

src/project/cache/flowr-analyzer-cache.ts
MaxAtoms: We now store the fact that this is a "quick CFG" in the cache.
The incremental CFG system is described in #1963.

src/documentation/print-dataflow-graph-wiki.ts
MaxAtoms: But it is sufficient within the comment above, right? Cf. https://github.com/flowr-analysis/flowr/pull/1950/files#diff-e40d4ac56ffdf2cb0a7bdfad4a080efd9c848a9f2a37c759c472298e78b0fd9bR1058

We also have the analyzer call (`FlowrAnalyzer.dataflow()`) on the Wiki page, but the quick CFG is not exposed in the type system there.

PR #1956, opened 2025-10-01
Re-use analyzer in the REPL API

MaxAtoms:


MaxAtoms:
The state is automatically reset between each REPL statement at the moment

src/project/flowr-analyzer.ts
```diff
@@ -15,16 +14,25 @@ import type { FlowrSearchLike, SearchOutput } from '../search/flowr-search-build
...
 /**
  * Exposes the central analyses and information provided by the {@link FlowrAnalyzer} to the linter, search, and query APIs.
  * This allows us to exchange the underlying implementation of the analyzer without affecting the APIs.
  */
 export interface FlowrAnalysisProvider {
 	/**
-	 * Get the name of the parser used by the analyzer.
+     * Get the name of the parser used by the analyzer.
+	 */
+    parserInformation():   Promise<{ name: KnownParserName, isRShell: false } | { name: KnownParserName, isRShell: true, rVersion: string }>;
```
EagleoutIce: Why do we need the additional `isRShell` Information? Is it just to separate the types in the union? In general i would prefer to "outsource" these types into a named interface + union

src/project/flowr-analyzer.ts
```diff
@@ -15,16 +14,25 @@ import type { FlowrSearchLike, SearchOutput } from '../search/flowr-search-build
...
+	/**
+	 * Sends a command to the underlying R engine and collects the output.
+	 * @param command - The command to send to the R engine.
```
EagleoutIce: I at least always try to align the `-` in the tsdoc - even though the linter has afaik no rule to enforce this

src/project/flowr-analyzer.ts
MaxAtoms: I seem to have forgotten that TypeScript can discriminate on literal strings 🙈 
I have now created new interfaces and moved them into `parser.ts`

PR #1966, opened 2025-10-20
Integrate the new structure of abstract domains

OliverGerstl:
- Use the new abstract domain structure introduced in #1752 for the data frame shape inference
- Adapt the abstract domains to be more flexible and better expandable


src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -34,41 +36,63 @@ extends Lattice<Abstract, Top, Bot, Lift> {
...
 	 * Maps a set of possible concrete values into an abstract value as abstraction function of the abstract domain (should additionally be provided as static function).
 	 */
-	abstract(concrete: ReadonlySet<Concrete> | typeof Top): AbstractDomain<Concrete, Abstract, Top, Bot>;
+	abstract(concrete: ReadonlySet<Concrete> | typeof Top): Domain;
 }
 
+/**
+ * An abstract domain with satisfiability checks for concrete values.
+ */
+export interface SatifiableDomain<T> {
```
EagleoutIce: Is the `Satifiable` typo intentional?

OliverGerstl: No :sweat_smile:

src/abstract-interpretation/data-frame/absint-visitor.ts
```diff
@@ -142,12 +142,12 @@ export class DataFrameShapeInferenceVisitor<
 		const value = resolveIdToDataFrameShape(node.info.dataFrame.expression, this.config.dfg, this.newDomain);
 
 		if(value !== undefined) {
-			this.newDomain.set(node.info.dataFrame.identifier, value);
+			this.newDomain.value.set(node.info.dataFrame.identifier, value);
```
EagleoutIce: why do we allow a direct access to the underlying value map? similar to `create`, `join` etc. i would prefer this to be wrapped in an `assign` or something similar

src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -58,18 +58,18 @@ export function resolveIdToDataFrameShape(
...
 	const call = vertex?.tag === VertexType.FunctionCall ? vertex : undefined;
 	const origins = Array.isArray(call?.origin) ? call.origin : [];
 
 	if(node.type === RType.Symbol) {
-		const values = getVariableOrigins(node.info.id, dfg).map(origin => domain.get(origin.info.id));
+		const values = getVariableOrigins(node.info.id, dfg).map(origin => domain.value.get(origin.info.id));
 
 		if(values.length > 0 && values.every(isNotUndefined)) {
-			return joinDataFrames(...values);
+			return values[0].join(...values.slice(1));
```
EagleoutIce: this feels... hella dirty :P Especially if we allow to mix various domains later (forcing them to be joined on the first value) - but fine so far

src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -58,18 +58,18 @@ export function resolveIdToDataFrameShape(
...
+	} else if(domain.value.has(node.info.id)) {
+		return domain.value.get(node.info.id);
 	}
 	const vertex = dfg.getVertex(node.info.id);
 	const call = vertex?.tag === VertexType.FunctionCall ? vertex : undefined;
 	const origins = Array.isArray(call?.origin) ? call.origin : [];
 
 	if(node.type === RType.Symbol) {
-		const values = getVariableOrigins(node.info.id, dfg).map(origin => domain.get(origin.info.id));
+		const values = getVariableOrigins(node.info.id, dfg).map(origin => domain.value.get(origin.info.id));
```
EagleoutIce: I would prefer having this as a delegate, again to avoid the value map being exposed

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -34,44 +37,97 @@ extends Lattice<Abstract, Top, Bot, Lift> {
...
+	 * Checks whether the current abstract value satisfies a concrete value (i.e. includes a concrete value).
+	 * @see {@link Ternary} for the returned satisfiability result
+	 */
+	satisfies(value: T): Ternary;
+
+	/**
+	 * Checks whether the current abstract value is an upper bound for a concrete value (i.e. the concrete value is less or equal than the upper bound of the abstract value).
+	 * @see {@link Ternary} for the returned satisfiability result
+	 */
+	satisfiesLeq(value: T): Ternary;
```
EagleoutIce: Having a separate satisfies leq feels.. weird? why not offer a version of satisfies that optionally allows for a comparator to be passed along? This would avoid the need for new functions of someone want to, say, check against a set of numbers to be bound etc.

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -34,44 +37,97 @@ extends Lattice<Abstract, Top, Bot, Lift> {
...
+	} else if(value === Bottom) {
+		return '⊥';
 	}
 	return JSON.stringify(value);
 }
+
+/**
+ * Converts an element of an abstract domain into a JSON-representable object or value.
+ */
+export function domainElementToJson(value: AnyAbstractDomain | unknown): unknown {
```
EagleoutIce: It feels kinda weird to move widen etc. into the domain but not the control to jsonify the value. Why isnt there a default `value.toJSON()?` (which would be called implicitly if found by stringify)

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -34,44 +37,97 @@ extends Lattice<Abstract, Top, Bot, Lift> {
...
+	if(isAbstractDomain(value)) {
+		return domainElementToJson(value.value);
+	} else if(value instanceof Map) {
+		return Object.fromEntries(value.entries().map(([key, value]) => [domainElementToJson(key), domainElementToJson(value)]));
+	} else if(Array.isArray(value) || value instanceof Set) {
+		return [...value].map(domainElementToJson);
+	} else if(typeof value === 'object' && value !== null) {
+		return Object.fromEntries(Object.entries(value).map(([key, value]) => [key, domainElementToJson(value)]));
+	} else if(value === Top) {
```
EagleoutIce: Given you check for an abstract domain in `112` it feels kinda weird to have this check so far down

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -34,44 +37,97 @@ extends Lattice<Abstract, Top, Bot, Lift> {
...
+		return 'bottom';
+	}
+	return value;
+}
+
+export function isAbstractDomain(value: unknown): value is AnyAbstractDomain {
+	if(typeof value !== 'object' || value === null) {
+		return false;
+	}
+	return ['value', 'top', 'bottom', 'leq', 'join', 'meet', 'widen', 'narrow', 'concretize', 'abstract'].every(property => property in value);
```
EagleoutIce, line 170: lovely contract check

src/abstract-interpretation/domains/bounded-set-domain.ts
```diff
@@ -121,6 +147,19 @@ implements AbstractDomain<T, BoundedSetValue<T>, BoundedSetTop, BoundedSetBottom
 		return BoundedSetDomain.abstract(concrete, this.limit);
 	}
 
+	public satisfies(value: T): Ternary {
+		if(this.isValue() && this.value.has(value)) {
+			return this.value.size === 1 ? Ternary.Always : Ternary.Maybe;
```
EagleoutIce, line 146: i mean given the overapproximation idea this should map to an `exact` and `subset` comparison, not directly to maybe?

src/abstract-interpretation/domains/bounded-set-domain.ts
```diff
@@ -121,6 +147,19 @@ implements AbstractDomain<T, BoundedSetValue<T>, BoundedSetTop, BoundedSetBottom
...
+		} else if(this.isTop()) {
+			return Ternary.Maybe;
```
EagleoutIce, line 148: Again this feels like true-ish maybe given the over-approximation

src/abstract-interpretation/domains/bounded-set-domain.ts
EagleoutIce: + this could also entail a missed `bottom` so `always` for a non-negated check (i.e., unsat) feels weird

src/abstract-interpretation/domains/interval-domain.ts
```diff
@@ -139,47 +158,69 @@ implements AbstractDomain<number, IntervalValue, IntervalTop, IntervalBottom, Va
...
+		}
+		return Ternary.Never;
+	}
+
 	/**
 	 * Adds another abstract value to the current abstract value by adding the two lower and upper bounds, respectively.
 	 */
-	public add(other: IntervalDomain): IntervalDomain {
-		if(this.value === Bottom || other.value === Bottom) {
+	public add(other: IntervalDomain | IntervalLift): IntervalDomain {
```
EagleoutIce: I accept that we follow a more lisa like implementation of language operators, _but_ what is your opinion on joining this with our function database as i tried to in 

https://github.com/flowr-analysis/flowr/blob/f4e3edcca6d1fe4dcc289d0bc0ba22f75c60bbc2/src/dataflow/eval/values/intervals/interval-binary.ts#L48-L74

src/abstract-interpretation/domains/lattice.ts
```diff
@@ -10,46 +10,52 @@ export const Bottom = Symbol('bottom');
 
 /**
  * A complete lattice with a partially ordered set, join operator (LUB), meet operator (GLB), top element, and bottom element (e.g. for abstract domains).
+ * @template Lat   - Type of the implemented lattice
```
EagleoutIce: i am kinda afraid that this recursion will break us for more complicated lattices, but lets see if TS can keep up

src/abstract-interpretation/domains/mapped-abstract-domain.ts
```diff
@@ -0,0 +1,215 @@
...
+			const otherValue = other.value.get(key);
+
+			if(otherValue === undefined || !value.leq(otherValue)) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	public join(...values: MapDomain[]): MapDomain {
```
EagleoutIce: why are these not marked as readonly? And, given the operations you perform on them: why varargs? is there a benefit? otherwise the spread feels like performance-hazard

src/abstract-interpretation/domains/mapped-abstract-domain.ts
```diff
@@ -0,0 +1,215 @@
...
+			const currValue = result.value.get(key);
+
+			if(currValue !== undefined) {
+				result.value.set(key, currValue.narrow(value) as Domain);
+			}
+		}
+		return result;
+	}
+
+	public concretize(limit: number = DEFAULT_INFERENCE_LIMIT): ReadonlySet<ConcreteMap<Key, Domain>> | typeof Top {
```
EagleoutIce: Why does the default not depend on a flowr configuration? It can rely on an attribute value

src/abstract-interpretation/domains/positive-interval-domain.ts
```diff
@@ -1,14 +1,17 @@
...
 import { Bottom, Top } from './lattice';
 
+/** The Top element of the positive interval domain as interval [0, +∞] */
+export const PosIntervalTop = [0, +Infinity] satisfies readonly [number, number];
+
 /** The type of the actual values of the positive interval domain as tuple of the lower and upper bound */
-export type PosIntervalValue = readonly [number, number];
-/** The type of the Top element of the positive interval domain as interval [0, +∞] from 0 to +∞ */
-export type PosIntervalTop = readonly [0, typeof Infinity];
+type PosIntervalValue = readonly [number, number];
```
EagleoutIce:
```suggestion
type PosIntervalValue = readonly [lower: number, upper: number];
```
Maybe use named tuples here for readability?

src/abstract-interpretation/domains/positive-interval-domain.ts
```diff
@@ -49,13 +56,51 @@ export class PosIntervalDomain<Value extends PosIntervalLift = PosIntervalLift>
 		return PosIntervalDomain.bottom();
 	}
 
+	public join(...values: PosIntervalDomain[]): PosIntervalDomain;
+	public join(...values: PosIntervalLift[]): PosIntervalDomain;
+	public join(...values: PosIntervalDomain[] | PosIntervalLift[]): PosIntervalDomain {
```
EagleoutIce: same argument as before with `readonly` and spread

src/abstract-interpretation/domains/positive-interval-domain.ts
```diff
@@ -49,13 +56,51 @@ export class PosIntervalDomain<Value extends PosIntervalLift = PosIntervalLift>
 		return PosIntervalDomain.bottom();
 	}
 
+	public join(...values: PosIntervalDomain[]): PosIntervalDomain;
+	public join(...values: PosIntervalLift[]): PosIntervalDomain;
```
EagleoutIce: are these... necessary? You do not define a dependent type relation wrt. the return type so this is just the signature below

src/abstract-interpretation/domains/positive-interval-domain.ts
```diff
@@ -49,13 +56,51 @@ export class PosIntervalDomain<Value extends PosIntervalLift = PosIntervalLift>
...
+	public join(...values: PosIntervalDomain[] | PosIntervalLift[]): PosIntervalDomain {
+		let result: PosIntervalLift = this.value;
+
+		for(const other of values) {
+			const otherValue = other instanceof PosIntervalDomain ? other.value : other;
+
+			if(result === Bottom) {
+				result = otherValue;
+			} else if(otherValue === Bottom) {
+				continue;
```
EagleoutIce: Am I biased? yes, _but_
I prefer to:
1. let client domains implement the binary case to handle the reduction in a default implementation
2. outsource the bottom top guards as in https://github.com/flowr-analysis/flowr/blob/3e31fff7d256f26bdae72283ff54c07be9573ce4/src/dataflow/eval/values/general.ts#L12-L18 (even though this was a sloppy setup)

src/abstract-interpretation/domains/set-upper-bound-domain.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+export class SetUpperBoundDomain<T, Value extends SetUpperBoundLift<T> = SetUpperBoundLift<T>>
+implements AbstractDomain<SetUpperBoundDomain<T>, ReadonlySet<T>, SetUpperBoundValue<T>, SetUpperBoundTop, SetUpperBoundBottom, Value>, SatisfiableDomain<ReadonlySet<T>> {
+	public readonly limit:   number;
+	private readonly _value: Value;
+
+	constructor(value: Value | T[], limit: number = DEFAULT_INFERENCE_LIMIT) {
+		if(value !== Top && value !== Bottom) {
+			if(Array.isArray(value)) {
+				this._value = (value.length > limit ? Top : new Set(value)) as Value;
```
EagleoutIce: Given that we do not know the tpe of `T` just transforming it into a set becomes highly problematic if the type is not hashable by JavaScripts hash set so maybe we have to offer the ability to provide a `ReadonlySet` constructor (that can then be filled by, e.g. an object set)

EagleoutIce: Or, of course restrict the values

src/abstract-interpretation/data-frame/shape-inference.ts
OliverGerstl: The idea here is to use the config options of the first domain (e.g. the limit for the number of inferred values of a bounded set domain) when joining the domains

EagleoutIce: I see :D - I am still not sure why this uglyness cannot be hidden within a `Domain.joinall` or something but fine :D

src/abstract-interpretation/domains/lattice.ts
OliverGerstl: Yeah, that was to simplify inheritance for join, meet, widen etc.

src/abstract-interpretation/domains/mapped-abstract-domain.ts
OliverGerstl: I use the varargs here to allow for single arguments (e.g. as in the abstract semantics) and list arguments (e.g. as in the visitor) to have a uniform interface (so that I can call `join(a, b)`, `meet(a, b)`, `widen(a, b)`, `narrow(a, b)` etc. but join and meet can also be called with lists as arguments)

src/abstract-interpretation/domains/bounded-set-domain.ts
OliverGerstl: I think `Ternary` actually fits quite well here to answer the question "Is this concrete value satisfied by the abstract value?":
- Always => The value is certainly satisfied (~`true`)
- Maybe => The value could be satisfied, but we don't know (~`unknown`)
- Never => The value cannot be satisfied (~`false`)

src/abstract-interpretation/domains/interval-domain.ts
OliverGerstl: Yes, that would definitely make sense, but we would have to restructure this to differentiate between operators/operations and (satisfiability) checks.
I created a separate issue for that #1992

src/abstract-interpretation/domains/mapped-abstract-domain.ts
OliverGerstl: This is just a static default value to limit the number of concretized values, as this can be used for many different domains. I can also make this non-optional to always enforce a value, e.g. resulting from the config

src/abstract-interpretation/domains/positive-interval-domain.ts
OliverGerstl: Oh lol, I didn't know that TypeScript supports named tubles, nice :)

src/abstract-interpretation/domains/positive-interval-domain.ts
OliverGerstl: The problem is that TypeScript somehow needs the redundant method prototypes for `join` and `meet` here, as the arguments are arrays. This is however not needed for the `add` and `subtract` method of the interval domain, for example

src/abstract-interpretation/domains/positive-interval-domain.ts
OliverGerstl: 1. The problem is that TypeScript does not allow default implementations for interface methods nor static methods with implementation in interfaces
2. Most of the time I don't need a bottom top guard but rather just a bottom guard or top guard, and I think it is better readable when having this explicit checks

src/abstract-interpretation/domains/positive-interval-domain.ts
EagleoutIce: crazy restricted, i thought they solved this :D - but fine!

src/abstract-interpretation/domains/positive-interval-domain.ts
EagleoutIce: allright then maybe i was biased (even though i would have moved it to a default prototype for the interface as they did to bridge javascript news)

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -15,17 +15,17 @@ export const DEFAULT_INFERENCE_LIMIT = 50;
...
  */
-export interface AbstractDomain<Domain extends AbstractDomain<Domain, Concrete, Abstract, Top, Bot>, Concrete, Abstract, Top = typeof Top, Bot = typeof Bottom, Value extends Abstract | Top | Bot = Abstract | Top | Bot>
-extends Lattice<Domain, Abstract, Top, Bot, Value> {
+export interface AbstractDomain<Concrete, Abstract, Top, Bot, Value extends Abstract | Top | Bot = Abstract | Top | Bot>
+extends Lattice<Abstract, Top, Bot, Value> {
 	/**
 	 * Widens the current abstract value with another abstract value as a sound over-approximation of the join (least upper bound) for fixpoint iteration acceleration.
 	 */
-	widen(other: Domain): Domain;
+	widen(other: this): this;
```
EagleoutIce: if you change the paramter to this as well I think this is too restrictive wrt. types that this one can be matched with (if you want exact class equivalence this is fine, but if you want to, later, for example allow to widen a interval domain by a scalar, this could be genralized by an `T extends this` clause.

OliverGerstl: I am not sure what you mean by that. Would a `T extends this` not have the same meaning as just `this`, as `this` also allows subclasses of it? And, on the other hand, a scalar domain is not a subclass of a interval domain, so I would rather need something like `AbstractDomain<number>`, but than I would have to include extra logic to widen between arbitrary numeric domains

EagleoutIce: for now this is fine but we can discuss the future implications regarding locking-in to class hierarchies :)

src/abstract-interpretation/domains/bounded-set-domain.ts
EagleoutIce: Ok if you do not plan on subset satisfies this is fine :D

src/abstract-interpretation/domains/mapped-abstract-domain.ts
EagleoutIce: i would prefer this non optional then

PR #1985, opened 2025-10-23
[resolve]: Return Top in situations that we cannot resolve yet (when using Graph only)

gigalasr:


EagleoutIce:
_This pull request is included in v2.6.2 (see [Release v2.6.2 (TS-Queries, new Domain Structure, more Reserved Words, removed Lineage Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.6.2))._

PR #1989, opened 2025-10-24
[REPL] Add line parser for the linter query

### LINKED ISSUE #1943: [Bug]: Config Updates via Query Are Evaluated As Code
### What happened?

When updating the flowR configuration via the config query, the "update string" (part of the string that describes the update) is being evaluated internally as R code. This has no negative effect on the output, but it is something we would like to avoid.

E.g. `R> :query @config +ignoreSourceCalls=true`

See: https://github.com/flowr-analysis/flowr/blob/293fe1972e518eecfc938a4909bcefa3dde5decd/src/cli/repl/commands/repl-query.ts#L76

Internal state:

Image

4.5.1

v2.4.8 (Sep 3, 2025)

### Relevant log output

```Shell

```

### Meta Information

- [ ] The issue is urgent

MaxAtoms:
Allows queries with and without naming individual rules and with and without an explicit input code request.

E.g.

```
R> :query @linter file://somefile.R
R> :query @linter rules:dead-code,file-path-validity
R> :query @linter
```

Closes #1943 as side-effect

### DIFF
--- a/src/cli/repl/commands/repl-lineage.ts
+++ b/src/cli/repl/commands/repl-lineage.ts
@@ -7,6 +7,7 @@ import type { DataflowGraphEdge } from '../../../dataflow/graph/edge';
 import { edgeIncludesType, EdgeType } from '../../../dataflow/graph/edge';
 import type { AstIdMap } from '../../../r-bridge/lang-4.x/ast/model/processing/decorate';
 import { guard } from '../../../util/assert';
+import { startAndEndsWith } from '../../../util/text/strings';
 
 function splitAt(str: string, idx: number) {
 	return [str.slice(0, idx), str.slice(idx)];
@@ -65,7 +66,7 @@ export const lineageCommand: ReplCodeCommand = {
 		const [criterion, rest] = splitAt(args, args.indexOf(' '));
 		const code = rest.trim();
 		return {
-			rCode:     code.startsWith('"') ? JSON.parse(code) as string : code,
+			rCode:     startAndEndsWith(code, '"') ? JSON.parse(code) as string : code,
 			remaining: [criterion]
 		};
 	},
--- a/src/queries/catalog/linter-query/linter-query-format.ts
+++ b/src/queries/catalog/linter-query/linter-query-format.ts
@@ -36,8 +36,8 @@ export interface LinterQueryResult extends BaseQueryResult {
 
 function rulesFromInput(output: ReplOutput, rulesPart: readonly string[]): {valid: (LintingRuleNames | ConfiguredLintingRule)[], invalid: string[]} {
 	return rulesPart
-		.map(r => r.trim())
 		.reduce((acc, ruleName) => {
+			ruleName = ruleName.trim();
 			if(ruleName in LintingRules) {
 				acc.valid.push(ruleName as LintingRuleNames);
 			} else {
--- a/src/cli/repl/core.ts
+++ b/src/cli/repl/core.ts
@@ -22,6 +22,7 @@ import type { FlowrConfigOptions } from '../../config';
 import type { SupportedQuery } from '../../queries/query';
 import { SupportedQueries } from '../../queries/query';
 import type { FlowrAnalyzer } from '../../project/flowr-analyzer';
+import { startAndEndsWith } from '../../util/text/strings';
 
 let _replCompleterKeywords: string[] | undefined = undefined;
 function replCompleterKeywords() {
@@ -99,7 +100,7 @@ export function makeDefaultReplReadline(config: FlowrConfigOptions): readline.Re
 
 export function handleString(code: string) {
 	return {
-		rCode:     code.length == 0 ? undefined : code.startsWith('"') ? JSON.parse(code) as string : code,
+		rCode:     code.length == 0 ? undefined : startAndEndsWith(code, '"') ? JSON.parse(code) as string : code,
 		remaining: []
 	};
 }


src/cli/repl/commands/repl-main.ts
```diff
@@ -86,5 +86,5 @@ export interface ReplCodeCommand extends ReplBaseCommand {
 	/**
 	 * Argument parser function which handles the input given after the repl command
 	 */
-	argsParser:    (remainingLine: string) => { input: string | undefined, remaining: string[]}
+	argsParser:    (remainingLine: string) => { input?: string | undefined, remaining: string[]}
```
EagleoutIce: Please document the semantics of not returning an input (i.e. that this reuses the last state)

src/cli/repl/commands/repl-query.ts
```diff
@@ -57,10 +61,16 @@ async function processQueryArgs(output: ReplOutput, analyzer: FlowrAnalysisProvi
...
 			return;
 		}
+		input = remainingArgs.join(' ').trim();
 	} else {
 		parsedQuery = [{ type: 'call-context', callName: query }];
 	}
 
+	if(input) {
+		analyzer.reset();
+		analyzer.context().addRequest(requestFromInput(input));
```
EagleoutIce: I dont think adding should happen only on the context layer as this *has* to retrigger loading order passes etc. (and maybe even lead to the identification of a new root description file) could you please delegate these to the `analyzer`? (in this case we could also just make the builder functions delegate to the analyzer). 
+ this would encourage to not just offer an inspectable (readonly) context but also a readonly analyzer which no longer ships with these functions - do you follow me there?

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -27,6 +33,29 @@ export interface LinterQueryResult extends BaseQueryResult {
...
 
+function rulesFromInput(rulesPart: string[]): (LintingRuleNames | ConfiguredLintingRule)[] {
+	return rulesPart.map(rule => {
+		const ruleName = rule.trim();
+		if(!(ruleName in LintingRules)) {
+			console.error(`Unknown linting rule '${ruleName}'`);
+			return;
+		}
+		return ruleName as LintingRuleNames;
+	}).filter(r => !(r === undefined));
```
EagleoutIce: we do have a `isNotUndefined` for this

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -27,6 +33,29 @@ export interface LinterQueryResult extends BaseQueryResult {
 	readonly results: { [L in LintingRuleNames]?: LintingResults<L>}
 }
 
+function rulesFromInput(rulesPart: string[]): (LintingRuleNames | ConfiguredLintingRule)[] {
```
EagleoutIce:
```suggestion
function rulesFromInput(rulesPart: readonly string[]): (LintingRuleNames | ConfiguredLintingRule)[] {
```

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -27,6 +33,29 @@ export interface LinterQueryResult extends BaseQueryResult {
...
+	return rulesPart.map(rule => {
+		const ruleName = rule.trim();
+		if(!(ruleName in LintingRules)) {
+			console.error(`Unknown linting rule '${ruleName}'`);
```
EagleoutIce: could you please use the `output` passed by the repl and not just dump to the console? this output should also provide formatting options

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -27,6 +33,29 @@ export interface LinterQueryResult extends BaseQueryResult {
...
+		}
+		return ruleName as LintingRuleNames;
+	}).filter(r => !(r === undefined));
+}
+
+function linterQueryLineParser(line: readonly string[], _config: FlowrConfigOptions): ParsedQueryLine {
+	let rules: (LintingRuleNames | ConfiguredLintingRule)[] | undefined = undefined;
+	let input: string | undefined = undefined;
+	if(line.length > 0 && line[0].startsWith('rules:')) {
+		const rulesPart = line[0].slice('rules:'.length).split(',');
```
EagleoutIce: Please outsource the `rules:` constant string

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -38,7 +67,8 @@ export const LinterQueryDefinition = {
 		}
 		return true;
 	},
-	schema: Joi.object({
+	fromLine: linterQueryLineParser,
```
EagleoutIce, line 79: so this does not allow for autocompleetion currently right? it is fine if we do this separately, just to clarify the goal with this

src/queries/query.ts
```diff
@@ -89,12 +88,17 @@ type SupportedQueriesType = {
 	[QueryType in Query['type']]: SupportedQuery<QueryType>
 }
 
+export interface ParsedQueryLine {
```
EagleoutIce, line 95: Maybe we should at least add a ruff documentation for this

src/queries/query.ts
```diff
@@ -89,12 +88,17 @@ type SupportedQueriesType = {
...
+	query:  Query | Query[] | undefined;
+	input?: string;
```
EagleoutIce: likewise maybe not call it `input` but something like `rCode`

src/queries/catalog/linter-query/linter-query-format.ts
MaxAtoms: I should've asked why this was not used for the config query parser.

It looks like this now:

image

And for the linter query:

image

src/queries/catalog/linter-query/linter-query-format.ts
MaxAtoms: Yes, completions for linter rules and slicing criteria are next. I would open a separate PR for this :)

src/cli/repl/commands/repl-query.ts
MaxAtoms: I think I understand what you mean. I introduced a `ModifiableFlowrAnalysisProvider` interface now, but we can also turn it around and rename the existing interface to `Readonly`.
Related commits: 775da56a6099976dfd3a453acc7c237e522defc5, df03321ddabaa8d592ceec9138199741c86531d3

src/queries/catalog/linter-query/linter-query-format.ts
EagleoutIce: Perfectly fine!

src/cli/repl/commands/repl-query.ts
EagleoutIce: I would prefer having it the other way around with a `Readonly` version to stay in sync with the context and other layers in flowR

src/cli/repl/commands/repl-lineage.ts
```diff
@@ -65,7 +65,7 @@ export const lineageCommand: ReplCodeCommand = {
 		const [criterion, rest] = splitAt(args, args.indexOf(' '));
 		const code = rest.trim();
 		return {
-			input:     code.startsWith('"') ? JSON.parse(code) as string : code,
+			rCode:     code.startsWith('"') ? JSON.parse(code) as string : code,
```
EagleoutIce: we technically also have a startsEndsWith (or something like this) to check whether it also *ends* with a quote and isnt just `"a" <- 3`

src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -27,6 +34,37 @@ export interface LinterQueryResult extends BaseQueryResult {
 	readonly results: { [L in LintingRuleNames]?: LintingResults<L>}
 }
 
+function rulesFromInput(output: ReplOutput, rulesPart: readonly string[]): {valid: (LintingRuleNames | ConfiguredLintingRule)[], invalid: string[]} {
+	return rulesPart
+		.map(r => r.trim())
```
EagleoutIce: i think moving this into the reduce would be more performant :D

EagleoutIce:
_This pull request is included in v2.6.2 (see [Release v2.6.2 (TS-Queries, new Domain Structure, more Reserved Words, removed Lineage Query)](https://github.com/flowr-analysis/flowr/releases/tag/v2.6.2))._

PR #2016, opened 2025-11-10
Replace deprecated Vitest workspace with config

MaxAtoms:
Cf. https://vitest.dev/blog/vitest-3-2.html#workspace-is-deprecated

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -1,0 +1,10 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	test: {
+		projects: [
+			'./test/vitest.config.mts',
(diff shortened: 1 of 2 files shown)


EagleoutIce:
And even though we use npm runs, people still want to use the IDE runners, but let's have the people with problems (maybe @Ellpeck was one of them) have a look at this and see whether this also resolves the problem.

But given that it apprently works for you @MaxAtoms maybe it works for the others too :D

EagleoutIce:
_This pull request is included in v2.6.3 (see [Release v2.6.3 (Doc, Refined Analyzer, Caching, and Perf)](https://github.com/flowr-analysis/flowr/releases/tag/v2.6.3))._

PR #2017, opened 2025-11-11
Add `Add` method to analyzer

### LINKED ISSUE #2008: [Project] Analyzer should provide same `add` methods
With the builder you can do `add`, but with the analyzer you have to do `analyzer.context().addRequest(...)` with an optional `requestFromInput` wrapper and this should be made simpler!

MaxAtoms:
Closes #2008

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -1,0 +1,10 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	test: {
+		projects: [
+			'./test/vitest.config.mts',
+			'./test/system-tests/vitest.config.mts'
+		]
+	}
+});
--- a/vitest.workspace.js
+++ b/vitest.workspace.js
@@ -1,6 +1,0 @@
-import { defineWorkspace } from 'vitest/config';
-
-export default defineWorkspace([
-	'./test/vitest.config.mts',
-	'./test/system-tests/vitest.config.mts'
-]);


src/project/flowr-analyzer.ts
```diff
@@ -27,13 +27,10 @@ export interface FlowrAnalysisProvider extends ReadonlyFlowrAnalysisProvider {
...
-	 * Add multiple analysis requests to the analyzer instance
+	 * Add one or multiple requests to analyze.
+	 * @param request - One or multiple requests or a file path (with the `file://` protocol). If you just enter a string, it will be interpreted as R code.
 	 */
-	addRequests(requests: readonly RAnalysisRequest[]): void
-	/**
-	 * Add a single analysis request to the analyzer instance
-	 */
-	addRequest(request: RAnalysisRequest): void
+	add(request: RAnalysisRequest | readonly RAnalysisRequest[] | `${typeof fileProtocol}${string}` | string): void
```
MaxAtoms: I am conflicted whether `addRequest` would be at least a bit more descriptive (although the parameter list clearly shows that this is for adding an `AnalysisRequest`

src/project/flowr-analyzer.ts
```diff
@@ -147,12 +144,38 @@ export class FlowrAnalyzer<Parser extends KnownParser = KnownParser> implements
...
+			this.addRequest(request);
+		}
+		return this;
+	}
+
+	/**
+	 * Add a request created from the given input.
+	 * This is a convenience method that uses {@link requestFromInput} internally.
+	 */
+	private addRequestFromInput(input: Parameters<typeof requestFromInput>[0]): this {
```
MaxAtoms, line 167: I did not expose these two methods because I did not see a need for it.
Should they be `public`? I guess the builder and the analyzer should be consistent with regard to this

src/project/flowr-analyzer.ts
```diff
@@ -147,12 +144,38 @@ export class FlowrAnalyzer<Parser extends KnownParser = KnownParser> implements
 		this.cache.reset();
 	}
 
-	public addRequests(requests: readonly RAnalysisRequest[]): void {
-		this.ctx.addRequests(requests);
+	public add(request: RAnalysisRequest | readonly RAnalysisRequest[] | `${typeof fileProtocol}${string}` | string): this {
```
MaxAtoms: FYI: WebStorm now shows this a as duplicate code, but I did not see a benefit in sharing code between the analyzer and the builder

src/project/flowr-analyzer.ts
EagleoutIce: I would want to have both `add` is intended to be magic!

src/project/flowr-analyzer.ts
EagleoutIce: I am torn but for now this is fine

src/project/flowr-analyzer.ts
```diff
@@ -147,12 +144,38 @@ export class FlowrAnalyzer<Parser extends KnownParser = KnownParser> implements
...
+		if(Array.isArray(request) || isParseRequest(request)) {
```
EagleoutIce, line 148: I would at least outsource the magic? of testing what kind of rquest we have at hand,
likewsie we could wrap the request list into a class/object that exposes these things (on second thought this feels better?)

PR #2023, opened 2025-11-13
[Analyzer, CFG] Incremental caching

### LINKED ISSUE #1963: [Projects, CFG] Optimized, incremental CFG Caching
We want to optimize the CFG caching for additional performance gains:

- If we have an existing CFG we want to re-use it and apply simplifications on top of it
- Some simplifications are confluent. If that is the case, the order of their application is irrelevant

MaxAtoms:
Closes #1963


src/project/cache/flowr-analyzer-controlflow-cache.ts
```diff
@@ -0,0 +1,103 @@
+import { ObjectMap } from '../../util/collections/objectmap';
```
MaxAtoms, line 1: This class is actually somewhat independent of the analyzer.
We could put it into `src/control-flow`. But I do not expect that it will be re-used outside the context of the `FlowrAnalyzer`, so I wouldn't move it.

src/project/cache/flowr-analyzer-controlflow-cache.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { ControlFlowInformation } from '../../control-flow/control-flow-graph';
+import { guard } from '../../util/assert';
+import { extractCfg, extractCfgQuick } from '../../control-flow/extract-cfg';
+import type { NormalizedAst } from '../../r-bridge/lang-4.x/ast/model/processing/decorate';
+import type { FlowrConfigOptions } from '../../config';
+import type { DataflowInformation } from '../../dataflow/info';
+
+type ControlFlowCache = ObjectMap<[passes: readonly CfgSimplificationPassName[], kind: CfgKind], ControlFlowInformation>;
+
+interface CfgInfo {
```
MaxAtoms, line 14: The only awkward part of having this separate class from the `FlowrAnalyzerCache` is the AST and dataflow information passing

src/project/cache/flowr-analyzer-controlflow-cache.ts
```diff
@@ -0,0 +1,103 @@
...
+		}
+		return { cfg: undefined, missingSimplifications: simplifications };
+	}
+
+	/**
+	 * Normalize the order of simplification passes to represent their order of dependence.
+	 * @param simplifications - the requested simplification passes.
+	 */
+	private normalizeSimplificationOrder(simplifications: readonly CfgSimplificationPassName[]): readonly CfgSimplificationPassName[] {
+		return this.canonicalOrder.filter(p => simplifications.includes(p));
```
MaxAtoms: Is this "canonical order" fine for now or would you rather have an identity function as discussed? The canonical order prevents the user from shooting themselves in the foot by having the simplifications incorrectly ordered

test/functionality/project/cache/flowr-analyzer-cache.test.ts
```diff
@@ -0,0 +1,91 @@
...
+describe('Analyzer Cache', withTreeSitter( (shell) => {
+	const data = {
+		parser:  shell,
+		config:  defaultConfigOptions,
+		request: [requestFromInput('x <- 1')]
+	};
+
+	describe('Control Flow', () => {
+		test('CFG without dataflow', async() => {
+			const cache = FlowrAnalyzerCache.create({ parser: shell, config: defaultConfigOptions, request: requestFromInput('f <- function(x) x\nf()') });
```
MaxAtoms: Maybe we should rather test the difference between a CFG with dataflow and one without in `control-flow-graph.test` (cf. `assert-control-flow-graph#assertCfg`)? However, I did only find tests there asserting the CFG with dataflow and the presence of `callTargets` does not seem to be checked

test/functionality/project/cache/flowr-analyzer-cache.test.ts
```diff
@@ -0,0 +1,91 @@
...
+			const actual = await cache.controlflow(false, CfgKind.WithDataflow, []);
+			const expected = extractCfg(await cache.normalize(), data.config, (await cache.dataflow()).graph);
+			expect(expected).toEqual(actual);
+		});
+
+		test('CFG Quick', async() => {
+			const cache = FlowrAnalyzerCache.create({ parser: shell, config: defaultConfigOptions, request: requestFromInput('f <- function(x) x\nf()') });
+			const quick = await cache.controlflow(false, CfgKind.Quick);
+			const regular = extractCfg(await cache.normalize(), data.config);
+			expect(regular).not.toEqual(quick);
```
MaxAtoms: How do I actually create a CFG without dataflow and simplifications that is different to the CFG Quick, @EagleoutIce? Is there actually a difference?

EagleoutIce: Right now, no, in the future potentially yes :D - or how to understand this question

src/project/cache/flowr-analyzer-controlflow-cache.ts
EagleoutIce: oh this is fine it is part of the analyzer layer

src/project/cache/flowr-analyzer-controlflow-cache.ts
EagleoutIce: fair, but i think this is fine

src/project/cache/flowr-analyzer-controlflow-cache.ts
EagleoutIce: Sadly no because if the user wants `remove-dead-code, analyze-dead-code` in that order this could break him

test/functionality/project/cache/flowr-analyzer-cache.test.ts
EagleoutIce: unsure what you mean with that but the difference is currently not stable

src/dataflow/graph/graph.ts
```diff
@@ -109,11 +114,11 @@ export interface DataflowGraphJson {
 
 /**
  * An unknown side effect describes something that we cannot handle correctly (in all cases).
- * For example, `eval` will be marked as an unknown side effect as we have no idea of how it will affect the program.
+ * For example, `cat` will be marked as an unknown side effect as we have no idea of how it will affect the program.
```
EagleoutIce: why cat?

MaxAtoms: I was looking for a replacement for `eval` that is commonly known by R users and still has unknown side effects

EagleoutIce: But `cat` will not be marked as an unknown side effect in general, it will be hooked to a sink etc. and only be treated as unknown if we do not know what it affects

MaxAtoms: Yeah, as I looked at it again I realized that it is not the best example here because of the `link-to-last-call`. 
What function should we mention here in your opinion? `load` or `setwd` perhaps?

EagleoutIce: i think if we have to change then `load`

PR #2031, opened 2025-11-18
[DF-Shape] Fix issues identified during evaluation of data frame access validation

OliverGerstl:
This fixes issues identified during the evaluation of the data frame access linter rule:
- Only map column access for columns in `summarize` that are not created by `summarize`
- Support the addition of columns to empty data frames via `rbind`

### DIFF
--- a/test/functionality/abstract-interpretation/data-frame/inference.test.ts
+++ b/test/functionality/abstract-interpretation/data-frame/inference.test.ts
@@ -2657,6 +2657,33 @@ df <- subset(df, select = c(id, name), drop = TRUE)
 			]
 		);
 
+		assertDataFrameOperation(
+			shell,
+			`
+df <- data.frame(id = 1:3, name = 4:6, label = "A")
+result <- subset(df, id > 1, c("name", "label"))
+			`.trim(),
+			[['2@subset', [{ operation: 'accessCols', columns: ['id', 'name', 'label'] }]]]
+		);
+
+		assertDataFrameOperation(
+			shell,
+			`
+df <- data.frame(id = 1:3, name = 4:6, label = "A")
+result <- subset(df, label != "A", -id)
+			`.trim(),
+			[['2@subset', [{ operation: 'accessCols', columns: ['id', 'label'] }]]]
+		);
+
+		assertDataFrameOperation(
+			shell,
+			`
+df <- data.frame(id = 1:3, name = 4:6, label = "A")
+result <- subset(df, select = 1:2)
+			`.trim(),
+			[['2@subset', [{ operation: 'accessCols', columns: [1, 2] }]]]
+		);
+
 		describe('Unsupported', { fails: true }, () => {
 			testDataFrameDomainAgainstReal(
 				shell,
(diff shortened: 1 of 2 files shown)


src/queries/catalog/df-shape-query/df-shape-query-format.ts
```diff
@@ -47,14 +33,14 @@ export const DfShapeQueryDefinition = {
 		return true;
 	},
 	jsonFormatter: (queryResults: BaseQueryResult) => {
-		const out = queryResults as QueryResults<'df-shape'>['df-shape'];
-		const domains = out.domains instanceof DataFrameStateDomain ? out.domains.value : out.domains;
-		const json = Object.fromEntries(domains.entries().map(([key, domain]) => [key, domain?.toJson()])) as object;
+		const { domains, ...out } = queryResults as QueryResults<'df-shape'>['df-shape'];
+		const state = domains instanceof DataFrameStateDomain ? domains.value : domains;
+		const json = Object.fromEntries(state.entries().map(([key, domain]) => [key, domain?.toJson() ?? null]));
```
EagleoutIce, line 38: how about adding a `compactObject` to prune missing jsons? or do you *want* to have them explicitly null here?

test/functionality/abstract-interpretation/data-frame/inference.test.ts
```diff
@@ -2207,6 +2076,18 @@ df <- rbind(df, data.frame(id = 1:5, name = "A"))
 			]
 		);
 
+		testDataFrameDomain(
+			shell,
+			`
+df <- data.frame()
+df <- rbind(df, 12)
```
EagleoutIce, line 2083: any reason there is no new test case for the `summarize` patch?

src/queries/catalog/df-shape-query/df-shape-query-format.ts
OliverGerstl: They can only be `undefined` if the data frame shape is requested for a slicing criterion for which we did not infer a data frame shape. Therefore, I want to have `null` explicit here for requested slicing criteria without data frame shapes

EagleoutIce: Allright, fine by me - although we could use a more readable `none` or something.

test/functionality/abstract-interpretation/data-frame/inference.test.ts
OliverGerstl: The problem is that the summarize patch only removes some mappings to `accessCols` and does not change the inferred data frame shapes

src/queries/catalog/df-shape-query/df-shape-query-format.ts
OliverGerstl: Yeah, the `null` is only for JSON serialization

test/functionality/abstract-interpretation/data-frame/inference.test.ts
EagleoutIce: Fine but shouldn't there still be a test checking these mappings? :sweat_smile:

Issue #1791, opened 2025-07-02
Use value lifts with `Top` for arguments of abstract operations


OliverGerstl:
Use value lifts with `Top` for the values of the arguments of the data frame operations to explicitly mark arguments as `Top` instead of using `undefined` (maybe even as `RValue`s)

PR #2032, opened 2025-11-18
[DF-Shape] Use a set range domain for the column names of data frames

OliverGerstl:
- Add a set range domain to track a minimum set and maximum set of values of a specific type (e.g. strings)
- Integrate the set range domain as column names domain for the data frame shape domain
- Adapt the tests to the new column names domain

### DIFF
--- a/src/abstract-interpretation/data-frame/semantics.ts
+++ b/src/abstract-interpretation/data-frame/semantics.ts
@@ -130,7 +130,7 @@ function applyAccessColsSemantics(
 ): DataFrameDomain {
 	if(columns?.every(col => typeof col === 'string')) {
 		return new DataFrameDomain({
-			colnames: value.colnames.union([columns, []]),
+			colnames: value.colnames.union(setRange(columns)),
 			cols:     value.cols,
 			rows:     value.rows
 		});
@@ -166,7 +166,7 @@ function applyAssignColsSemantics(
 		const cols = columns.length;
 
 		return new DataFrameDomain({
-			colnames: value.colnames.union([columns, []]),
+			colnames: value.colnames.union(setRange(columns)),
 			cols:     value.cols.add([0, cols]).max([cols, cols]),
 			rows:     value.rows
 		});
@@ -217,7 +217,7 @@ function applySetColNamesSemantics(
 	const allColNames = colnames?.every(isNotUndefined) && value.cols.value !== Bottom && colnames.length >= value.cols.value[1];
 
 	return new DataFrameDomain({
-		colnames: allColNames ? value.colnames.create([colnames, []]) : value.colnames.widenDown().union(setRange(colnames)).widenUp(),
+		colnames: allColNames ? value.colnames.create(setRange(colnames)) : value.colnames.widenDown().union(setRange(colnames)).widenUp(),
 		cols:     value.cols,
 		rows:     value.rows
 	});
@@ -408,7 +408,7 @@ function applySummarizeSemantics(
 	{ colnames }: { colnames: (string | undefined)[] | undefined }
 ): DataFrameDomain {
 	return new DataFrameDomain({
-		colnames: colnames !== undefined ? value.colnames.join([[], []]).union(setRange(colnames)) : value.colnames.widenUp(),
+		colnames: colnames !== undefined ? value.colnames.join(setRange([])).union(setRange(colnames)) : value.colnames.widenUp(),
 		cols:     colnames !== undefined ? value.cols.add([0, colnames.length]).min([colnames.length, +Infinity]) : value.cols.widenUp(),
 		rows:     value.rows.min([1, +Infinity]).max([0, 1])
 	});
@@ -450,7 +450,7 @@ function applyJoinSemantics(
 		duplicateCols = commonCols !== Bottom ? commonCols !== Top ? [...commonCols] : undefined : [];
 		productRows = true;
 	} else if(by.every(isNotUndefined)) {
-		const remainingCols = value.colnames.intersect(other.colnames).subtract([by, []]).upper();
+		const remainingCols = value.colnames.intersect(other.colnames).subtract(setRange(by)).upper();
 		duplicateCols = remainingCols !== Bottom ? remainingCols !== Top ? [...remainingCols] : undefined : [];
 		productRows = false;
 	} else {
@@ -478,7 +478,7 @@ function applyJoinSemantics(
 	}
 	return new DataFrameDomain({
 		...value,
-		colnames: duplicateCols === undefined ? value.colnames.top() : duplicateCols.length > 0 ? value.colnames.union(other.colnames).subtract([duplicateCols, []]).widenUp() : value.colnames.union(other.colnames),
+		colnames: duplicateCols === undefined ? value.colnames.top() : duplicateCols.length > 0 ? value.colnames.union(other.colnames).subtract(setRange(duplicateCols)).widenUp() : value.colnames.union(other.colnames),
 		cols:     by !== undefined ? value.cols.add(other.cols).subtract([by.length, by.length]) : mergeInterval(value.cols, other.cols),
 		rows:     productRows ? productInterval(rows, value.rows, other.rows) : rows
 	});
@@ -503,5 +503,5 @@ function applyUnknownSemantics(
 function setRange(colnames: (string | undefined)[] | undefined): ArrayRangeValue<string> {
 	const names = colnames?.filter(isNotUndefined) ?? [];
 
-	return [names, names.length === colnames?.length ? [] : Top];
+	return { min: names, range: names.length === colnames?.length ? [] : Top };
 }
--- a/src/abstract-interpretation/domains/set-range-domain.ts
+++ b/src/abstract-interpretation/domains/set-range-domain.ts
@@ -8,10 +8,10 @@ import { SetComparator } from './satisfiable-domain';
 /* eslint-disable @typescript-eslint/unified-signatures */
 
 /** The Top element of the set range domain with an empty set as minimum set and {@link Top} as range set */
-export const SetRangeTop = [new Set<never>(), Top] as const satisfies SetRangeValue<unknown>;
+export const SetRangeTop = { min: new Set<never>(), range: Top } as const satisfies SetRangeValue<unknown>;
 
 /** The type of the actual values of the set range domain as tuple with a minimum set and range set of additional possible values (i.e. `[{"id","name"}, ∅]`, or `[{"id"}, {"score"}]`) */
-type SetRangeValue<T> = readonly [min: ReadonlySet<T>, range: ReadonlySet<T> | typeof Top];
+type SetRangeValue<T> = { readonly min: ReadonlySet<T>, readonly range: ReadonlySet<T> | typeof Top };
 /** The type of the Top element of the set range domain as tuple with the empty set as minimum set and {@link Top} as range set (i.e. `[∅, Top]`) */
 type SetRangeTop = typeof SetRangeTop;
 /** The type of the Bottom element of the set range domain as {@link Bottom} */
@@ -20,11 +20,11 @@ type SetRangeBottom = typeof Bottom;
 type SetRangeLift<T> = SetRangeValue<T> | SetRangeTop | SetRangeBottom;
 
 /** The type of the actual values of the set range domain as array tuple with a minimum array and range array for better readability (e.g. `[["id","name"], []]`, or `[["id"], ["score"]]`) */
-export type ArrayRangeValue<T> = readonly [min: T[], range: T[] | typeof Top];
+export type ArrayRangeValue<T> = { readonly min: T[], readonly range: T[] | typeof Top };
 
 /** The type for the maximum number of elements in the minimum set and maximum set of the set range domain before over-approximation */
-export type SetRangeLimit = readonly [min: number, range: number];
-const DefaultLimit = [DEFAULT_INFERENCE_LIMIT, DEFAULT_INFERENCE_LIMIT] as const satisfies SetRangeLimit;
+export type SetRangeLimit = { readonly min: number, readonly range: number };
+const DefaultLimit = { min: DEFAULT_INFERENCE_LIMIT, range: DEFAULT_INFERENCE_LIMIT } as const satisfies SetRangeLimit;
 
 /**
  * The set range abstract domain as range of possible value sets with a minimum set of values and a range of possible additional values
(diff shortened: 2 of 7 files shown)


src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -1127,20 +1131,20 @@ function mapDataFrameMutate(
 		});
 	}
 
-	if(deletedCols === undefined || deletedCols.length > 0) {
+	if(mutatedCols === undefined || mutatedCols.length > 0 || deletedCols?.length === 0) {
```
EagleoutIce, line 1130: Why do we *also* trigger this if `deletedCols` has no length? Maybe we could outsource these guards into well-named fns?

Yet, generally this implies some exclusivity. If, for example, we have `mutateCols` *and* `deleteCols` sometime in the future, there will always be a "non-maybe" mutation. Is this intended?

src/abstract-interpretation/data-frame/semantics.ts
```diff
@@ -165,19 +166,19 @@ function applyAssignColsSemantics(
 		const cols = columns.length;
 
 		return new DataFrameDomain({
-			colnames: value.colnames.join(columns),
+			colnames: value.colnames.union([columns, []]),
```
EagleoutIce: I am not sure whether I like this notation with an empty upper. Maybe move this implicitly into `union`?

src/abstract-interpretation/data-frame/semantics.ts
```diff
@@ -165,19 +166,19 @@ function applyAssignColsSemantics(
...
 			cols:     value.cols.add([0, cols]).max([cols, cols]),
 			rows:     value.rows
 		});
 	} else if(columns?.every(col => typeof col === 'number')) {
 		return new DataFrameDomain({
-			colnames: value.colnames.top(),
+			colnames: value.colnames.extendUp(),
```
EagleoutIce: `extendUp` feels limited in the name, i currently cannot see the impl, but maybe a `widen` terminology helps which uses a default? or `widenUp`?

src/abstract-interpretation/data-frame/semantics.ts
```diff
@@ -251,13 +252,13 @@ function applyRemoveColsSemantics(
 ): DataFrameDomain {
 	if(options?.maybe) {
 		return new DataFrameDomain({
-			colnames: colnames !== undefined ? value.colnames.subtract(colnames.filter(isNotUndefined)) : value.colnames,
+			colnames: colnames !== undefined ? value.colnames.subtract([colnames.filter(isNotUndefined) ?? [], colnames.every(isNotUndefined) ? [] : Top]) : value.colnames.extendDown(),
```
EagleoutIce: Given that you often check whether all are undefined *and* filter, why not have a common `filter` and then check for whether the length is `0`? this feels more readable

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+import { AbstractDomain, DEFAULT_INFERENCE_LIMIT, domainElementToString } from './abstract-domain';
+import { Bottom, Top } from './lattice';
+import type { SatisfiableDomain } from './satisfiable-domain';
+import { SetComparator } from './satisfiable-domain';
+/* eslint-disable @typescript-eslint/unified-signatures */
+
+/** The Top element of the set range domain with an empty set as minimum set and {@link Top} as range set */
+export const SetRangeTop = [new Set<never>(), Top] as const satisfies SetRangeValue<unknown>;
+
+/** The type of the actual values of the set range domain as tuple with a minimum set and range set of additional possible values */
```
EagleoutIce: Maybe you could add 1-2 examples here with `@example` to exemplify abstract values :)

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+import { setEquals } from '../../util/collections/set';
+import { Ternary } from '../../util/logic';
+import { AbstractDomain, DEFAULT_INFERENCE_LIMIT, domainElementToString } from './abstract-domain';
+import { Bottom, Top } from './lattice';
+import type { SatisfiableDomain } from './satisfiable-domain';
+import { SetComparator } from './satisfiable-domain';
+/* eslint-disable @typescript-eslint/unified-signatures */
+
+/** The Top element of the set range domain with an empty set as minimum set and {@link Top} as range set */
+export const SetRangeTop = [new Set<never>(), Top] as const satisfies SetRangeValue<unknown>;
```
EagleoutIce: We could also try to `Object.freeze` these but there probably woudln't be that much benefit

EagleoutIce: \+ any benefit in outsourcing `Set<never>()` into a `EmptySet`?

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+/** The type of the actual values of the set range domain as tuple with a minimum set and range set of additional possible values */
+type SetRangeValue<T> = readonly [min: ReadonlySet<T>, range: ReadonlySet<T> | typeof Top];
+/** The type of the Top element of the set range domain as tuple with the empty set as minimum set and {@link Top} as range set (i.e. `[∅, Top]`) */
+type SetRangeTop = typeof SetRangeTop;
+/** The type of the Bottom element of the set range domain as {@link Bottom} */
+type SetRangeBottom = typeof Bottom;
+/** The type of the abstract values of the set range domain that are Top, Bottom, or actual values */
+type SetRangeLift<T> = SetRangeValue<T> | SetRangeTop | SetRangeBottom;
+
+export type ArrayRangeValue<T> = readonly [min: T[], range: T[] | typeof Top];
```
EagleoutIce: Any reason for this to be without a doc comment? it is the only exported type :P so far :D

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+
+export type SetRangeLimit = readonly [min: number, range: number];
```
EagleoutIce: Likewise here

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+	 * @param limit -  A limit for the maximum number of elements to store in the set
+	 * @param newSet - An optional set constructor for the domain elements if the type `T` is not storable in a HashSet
+	 */
+	constructor(value: Value | ArrayRangeValue<T>, limit: SetRangeLimit | number = DefaultLimit, setType: typeof Set<T> = Set) {
+		limit = typeof limit === 'number' ? [limit, limit] : limit;
+
+		if(value !== Bottom) {
+			const sizeMin = Array.isArray(value[0]) ? value[0].length : value[0].size;
+			const sizeRange = value[1] === Top ? Infinity : Array.isArray(value[1]) ? value[1].length : value[1].size;
+			const min = sizeMin > limit[0] ? new setType([...value[0]].slice(0, limit[0])) : new setType(value[0]);
```
EagleoutIce: Why the spread for value?

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+			const range = (value[1] === Top || sizeRange > limit[1] || sizeMin + sizeRange > limit[0] + limit[1]) ? Top : new setType([...value[0], ...value[1]]).difference(min);
```
EagleoutIce: Likeise here, respreading the values feels unnecessary
+ the guard is veeery long, maybe outsource for readability?

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	/**
+	 * The maximum set of the set range representing all values that can possibly exist (union of {@link min} and range).
+	 */
+	public get max(): Value extends SetRangeBottom ? SetRangeValue<T>[1] | typeof Bottom : SetRangeValue<T>[1] {
+		if(this.value === Bottom) {
+			return Bottom as Value extends SetRangeBottom ? SetRangeValue<T>[1] | typeof Bottom : SetRangeValue<T>[1];
+		}
+		return this.value[1] === Top ? Top : this.value[0].union(this.value[1]);
```
EagleoutIce: Given that this *does* computation I am not a fan of having this being a mere `getter`

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+
+export type ArrayRangeValue<T> = readonly [min: T[], range: T[] | typeof Top];
+
+export type SetRangeLimit = readonly [min: number, range: number];
+const DefaultLimit = [DEFAULT_INFERENCE_LIMIT, DEFAULT_INFERENCE_LIMIT] as const satisfies SetRangeLimit;
+
+/**
+ * The set range abstract domain as range of possible value sets with a minimum set of values and a range of possible additional values
+ * (similar to an interval like structure with a lower bound and a difference to the upper bound).
+ * The Bottom element is defined as {@link Bottom} symbol and the Top element is defined as the range `[∅, Top]` where the minimum set is the empty set and the range is {@link Top}.
```
EagleoutIce, line 32: Maybe add a small note that all operations do not modify the domain in-place but `#create` a new one?

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+	implements SatisfiableDomain<ReadonlySet<T>> {
+
+	public readonly limit:    SetRangeLimit;
+	private readonly setType: typeof Set<T>;
+
+	/**
+	 * @param limit -  A limit for the maximum number of elements to store in the set
+	 * @param newSet - An optional set constructor for the domain elements if the type `T` is not storable in a HashSet
+	 */
+	constructor(value: Value | ArrayRangeValue<T>, limit: SetRangeLimit | number = DefaultLimit, setType: typeof Set<T> = Set) {
```
EagleoutIce, line 47: You do a lot of convenience checks here, but *also* use this for the hot function `.create`.
How about outsourcing some of the checks into a `.make` or something factory function (e.g. the `limit` check) and remove them from the then private/more direct ctor to make all of these operations and object re-creations more performant?

src/abstract-interpretation/domains/set-range-domain.ts
```diff
@@ -0,0 +1,421 @@
...
+		}
+		const min = this.value[0].values().toArray();
+		const range = this.value[1] === Top ? this.value[1].description : this.value[1].values().toArray();
+
+		return [min, range];
+	}
+
+	public toString(): string {
+		if(this.value === Bottom) {
+			return '⊥';
```
EagleoutIce: Why isn't there are global string representation for these? 
like `Bottom.toString()`, or at least a `Domains.bottomString()`

src/benchmark/summarizer/first-phase/process.ts
```diff
@@ -250,26 +253,29 @@ function summarizeDfShapeStats({ perNodeStats, ...stats }: SlicerStatsDfShape):
 		numberOfEntriesPerNode:   summarizeMeasurement(nodeStats.map(s => s.numberOfEntries)),
 		numberOfOperations:       arraySum(nodeStats.map(s => s.mappedOperations?.length).filter(isNotUndefined)),
 		numberOfTotalValues:      nodeStats.filter(s => isValue(s.inferredColNames) && isValue(s.inferredColCount) && isValue(s.inferredRowCount)).length,
-		numberOfTotalTop:         nodeStats.filter(s => isTop(s.inferredColNames) && isTop(s.inferredColCount) && isTop(s.inferredRowCount)).length,
 		numberOfTotalBottom:      nodeStats.filter(s => s.inferredColNames === 0 && isBottom(s.inferredColCount) && isBottom(s.inferredRowCount)).length,
+		numberOfTotalTop:         nodeStats.filter(s => isTop(s.inferredColNames) && isTop(s.inferredColCount) && isTop(s.inferredRowCount)).length,
```
EagleoutIce, line 254: we have to change this lol xD 
This is waaay to much object clutter if we keep adding keys 🤣

EagleoutIce (commented):
This is just a general quick readover for now, I had no time to actually read and follow all of the lattice op impls

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: I just flipped the order of the `mutateCols` and `removeCols` operation. The guard with `deletedCols?.length === 0` ensures that we map the function to any abstract operation (in this case `mutateCols`) if we have an empty `mutate` call, such as `mutate(df)`.

Mutate is never "maybe", as the assigned column names have to exist after the function call (only the column removal by setting a column to `NULL` is "maybe")

src/abstract-interpretation/data-frame/semantics.ts
OliverGerstl: I want to make explicit that the accessed colnames are unioned with the minimum set and not only range/maximum set

src/abstract-interpretation/data-frame/semantics.ts
OliverGerstl: They have different semantics: if the whole array is `undefined` (i.e. `Top`), we don't even know the length, and, if only elements are `undefined` (i.e. `Top`), we may have partial information of the column names and know the length.

I hope this gets more intuitive when I move the `undefined` values to `Top` in #1791

OliverGerstl: But I will outsource the duplicate code into a separate function

src/abstract-interpretation/domains/set-range-domain.ts
OliverGerstl: I think leaving `Set<never>` is better to avoid confusions between `EmptySet` and empty sets resulting from intersections etc.

src/abstract-interpretation/domains/set-range-domain.ts
OliverGerstl: That was because of the fact that `value` could be a tuple of arrays or sets. But I replaced the array `slice()` with set `values().take()` and the second spread with set `union()`

src/abstract-interpretation/data-frame/semantics.ts
EagleoutIce: fine, but it feels weird to have a `[]` as the second argument without any semantic meaning

OliverGerstl: Yes, but I found `[columns, []]` more intuitive to read as something like `{min: new Set(columns), range: new Set<never>}`, since the former looks more like an interval

src/abstract-interpretation/data-frame/semantics.ts
OliverGerstl: I am not really happy with `widenUp` and `widenDown` either, as they don't really have anything to do with the classical widening, but they're probably still better than `extendUp` and `extendDown`

src/abstract-interpretation/data-frame/semantics.ts
EagleoutIce: Maybe we can discuss this later i have the feeling that something like `[columns, EmpyRange]` is still more readable

src/abstract-interpretation/domains/set-range-domain.ts
OliverGerstl: I can move this into a function `max()` or so, but I like the idea of providing a getter for the minimum and maximum set

src/abstract-interpretation/data-frame/semantics.ts
EagleoutIce: I mean they can be considered widenings with `[colnames.max, +Inf]` and `[-Inf, colnames.min]`, but we can discuss this too

src/abstract-interpretation/domains/set-range-domain.ts
OliverGerstl: The problem is that I need these checks also for the `create` calls in `join`, `meet` etc., as the number of elements may increase through `join`, `meet` etc., and I think that the overhead due to the new set construction is much higher than due to the limit checks anyways

src/benchmark/summarizer/first-phase/process.ts
OliverGerstl: Yes, but I think this is a general issue of the current benchmarking system

PR #2034, opened 2025-11-18
[Project] Move built-in environment into the analyzer context

### LINKED ISSUE #2027: [Project] Move built-in environment into the analyzer context

MaxAtoms:
Closes #2027

### DIFF
--- a/src/project/context/flowr-analyzer-environment-context.ts
+++ b/src/project/context/flowr-analyzer-environment-context.ts
@@ -60,11 +60,11 @@ export class FlowrAnalyzerEnvironmentContext implements ReadOnlyFlowrAnalyzerEnv
 		this.emptyBuiltInEnv.memory = builtIns.emptyBuiltInMemory;
 	}
 
-	public get builtInEnvironment(): IEnvironment {
+	public get builtInEnvironment(): DeepReadonly<IEnvironment> {
 		return this.builtInEnv;
 	}
 
-	public get emptyBuiltInEnvironment(): IEnvironment {
+	public get emptyBuiltInEnvironment(): DeepReadonly<IEnvironment> {
 		return this.emptyBuiltInEnv;
 	}
 
--- a/src/dataflow/environments/remove.ts
+++ b/src/dataflow/environments/remove.ts
@@ -3,11 +3,12 @@ import type { Identifier } from './identifier';
 import { happensInEveryBranch } from '../info';
 import type { NodeId } from '../../r-bridge/lang-4.x/ast/model/processing/node-id';
 import { cloneEnvironmentInformation } from './clone';
+import type { DeepReadonly } from 'ts-essentials';
 
 /**
  * Removes all definitions of a given name from the environment.
  */
-export function remove(name: Identifier, environment: REnvironmentInformation, defaultEnvironment: IEnvironment): REnvironmentInformation {
+export function remove(name: Identifier, environment: REnvironmentInformation, defaultEnvironment: DeepReadonly<IEnvironment>): REnvironmentInformation {
 	let current: IEnvironment = environment.current;
 	do{
 		const definition = current.memory.get(name);


src/project/context/flowr-analyzer-context.ts
```diff
@@ -41,6 +42,10 @@ export interface ReadOnlyFlowrAnalyzerContext {
 	 * The dependencies context provides access to the identified dependencies and their versions.
 	 */
 	readonly deps:   ReadOnlyFlowrAnalyzerDependenciesContext;
+	/**
+	 * The environment context provides access to the R environment information used during analysis.
+	 */
+	readonly env:    FlowrAnalyzerEnvironmentContext;
```
EagleoutIce: I think this shod also expose a readonly version that returns a DeepReadonly Environment

src/dataflow/extractor.ts
```diff
@@ -109,8 +105,7 @@ export function produceDataFlowGraph<OtherInfo>(
 	const dfData: DataflowProcessorInformation<OtherInfo & ParentInformation> = {
 		parser,
 		completeAst,
-		environment:         env,
-		builtInEnvironment:  env.current.parent,
+		environment:         initializeCleanEnvFromBuiltIn(ctx.env.builtInEnvironment),
```
EagleoutIce: Maybe rewrite this so there is `ctx.env.getCleanBuiltIn()`?

EagleoutIce: Likewise with empty

MaxAtoms: Should `getCleanBuiltIn` provide an empty environment with the built-in environment as parent? In that case, I would prefer a method name like `getCleanEnv`. `getCleanBuiltIn` sounds like it would directly return the built-in environment without additional wrapping

MaxAtoms: By `empty` you mean that the environment context should always also provide access to the `emptyBuiltInMemory`, regardless of configuration?

EagleoutIce: I am fine with `cleanEnv`, likewise with empty i mean that this should be the same for the emptyBuiltIn environment, I am unsure what you mean with regardless of configuration, we need both during the analysis

MaxAtoms: > I am unsure what you mean with regardless of configuration, we need both during the analysis

Oh right, I missed that 😬

src/project/context/flowr-analyzer-environment-context.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * Create a new {@link REnvironmentInformation|environment} with the configured built-in environment as base.
+	 */
+	getCleanEnv(builtInEnv: IEnvironment): REnvironmentInformation;
+
+	/**
+	 * Create a new {@link REnvironmentInformation|environment} with an empty built-in environment as base.
+	 */
+	getCleanEnvWithEmptyBuiltIns(): REnvironmentInformation;
```
EagleoutIce: this again feels weird we should have an empty environment that still contains the non-primitives:

https://github.com/flowr-analysis/flowr/blob/5509b9e347cb8b3832bdb6b4467bae74f59bb11b/src/dataflow/environments/environment.ts#L140

the `empty` env is not really fully empty (and a legacy name) *but* a subset:

https://github.com/flowr-analysis/flowr/blob/5509b9e347cb8b3832bdb6b4467bae74f59bb11b/src/dataflow/environments/built-in.ts#L315-L317

and we need both to be creatable from the context

src/dataflow/graph/dataflowgraph-builder.ts
```diff
@@ -14,13 +24,17 @@ import type { FlowrSearchLike } from '../../search/flowr-search-builder';
...
+import type { FlowrAnalyzerContext } from '../../project/context/flowr-analyzer-context';
 
 /**
- *
+ * Creates an empty dataflow graph.
+ * Should only be used in tests and documentation.
  */
 export function emptyGraph(idMap?: AstIdMap) {
-	return new DataflowGraphBuilder(idMap);
+	const ctx = new FlowrAnalyzerEnvironmentContext({ config: defaultConfigOptions } as FlowrAnalyzerContext);
```
MaxAtoms: Is this approach okay considering that it is only code for test/documentation?

src/benchmark/slicer.ts
```diff
@@ -339,7 +347,7 @@ export class BenchmarkSlicer {
 		};
 		this.perSliceMeasurements.set(slicingCriteria, stats);
 
-		this.executor.updateRequest({ criterion: slicingCriteria });
+		this.executor.updateRequest({ criterion: slicingCriteria, context: this.context as ReadOnlyFlowrAnalyzerContext });
```
EagleoutIce: Why does the update request need to pass the ctx reference?

src/control-flow/cfg-dead-code.ts
```diff
@@ -101,7 +102,8 @@ class CfgConditionalDeadCodeRemoval extends SemanticCfgGuidedVisitor {
 			graph:   this.config.dfg,
 			full:    true,
 			idMap:   this.config.normalizedAst.idMap,
-			resolve: this.config.ctx.config.solver.variables
+			resolve: this.config.ctx.config.solver.variables,
```
EagleoutIce: Maybe we can make this the default if one passes the context?

src/core/steps/all/static-slicing/00-slice.ts
```diff
@@ -13,6 +14,8 @@ export interface SliceRequiredInput {
 	readonly threshold?: number
```
EagleoutIce, line 14: This should be defaulted by the context too

src/dataflow/eval/resolve/resolve.ts
```diff
@@ -70,16 +71,17 @@ export function resolveNode(resolve: VariableResolve, a: RNodeWithParent, env?:
  * by recursively resolving the values of the arguments by calling {@link resolveIdToValue}
  * @param resolve - Variable resolve mode
  * @param node    - Node of the vector function to resolve
- * @param env     - Environment to use
+ * @param environment     - Environment to use
```
EagleoutIce: now the indentation is broken :/ sadly no linter rule appears to be capable (atm) to do the layouting *without* doing anything else

src/dataflow/graph/dataflowgraph-builder.ts
```diff
@@ -31,6 +45,13 @@ export type DataflowGraphEdgeTarget = NodeId | (readonly NodeId[]);
  * simplifies writing tests for dataflow graphs.
  */
 export class DataflowGraphBuilder extends DataflowGraph {
+	private defaultEnvironment: REnvironmentInformation;
+
+	constructor(cleanEnv: REnvironmentInformation, idMap?: AstIdMap) {
```
EagleoutIce: Given that the builder is for tests this should also have a default initializer so that we do *not* have to pass this all the time

src/dataflow/graph/dataflowgraph-builder.ts
EagleoutIce: Please create a context from input, not calling the class directly, this is safer (i.e., use either from sources without files or from input without input)

src/dataflow/graph/dataflowgraph-builder.ts
```diff
@@ -169,7 +190,7 @@ export class DataflowGraphBuilder extends DataflowGraph {
 		}, {
 			...info,
 			cds: info?.cds?.map(c => ({ ...c, id: normalizeIdToNumberIfPossible(c.id) }))
-		} as Partial<DataflowGraphVertexUse>), asRoot);
+		} as Partial<DataflowGraphVertexUse>), this.defaultEnvironment, asRoot);
```
EagleoutIce: why do we need the default env to a this method? we can add `this.defaultEnvironment` within the `addVertex` method

src/dataflow/internal/process/functions/call/built-in/built-in-function-definition.ts
```diff
@@ -122,7 +126,7 @@ export function processFunctionDefinition<OtherInfo>(
 		cds:         data.controlDependencies,
 		subflow:     flow,
 		exitPoints:  exitPoints?.filter(e => e.type === ExitPointType.Return || e.type === ExitPointType.Default).map(e => e.nodeId) ?? []
-	});
+	}, data.ctx.env.getCleanEnv());
```
EagleoutIce: thinking about it: maybe rename these to `makeCleanEnv()`? given that the return new instances

src/documentation/wiki-analyzer.ts
```diff
@@ -434,6 +436,17 @@ Probably the most important method is
 ${ctx.linkM(FlowrAnalyzerDependenciesContext, 'getDependency', { codeFont: true, realNameWrapper: 'i' })}
 that allows you to query for a specific dependency by name.
 
+${section('Environment Context', 3)}
+
+Here is the structure of the ${ctx.link(FlowrAnalyzerEnvironmentContext.name)} that provides access to the built-in environment:
```
EagleoutIce: you do not have to pass `.name`:

```suggestion
Here is the structure of the ${ctx.link(FlowrAnalyzerEnvironmentContext)} that provides access to the built-in environment:
```

src/documentation/wiki-analyzer.ts
```diff
@@ -434,6 +436,17 @@ Probably the most important method is
...
+
+${ctx.hierarchy(FlowrAnalyzerEnvironmentContext, { showImplSnippet: false })}
+
+The environment context provides access to the built-in environment via
+${ctx.link(FlowrAnalyzerEnvironmentContext.name + '::' + FlowrAnalyzerEnvironmentContext.prototype.getCleanEnv.name, { codeFont: true, realNameWrapper: 'i' })}.
```
EagleoutIce: Likewise, you can use `linkM`:


```suggestion
${ctx.linkM(FlowrAnalyzerEnvironmentContext, 'getCleanEnv', { codeFont: true, realNameWrapper: 'i' })}.
```

src/documentation/wiki-analyzer.ts
```diff
@@ -434,6 +436,17 @@ Probably the most important method is
...
+It also provides the empty built-in environment, which only contains primitives, via
+${ctx.link(FlowrAnalyzerEnvironmentContext.name + '::' + FlowrAnalyzerEnvironmentContext.prototype.getCleanEnvWithEmptyBuiltIns.name, { codeFont: true, realNameWrapper: 'i' })}.
```
EagleoutIce:
```suggestion
${ctx.linkM(FlowrAnalyzerEnvironmentContext, 'getCleanEnvWithEmptyBuiltIns', { codeFont: true, realNameWrapper: 'i' })}.
```

src/linter/rules/absolute-path.ts
```diff
@@ -81,19 +89,21 @@ function buildQuickFix(str: RNode | undefined, filePath: string, wd: string | un
 	}];
 }
 
+type PathFunction = (df: DataflowGraph, vtx: DataflowGraphVertexFunctionCall, config: FlowrConfigOptions, ctx: ReadOnlyFlowrAnalyzerContext) => string[] | undefined;
+
 /** return all strings constructable by these functions */
-const PathFunctions: Record<string, (df: DataflowGraph, vtx: DataflowGraphVertexFunctionCall, config: FlowrConfigOptions) => string[] | undefined> = {
-	'file.path': (df: DataflowGraph, vtx: DataflowGraphVertexFunctionCall, config: FlowrConfigOptions): string[] | undefined => {
+const PathFunctions: Record<string, PathFunction> = {
+	'file.path': (df: DataflowGraph, vtx: DataflowGraphVertexFunctionCall, config: FlowrConfigOptions, ctx: ReadOnlyFlowrAnalyzerContext): string[] | undefined => {
```
EagleoutIce: Given that we also pass the `ctx` please remove the `config` argument (it should be replaced as the ctx contains the config)

src/project/context/flowr-analyzer-environment-context.ts
```diff
@@ -0,0 +1,92 @@
...
+		const builtIns = getBuiltInDefinitions(builtInsConfig.definitions, builtInsConfig.loadDefaults);
+
+		this.builtInEnv = new Environment(undefined as unknown as IEnvironment, true);
+		this.builtInEnv.memory = builtIns.builtInMemory;
+
+		this.emptyBuiltInEnv = new Environment(undefined as unknown as IEnvironment, true);
+		this.emptyBuiltInEnv.memory = builtIns.emptyBuiltInMemory;
+	}
+
+	public get builtInEnvironment(): IEnvironment {
```
EagleoutIce: I think the type should be a `DeepReadonly<IEnvironment>` for both

src/slicing/static/static-slicer.ts
```diff
@@ -28,12 +30,13 @@ export const slicerLogger = log.getSubLogger({ name: 'slicer' });
...
  */
 export function staticSlice(
+	ctx: ReadOnlyFlowrAnalyzerContext,
 	info: DataflowInformation,
 	{ idMap }: NormalizedAst,
 	criteria: SlicingCriteria,
 	direction: SliceDirection,
 	threshold = 75,
-	cache?: Map<Fingerprint, Set<NodeId>>
+	cache?: Map<Fingerprint, Set<NodeId>>,
```
EagleoutIce: Pffh, maybe we make this a linter rule, could you add it as an issue with the flowr-lint repo?

test/functionality/_helper/shell.ts
```diff
@@ -42,8 +46,8 @@ import { resolveByName } from '../../../src/dataflow/environments/resolve-by-nam
 import type { GraphDifferenceReport, ProblematicDiffInfo } from '../../../src/util/diff-graph';
 import { extractCfg } from '../../../src/control-flow/extract-cfg';
 import { cfgToMermaidUrl } from '../../../src/util/mermaid/cfg';
-import { type CfgProperty , assertCfgSatisfiesProperties } from '../../../src/control-flow/cfg-properties';
-import { type FlowrConfigOptions , cloneConfig, defaultConfigOptions } from '../../../src/config';
+import { assertCfgSatisfiesProperties, type CfgProperty } from '../../../src/control-flow/cfg-properties';
```
EagleoutIce, line 49: What is the idea behind these reorderings?

test/functionality/dataflow/environments/resolve.test.ts
```diff
@@ -69,17 +69,19 @@ describe.sequential('Resolve', withShell(shell => {
 		const effectiveName = decorateLabelContext(label(name), ['resolve']);
 
 		test(effectiveName, async() => {
+			const context = contextFromInput(code.trim());
 			const dataflow = await new PipelineExecutor(DEFAULT_DATAFLOW_PIPELINE, {
 				parser:  shell,
-				context: contextFromInput(code.trim()),
+				context: context,
```
EagleoutIce:
```suggestion
				context,
```

test/functionality/slicing/slicing.bench.ts
```diff
@@ -37,15 +36,14 @@ for(i in 1:5) {
...
+				analyzer = await new FlowrAnalyzerBuilder()
 					.setParser(exec)
 					.build();
 				analyzer.addRequest(request);
-				result = { dataflow: await analyzer.dataflow(), normalize: await analyzer.normalize() };
 				ids = (await analyzer.runSearch(Q.var('print').first())).getElements().map(n => n.node.info.id);
 			}
 			guard(ids !== undefined, () => 'no result');
-			staticSlice(result.dataflow, result.normalize, [`$${ids[0]}`], SliceDirection.Backward, threshold);
+			staticSlice(analyzer.inspectContext(), await analyzer.dataflow(), await analyzer.normalize(), [`$${ids[0]}`], SliceDirection.Backward, threshold);
```
EagleoutIce: why the rewrite? please undo, this bench is supposed to test just the slice and now it indirectly relies on the caching capabilities of the tape rerun. The context can just be added to the object.

test/functionality/_helper/shell.ts
MaxAtoms: Apparently, I had "Optimize Imports" active on save in WebStorm 😬

test/functionality/slicing/slicing.bench.ts
MaxAtoms: Sorry, I did not think this through

src/slicing/static/static-slicer.ts
MaxAtoms: We do not want to have the trailing comma, right?

src/benchmark/slicer.ts
MaxAtoms: I might be missing the point here, but should `updateRequest` take a `Partial`?

I.e. change the signature to `PipelineExecutor::updateRequest(newRequestData: Partial<PipelinePerRequestInput<P>>): void`?

src/dataflow/graph/dataflowgraph-builder.ts
MaxAtoms: `addVertex` is implemented in the base class, but the `defaultEnvironment` property is only available in the derived builder class. As there is no (proper) method overloading in JS, I resorted to passing the property. I don't have a more elegant idea currently. What solution do you suggest?

src/benchmark/slicer.ts
EagleoutIce: Only the context should be optional I think, but semantically a partial would be fine too even though we wouldn't need it

EagleoutIce: Only the context should be optional I think, but semantically a partial would be fine too even though we wouldn't need it

src/dataflow/graph/dataflowgraph-builder.ts
EagleoutIce: Overwrite with a super delegate for the method?

EagleoutIce: Overwrite with a super delegate for the method?

src/slicing/static/static-slicer.ts
EagleoutIce: I am not against it, I just want it to be consistent

test/functionality/_helper/shell.ts
EagleoutIce: Fine :p

EagleoutIce: Fine :p

src/slicing/static/static-slicer.ts
MaxAtoms: https://github.com/flowr-analysis/flowr-lint/issues/17

PR #2040, opened 2025-11-20
[REPL] Improve error message if no requests are present!

EagleoutIce:


src/queries/catalog/linter-query/linter-query-format.ts
```diff
@@ -135,7 +149,8 @@ function addLintingRuleResult<Name extends LintingRuleNames>(ruleName: Name, res
 	result.push(`   ╰ **${rule.info.name}** (${ruleName}):`);
 
 	if(isLintingResultsError(results)) {
-		result.push(`       ╰ Error during execution of rule: ${results.error}`);
+		const error = results.error.includes('At least one request must be set') ? 'No requests to lint for were found in the analysis.' : 'Error during execution of rule: ' + results.error;
```
MaxAtoms, line 152: I'm not happy with this concrete string matching either, but I don't have an idea for a clean solution right now

PR #2073, opened 2025-12-04
[AbsInt] Create a generic CFG visitor for abstract interpretation

OliverGerstl:
- Create a generic CFG visitor for abstract interpretation independently of the data frame shape inference
- Store the inferred constraints separately (outside the AST nodes)
- Fix the widening mechanism of the abstract interpretation visitor
- Adapt the data frame shape inference to use the new abstract interpretation visitor

### DIFF
--- a/src/abstract-interpretation/absint-visitor.ts
+++ b/src/abstract-interpretation/absint-visitor.ts
@@ -21,13 +21,15 @@ extends Omit<SemanticCfgGuidedVisitorConfiguration<OtherInfo, ControlFlowInforma
 
 /**
  * A control flow graph visitor to perform abstract interpretation.
+ *
+ * However, the visitor does not yet support inter-procedural abstract interpretation and abstract condition semantics.
  */
 export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDomain, OtherInfo = NoInfo, Config extends AbsintVisitorConfiguration<Domain, OtherInfo> = AbsintVisitorConfiguration<Domain, OtherInfo>>
 	extends SemanticCfgGuidedVisitor<OtherInfo, ControlFlowInformation, NormalizedAst<OtherInfo>, DataflowGraph, Config & { defaultVisitingOrder: 'forward', defaultVisitingType: 'exit' }> {
 	/**
-	 * The state of the abstract interpretation visitor mapping node IDs to the abstract state at the respective node.
+	 * The abstract trace of the abstract interpretation visitor mapping node IDs to the abstract state at the respective node.
 	 */
-	protected readonly state: Map<NodeId, StateAbstractDomain<Domain>> = new Map();
+	protected readonly trace: Map<NodeId, StateAbstractDomain<Domain>> = new Map();
 
 	/**
 	 * A set of nodes representing variable definitions that have already been visited but whose assignment has not yet been processed.
@@ -35,69 +37,62 @@ export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDo
 	private readonly unassigned: Set<NodeId> = new Set();
 
 	/**
-	 * The current state domain at the currently processed AST node.
+	 * The current abstract state domain at the currently processed AST node.
 	 */
-	private currentDomain: StateAbstractDomain<Domain>;
+	private currentState: StateAbstractDomain<Domain>;
 
 	constructor(config: Config) {
 		super({ ...config, defaultVisitingOrder: 'forward', defaultVisitingType: 'exit' });
 
-		this.currentDomain = config.domain.bottom();
+		this.currentState = config.domain.bottom();
 	}
 
 	/**
 	 * Resolves the inferred abstract value of an AST node.
 	 * This requires that the abstract interpretation visitor has been completed, or at least started.
-	 * @param id     - The ID of the node to get the inferred value for
-	 * @param domain - An optional state abstract domain used to resolve the inferred abstract value (defaults to the state at the requested node)
+	 * @param id    - The ID of the node to get the inferred value for
+	 * @param state - An optional state abstract domain used to resolve the inferred abstract value (defaults to the state at the requested node)
 	 * @returns The inferred abstract value of the node, or `undefined` if no value was inferred for the node
 	 */
-	public getValue(id: RNode<ParentInformation & OtherInfo> | NodeId | undefined, domain?: StateAbstractDomain<Domain>): Domain | undefined {
+	public getAbstractValue(id: RNode<ParentInformation & OtherInfo> | NodeId | undefined, state?: StateAbstractDomain<Domain>): Domain | undefined {
 		const node = (id === undefined || typeof id === 'object') ? id : this.getNormalizedAst(id);
-		domain ??= node !== undefined ? this.getState(node.info.id) : undefined;
+		state ??= node !== undefined ? this.getAbstractState(node.info.id) : undefined;
 
 		if(node === undefined) {
 			return;
-		} else if(domain?.has(node.info.id)) {
-			return domain.get(node.info.id);
+		} else if(state?.has(node.info.id)) {
+			return state.get(node.info.id);
 		}
 		const vertex = this.getDataflowGraph(node.info.id);
 		const call = vertex?.tag === VertexType.FunctionCall ? vertex : undefined;
 		const origins = Array.isArray(call?.origin) ? call.origin : [];
 
 		if(node.type === RType.Symbol) {
-			const values = this.getVariableOrigins(node.info.id).map(origin => domain?.get(origin));
+			const values = this.getVariableOrigins(node.info.id).map(origin => state?.get(origin));
 
 			if(values.length > 0 && values.every(isNotUndefined)) {
 				return AbstractDomain.joinAll(values);
 			}
 		} else if(node.type === RType.Argument && node.value !== undefined) {
-			return this.getValue(node.value, domain);
+			return this.getAbstractValue(node.value, state);
 		} else if(node.type === RType.ExpressionList && node.children.length > 0) {
-			return this.getValue(node.children[node.children.length - 1], domain);
-		} else if(node.type === RType.Pipe) {
-			return this.getValue(node.rhs, domain);
+			return this.getAbstractValue(node.children[node.children.length - 1], state);
 		} else if(origins.includes('builtin:pipe')) {
-			if(node.type === RType.BinaryOp) {
-				return this.getValue(node.rhs, domain);
+			if(node.type === RType.Pipe || node.type === RType.BinaryOp) {
+				return this.getAbstractValue(node.rhs, state);
 			} else if(call?.args.length === 2 && call?.args[1] !== EmptyArgument) {
-				return this.getValue(call.args[1].nodeId, domain);
+				return this.getAbstractValue(call.args[1].nodeId, state);
 			}
-		} else if(node.type === RType.IfThenElse) {
-			if(node.otherwise !== undefined) {
-				const values = [node.then, node.otherwise].map(entry => this.getValue(entry, domain));
+		} else if(origins.includes('builtin:if-then-else')) {
+			let values: (Domain | undefined)[] = [];
 
-				if(values.length > 0 && values.every(isNotUndefined)) {
-					return AbstractDomain.joinAll(values);
-				}
+			if(node.type === RType.IfThenElse && node.otherwise !== undefined) {
+				values = [node.then, node.otherwise].map(entry => this.getAbstractValue(entry, state));
+			} else if(call?.args.every(arg => arg !== EmptyArgument) && call.args.length === 3) {
+				values = call.args.slice(1, 3).map(entry => this.getAbstractValue(entry.nodeId, state));
 			}
-		} else if(origins.includes('builtin:if-then-else') && call?.args.every(arg => arg !== EmptyArgument)) {
-			if(call.args.length === 3) {
-				const values = call.args.slice(1, 3).map(entry => this.getValue(entry.nodeId, domain));
-
-				if(values.length > 0 && values.every(isNotUndefined)) {
-					return AbstractDomain.joinAll(values);
-				}
+			if(values.length > 0 && values.every(isNotUndefined)) {
+				return AbstractDomain.joinAll(values);
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -20,40 +20,38 @@ const SpecialAccessArgumentsMapper: Record<RIndexAccess['operator'], string[]> =
...
+ * @param ctx  - The current flowR analyzer context
+ * @returns The mapped abstract data frame operations for the access operation, or `undefined` if the node does not represent a data frame access operation
  */
 export function mapDataFrameAccess(
 	node: RNode<ParentInformation>,
+	inference: DataFrameShapeInferenceVisitor,
 	dfg: DataflowGraph,
 	ctx: ReadOnlyFlowrAnalyzerContext
-): DataFrameExpressionInfo | undefined {
+): DataFrameOperation[] | undefined {
```
EagleoutIce: Given the repeated use of this now canonical type, why not have a type alias for `DataFrameOperations` that just is [] or undefined?

src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -20,40 +20,38 @@ const SpecialAccessArgumentsMapper: Record<RIndexAccess['operator'], string[]> =
...
-	}
-	if(operations !== undefined) {
-		return { type: 'expression', operations: operations };
+		return mapDataFrameIndexColRowAccess(node, inference, resolveInfo);
 	}
 }
 
 function mapDataFrameNamedColumnAccess(
 	access: RNamedAccess<ParentInformation>,
+	inference: DataFrameShapeInferenceVisitor,
```
EagleoutIce, line 49: I'd call it just `infer` but fine :D

src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -20,40 +20,38 @@ const SpecialAccessArgumentsMapper: Record<RIndexAccess['operator'], string[]> =
...
 	info: ResolveInfo
```
EagleoutIce, line 50: if it is backward compatibility fine, but why dont all mappers get the read only analyzer and be done with it?

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -206,17 +205,17 @@ export function hasCriticalArgument(
 }
 
 /**
- * Checks if a given argument has any data frame shape information and therefore may represent a data frame.
- * @param arg  - The argument to check
- * @param info - Argument resolve information
- * @returns Whether the argument has any data frame shape information and may represent a data frame
+ * Checks if a given argument has inferred data frame shape information and therefore represents a data frame
```
EagleoutIce:
```suggestion
 * Checks if a given argument has an already inferred data frame shape information and therefore represents a data frame
```

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -206,17 +205,17 @@ export function hasCriticalArgument(
...
+ * @param arg       - The argument to check
+ * @param inference - The data frame shape inference visitor to use
+ * @returns Whether the argument has inferred data frame shape information and represents a data frame
```
EagleoutIce:
```suggestion
```

This feels redundant

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -631,27 +612,24 @@ export function mapDataFrameFunctionCall<Name extends DataFrameFunction>(
 		const args = getFunctionArguments(node, dfg);
 
 		if(hasCriticalArgument(args, params.critical, resolveInfo)) {
-			operations = [{ operation: 'unknown', operand: undefined }];
+			return [{ operation: 'unknown', operand: undefined }];
 		} else {
-			operations = mapper(args, params, resolveInfo, ctx);
+			return mapper(args, params, inference, resolveInfo, ctx);
```
EagleoutIce: Given that either: the resolve info contains the context it feels weird passing it separately.

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -10,9 +10,8 @@ import type { ParentInformation } from '../../../r-bridge/lang-4.x/ast/model/pro
 import { visitAst } from '../../../r-bridge/lang-4.x/ast/model/processing/visitor';
 import { RType } from '../../../r-bridge/lang-4.x/ast/model/type';
 import { RNull } from '../../../r-bridge/lang-4.x/convert-values';
-import type { AbstractInterpretationInfo } from '../absint-info';
 import { resolveIdToArgName, resolveIdToArgValue, unquoteArgument } from '../resolve-args';
-import { resolveIdToDataFrameShape } from '../shape-inference';
+import type { DataFrameShapeInferenceVisitor } from '../shape-inference';
 
 /** Regular expression representing valid columns names, e.g. for `data.frame` */
 const ColNamesRegex = /^[A-Za-z.][A-Za-z0-9_.]*$/;
```
EagleoutIce, line 17: we should now have a r-regex or something file that contains the standrad regexes - this would be a good fit there :D 

And just for completeness, this is what R actually checks:

```c
static SEXP ValidName(SEXP tag)
{
	if(isNull(tag) || !(isalpha(CHAR(tag)[0]) || (CHAR(tag)[0]) == '.'))  {
		char buf[10];
		sprintf(buf,"X%d", cnt+1);
		return install(buf);
	}
	else {
		SEXP d = duplicate(tag);
		char *p = CHAR(d);
		while(*p) {
			if(!isalnum(*p)) *p = '.';
			p++;
		}
		return install(CHAR(d));
	}
}
```

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -856,7 +839,7 @@ function mapDataFrameRowBind(
 
 	for(const arg of args) {
 		if(arg !== dataFrame && arg !== EmptyArgument) {
-			const otherDataFrame = resolveIdToDataFrameShape(arg.value, info.graph);
+			const otherDataFrame = inference.getValue(arg.value);
```
EagleoutIce: Maybe too nit.picky but why not specifically `getAbstractValue`

src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -1,118 +1,138 @@
...
- * @param ast    - The abstract syntax tree to resolve node IDs to AST nodes
- * @param ctx    - The current flowr analyzer context
- * @returns The abstract data frame state at the exit node of the control flow graph (see {@link DataFrameStateDomain}).
- * The abstract data frame states for all other nodes are attached to the AST.
+ * An abstract data frame operation.
+ * - `operation` contains the type of the abstract operation (see {@link DataFrameOperationName})
+ * - `operand` contains the ID of the data frame operand of the operation (may be `undefined`)
+ * - `type` optionally contains the constraint type to overwrite the default type of the operation (see {@link ConstraintType})
+ * - `options` optionally contains additional options for the abstract operation (see {@link DataFrameOperationOptions})
+ * - `...args` contains the arguments of the abstract operation (see {@link DataFrameOperationArgs})
```
EagleoutIce: if you make the property an interface you can write these comments to each property and still have a type with the key-dependent types

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+		} else if(origins.includes('builtin:if-then-else') && call?.args.every(arg => arg !== EmptyArgument)) {
+			if(call.args.length === 3) {
+				const values = call.args.slice(1, 3).map(entry => this.getValue(entry.nodeId, domain));
+
+				if(values.length > 0 && values.every(isNotUndefined)) {
+					return AbstractDomain.joinAll(values);
+				}
+			}
+		}
+	}
```
EagleoutIce, line 98: Maybe we should discuss this, this should be abstract into a generic argument retriever based on the config, that is also to be used by others, we now have this code partially mirrored in various locations (besides your abstract semantic stuffsies afterward)

EagleoutIce (commented):
Partial

src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -1,118 +1,138 @@
...
-		const values = getVariableOrigins(node.info.id, dfg).map(origin => domain.get(origin.info.id));
+	public getGlobalState(): ReadonlyMap<NodeId, DataFrameStateDomain> {
+		return this.state;
+	}
 
-		if(values.length > 0 && values.every(isNotUndefined)) {
-			return AbstractDomain.joinAll(values);
+	/**
+	 * Gets the mapped abstract data frame operations for an AST node.
+	 * This requires that the abstract interpretation visitor has been completed, or at least started..
```
EagleoutIce:
```suggestion
	 * This requires that the abstract interpretation visitor has been completed, or at least started.
```

EagleoutIce: Generally the question would be to what regard this respects inter-procedurality if this is a function call or how these handle virtual argument wrappers or things like expression lists - does it return *everything* happening within an expr list or nothing?

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+			if(node.otherwise !== undefined) {
+				const values = [node.then, node.otherwise].map(entry => this.getValue(entry, domain));
+
+				if(values.length > 0 && values.every(isNotUndefined)) {
+					return AbstractDomain.joinAll(values);
+				}
+			}
+		} else if(origins.includes('builtin:if-then-else') && call?.args.every(arg => arg !== EmptyArgument)) {
+			if(call.args.length === 3) {
+				const values = call.args.slice(1, 3).map(entry => this.getValue(entry.nodeId, domain));
```
EagleoutIce: Likewise duplication

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+		}
+	}
+
+	/**
+	 * Gets the inferred abstract state at the location of a specific AST node.
+	 * This requires that the abstract interpretation visitor has been completed, or at least started.
+	 * @param id - The ID of the node to get the abstract state at
+	 * @returns The abstract state at the node, or `undefined` if the node has no abstract state (i.e. the node has not been visited or is unreachable).
+	 */
+	public getState(id: NodeId | undefined): StateAbstractDomain<Domain> | undefined {
```
EagleoutIce: Partially this opens the same questions than the comment to getOperations

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+	public getResult(): StateAbstractDomain<Domain> {
+		const exitPoints = this.config.controlFlow.exitPoints.map(id => this.getCfgVertex(id)).filter(isNotUndefined);
+		const exitNodes = exitPoints.map(vertex => getVertexRootId(vertex)).filter(isNotUndefined);
+		const domains = exitNodes.map(node => this.getState(node)).filter(isNotUndefined);
+
+		return this.config.domain.bottom().joinAll(domains);
+	}
+
+	public override start(): void {
+		guard(this.state.size === 0, 'Abstract interpretation visitor has already been started');
```
EagleoutIce: is this a good guard? I mean in the current implementation yes, but if you want to have program-wide fixpoint iterations or maybe want to initialize an abstract pass with assumptions, prior results or e.g. magic comments this feels sloppy

+ given that this class is abtract maybe go via `this.prototype...` to print the classname?

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+			if(this.shouldWiden(vertex)) {
+				this.currentDomain = oldDomain.widen(this.currentDomain);
+			}
+			this.state.set(nodeId, this.currentDomain);
+
+			const visitedCount = this.visited.get(vertex.id) ?? 0;
+			this.visited.set(vertex.id, visitedCount + 1);
+
+			// only continue visiting if the widening point is visited for the first time or the abstract state at the widening point changed
+			return visitedCount === 0 || !oldDomain.equals(this.currentDomain);
```
EagleoutIce: maybe outsource the "did it change" handler for the fixpoint so we can abstract it in the future?

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -0,0 +1,337 @@
...
+
+	/** Gets each variable origin that has already been visited and whose assignment has already been processed */
+	protected getVariableOrigins(nodeId: NodeId): NodeId[] {
+		return getOriginInDfg(this.config.dfg, nodeId)
+			?.filter(origin => origin.type === OriginType.ReadVariableOrigin)
+			.map(origin => origin.id)
+			.filter(origin => this.state.has(origin) && !this.unassigned.has(origin)) ?? [];
+	}
+
+	/** We only perform widening at `for`, `while`, or `repeat` loops with more than one incoming CFG edge */
```
EagleoutIce, line 306: and what about recursion? an is widening point would work if flowR would gve you the convex hull of the recursion, but right now there is no handler for this; discuss?

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: I don't think we have this kind of value fold in flowR yet, at least, I did not find another location

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: I think this is fine to ensure that the visitor is not started multiple times. We can easily change the guard when we want to initialize the state before starting the visitor

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: Yes, but currently we also do not support interprocedural abstract interpretation.
I added a comment for that

src/abstract-interpretation/data-frame/mappers/access-mapper.ts
OliverGerstl: The data frame shape inference is not yet integrated in the flowR analyzer

OliverGerstl: I created an issue #2090

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: I created an issue #2091

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: I created an issue #2092

PR #2086, opened 2025-12-15
[Mermaid]: Support includeOnly and Mark for ast and cfg printers

### LINKED ISSUE #2080: [Mermaid]: Support mark and includeOnly as used in dfg printer
Add support for the same includeOnly and mark features for the AST and CFG printers as implemented by the dfg printer.

### LINKED ISSUE #2084: [Bug]: includeOnlyIds does not remove all edges when excluding nodes
### What happened?

includeOnlyIds should also exclude edges that reference excluded nodes
Otherwise only the id will show up in the resulting graph!

_No response_

v2.7.0 (Dec 3, 2025)

### Relevant log output

```Shell

```

### Meta Information

- [ ] The issue is urgent

gigalasr:
- Closes #2080 
- Closes #2084

### DIFF
--- a/src/cli/repl/commands/repl-cfg.ts
+++ b/src/cli/repl/commands/repl-cfg.ts
@@ -7,16 +7,15 @@ import type { NormalizedAst } from '../../../r-bridge/lang-4.x/ast/model/process
 import { type CfgSimplificationPassName , DefaultCfgSimplificationOrder } from '../../../control-flow/cfg-simplification';
 import type { ReadonlyFlowrAnalysisProvider } from '../../../project/flowr-analyzer';
 import { handleString } from '../core';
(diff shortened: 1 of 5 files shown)


test/system-tests/repl.test.ts
```diff
@@ -47,7 +47,7 @@ describe('repl', () => {
 					assert.include(output, expect, `output ${output} does not contain ${expect}`);
 				});
 				test(':n', () => {
-					const expect = normalized ? normalizedAstToMermaid(normalized.ast) : '';
+					const expect = normalized ? normalizedAstToMermaid(normalized.ast, {}) : '';
```
EagleoutIce, line 50: Only comment: please make the config itself optional, it feels weird to be required to pass in empty braces if you do not want to change the configuration + it breaks the api

PR #2089, opened 2025-12-16
[DF-Shape] Use a reduced product domain for the data frame shape

OliverGerstl:
- Refine the inferred column names and count of data frames via a reduced product domain


test/functionality/abstract-interpretation/domains/domain.ts
```diff
@@ -2,7 +2,7 @@ import { assert, test } from 'vitest';
 import { type AnyAbstractDomain, type ConcreteDomain } from '../../../../src/abstract-interpretation/domains/abstract-domain';
 import { Top, TopSymbol } from '../../../../src/abstract-interpretation/domains/lattice';
 
-const ConcretizationLimit = 20;
+const ConcretizationLimit = 12;
```
EagleoutIce, line 5: this feels like a weirder change :D but fine

OliverGerstl: Yes, this is just because I don't wanted to write all 14 concrete states for a state abstract domain value in the tests :)

PR #2094, opened 2025-12-18
[DF-Shape] Fix shape inference issues identified during correctness evaluation

OliverGerstl:
- Fix support for CR line termination in read CSV files
- Fix wrong over-approximation of row count after join


src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -1401,7 +1402,7 @@ function parseRequestContent(
 
 	switch(requestType) {
 		case 'text':
-			request.content.split('\n').forEach(parser);
+			unescapeSpecialChars(request.content).split(/\r\n|\r|\n/).forEach(parser);
```
EagleoutIce: I would outsource the line independent split - especially because above, we reduce the `text` check for `\n` but not for `\r`

OliverGerstl: This is because _readr_ only interprets the argument as text if it contains `\n` but not when it contains `\r`.
I will outsource the split

PR #2169, opened 2026-01-04
refactor: assign with more than 2 args are fine, we just loose precisen

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.8.1 (see [Release v2.8.1 (Fix: Call-Graph Construction)](https://github.com/flowr-analysis/flowr/releases/tag/v2.8.1))._

PR #2322, opened 2026-02-10
Generalize Wiki

MaxAtoms:
This PR tracks the necessary changes to the flowR wiki in order to generalize it for usage within multiple projects.

Locally, I currently have this hacky solution of symlinking the `dist/src` directory of flowR into the other project's `node_modules`:

```
> rm -r node_modules/@eagleoutice/flowr/
> ln -s ../../../flowr/dist/src ./node_modules/@eagleoutice/flowr
```


src/documentation/wiki-mk/doc-context.ts
```diff
@@ -230,7 +233,7 @@ export interface GeneralDocContext {
 	 * @param linkText - Optional text to display for the link. If not provided, the page name will be used.
 	 * @param segment  - An optional segment within the page to link to (e.g., a header anchor).
 	 */
-	linkPage(pageName: ValidWikiDocumentTargetsNoSuffix | keyof typeof ConstantWikiLinkInfo, linkText?: string, segment?: string): string;
+	linkPage<T extends DocMakerLike[]>(pageName: ValidWikiDocumentTargetsNoSuffix<T> | keyof typeof ConstantWikiLinkInfo, linkText?: string, segment?: string): string;
```
MaxAtoms, line 236: I wanted to have this generic on the interface level, but would always end up in a circular dependency between the individual `DocMaker` implementations and the `DocMakerLike[]` constraint

MaxAtoms: We should also consider how we want to (type-safely) link to the flowR wiki from other wikis

EagleoutIce: jeah, lets talk about that with time, i can also try to make this transition - getting this type safe with external projects is a little bit harder but easily doable (i imagine a generic on the interface level which defaults to flowR and that is then initialized type-wise with the respective context)

EagleoutIce:
it is fine so far but I dislike the generics directly at the use of e.g. `linkPage` my original idea for this was (@MaxAtoms) to keep the wiki pages locked in right now (as this is only a type system safety) but then other wiki pages could add another extension layer (and theoretically we can do the same for flowR if you want to keep the general structure type-safe) which then `as` casts the context to the desired type in the `text` receiver; wdyt?

EagleoutIce:
for now, we do not need this, but thank you for the effort! <3

PR #2340, opened 2026-02-17
RNode helper

EagleoutIce:
What do you think we need @OliverGerstl, @Ellpeck, @MaxAtoms, @gigalasr, @schuler-henry?

How about:

1. `visitAst` - as  an alias for the currently global function
2. `directChildren` - which is done by the visitor but could help/ease working with it 
3. `depth` - to get the total depth as you needed it @schuler-henry, but probably lazily too
4. `nesting` - to make this maybe a lazy calculation, accessing the current field
5. `roleInParent` - to make this maybe a lazy calculation?
6. `fullRange` - to *remove* this key from the current storage and calculate it this way on demand (also for expr lists etc.)
7. `fullLexeme` - to likewise *remove* this key from the current storage and calculate it on demand (again also for expr lists etc.)
8. `collectAllIds` - to alias the otherwise global method

For now I will not add the lazy calculators, we'll see whether there is real demand

also do this for RProject to have the wrappers

Do we need any more? are some unclear? do we not need one of them?
We could also add some basic constructors but I would defer this for the source transformation

### DIFF
--- a/src/benchmark/slicer.ts
+++ b/src/benchmark/slicer.ts
@@ -37,7 +37,6 @@ import {
 } from '../r-bridge/retriever';
 import type { PipelineStepNames, PipelineStepOutputWithName } from '../core/steps/pipeline/pipeline';
 import { collectAllSlicingCriteria, type SlicingCriteriaFilter } from '../slicing/criterion/collect-all';
-import { RType } from '../r-bridge/lang-4.x/ast/model/type';
 import { getSizeOfDfGraph, safeSizeOf } from './stats/size-of';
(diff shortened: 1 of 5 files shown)


EagleoutIce:
I think given the vast amount of tyes it would be better to have an `RFunctionCall.is` and leave the rnode helper for the classes? Something like an `isComment` which encompasses multiple rnode types

EagleoutIce:
This now also cleans the Normalized Ast Printer for the wiki :D

EagleoutIce:
_This pull request is included in v2.9.11 (see [Release v2.9.11 (RNode Helper Objects)](https://github.com/flowr-analysis/flowr/releases/tag/v2.9.11))._

PR #2359, opened 2026-02-23
2326 interval pentagon ai interval condition semantics

schuler-henry:
Condition semantics for: ==, <, >, is.na and their negations

### DIFF
--- a/src/abstract-interpretation/interval/condition-semantics.ts
+++ b/src/abstract-interpretation/interval/condition-semantics.ts
@@ -14,15 +14,15 @@ import { FloatingPointComparison } from '../../util/floating-point';
 import { getMin } from '../../util/numbers';
 
 const IntervalConditionSemanticsMapper = [
-	[Identifier.make('!'), unaryOpSemantics(applyNegatedIntervalConditionSemantics), unaryOpSemantics(applyIntervalConditionSemantics)],
-	[Identifier.make('('), unaryOpSemantics(applyIntervalConditionSemantics), unaryOpSemantics(applyNegatedIntervalConditionSemantics)],
-	[Identifier.make('=='), binaryOpSemantics(intervalEqualsOp), binaryOpSemantics(intervalNotEqualsOp)],
-	[Identifier.make('!='), binaryOpSemantics(intervalNotEqualsOp), binaryOpSemantics(intervalEqualsOp)],
-	[Identifier.make('>'), binaryOpSemantics(intervalGreaterOp), binaryOpSemantics(intervalLessEqualOp)],
-	[Identifier.make('>='), binaryOpSemantics(intervalGreaterEqualOp), binaryOpSemantics(intervalLessOp)],
-	[Identifier.make('<'), binaryOpSemantics(intervalLessOp), binaryOpSemantics(intervalGreaterEqualOp)],
-	[Identifier.make('<='), binaryOpSemantics(intervalLessEqualOp), binaryOpSemantics(intervalGreaterOp)],
-	[Identifier.make('is.na'), unaryOpSemantics(intervalIsNaFn), unaryOpSemantics(intervalNegatedIsNaFn)]
+	[Identifier.make('!'), unaryCondOpSemantics(applyNegatedIntervalConditionSemantics), unaryCondOpSemantics(applyIntervalConditionSemantics)],
+	[Identifier.make('('), unaryCondOpSemantics(applyIntervalConditionSemantics), unaryCondOpSemantics(applyNegatedIntervalConditionSemantics)],
+	[Identifier.make('=='), binaryCondOpSemantics(intervalEqualsOp), binaryCondOpSemantics(intervalNotEqualsOp)],
+	[Identifier.make('!='), binaryCondOpSemantics(intervalNotEqualsOp), binaryCondOpSemantics(intervalEqualsOp)],
+	[Identifier.make('>'), binaryCondOpSemantics(intervalGreaterOp), binaryCondOpSemantics(intervalLessEqualOp)],
+	[Identifier.make('>='), binaryCondOpSemantics(intervalGreaterEqualOp), binaryCondOpSemantics(intervalLessOp)],
+	[Identifier.make('<'), binaryCondOpSemantics(intervalLessOp), binaryCondOpSemantics(intervalGreaterEqualOp)],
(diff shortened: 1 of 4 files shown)


src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -60,6 +67,11 @@ export class DataFrameShapeInferenceVisitor extends AbstractInterpretationVisito
 		}
 	}
 
+	protected getBottomValue(): DataFrameDomain {
+		const maxColNames = this.config.ctx.config.abstractInterpretation.dataFrame.maxColNames;
+		return DataFrameDomain.bottom(maxColNames);
+	}
```
OliverGerstl, line 73: This is fine for now, but maybe we find a better solution before merging this into main :)

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -0,0 +1,325 @@
...
+}
+
+/**
+ * Guard for unary operators, filtering all calls with more/less than 1 argument or with undefined argument.
+ * If the call has exactly 1 defined argument, the provided unary operator semantics is applied to it.
+ * Otherwise, the state is returned unmodified and a warning is logged.
+ * @param unaryOperatorSemantics - The semantics to apply if the call has exactly 1 defined argument.
+ * @returns The semantics to apply for a unary operator call, which includes the guard for the number of arguments.
+ */
+function unaryOpSemantics(unaryOperatorSemantics: UnaryOpSemantics): NAryFnSemantics {
```
OliverGerstl: If you do not plan to merge them with the expression semantics functions, I would recommend renaming them, such that is is clear that they refer to condition semantics here and not to expression semantics

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -0,0 +1,325 @@
...
+}
+
+function intervalEqualsOp(leftNodeId: NodeId, rightNodeId: NodeId, state: MutableStateAbstractDomain<IntervalDomain> | undefined, visitor: NumericInferenceVisitor) {
+	const leftValue = visitor.getAbstractValue(leftNodeId, state);
+	const rightValue = visitor.getAbstractValue(rightNodeId, state);
+
+	let meet: IntervalDomain | undefined = undefined;
+	if(isNotUndefined(leftValue) || isNotUndefined(rightValue)) {
+		meet = AbstractDomain.meetAll([leftValue, rightValue].filter(isNotUndefined));
+	}
```
OliverGerstl, line 184: This is not really clear to me: If we infer `undefined` (i.e. global top) for one side of the equal comparison, we should be able to directly return the current `state` without changes, as we cannot say anything new about the state? Or do I miss something?

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -0,0 +1,325 @@
...
+
+	if(meet?.isBottom()) {
+		if(isNotUndefined(state)) {
+			return state.bottom();
+		}
+		return new MutableStateAbstractDomain(new Map(), true);
```
OliverGerstl: Maybe we can simplify this somehow, as you are using this at multiple locations?
```suggestion
        return state?.bottom() ?? MutableStateAbstractDomain.bottom();
```
This also does not have to be done now :)

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -0,0 +1,325 @@
...
+		return state;
+	}
+
+	if(leftValue.isValue() && rightValue.isValue()) {
+		const [a, b] = leftValue.value;
+		const [c, d] = rightValue.value;
+
+		if(c < b || FloatingPointComparison.isNearlyLess(c, b, leftValue.significantFigures) != Ternary.Never) {
+			const smallestSignificantFigures = getMin([leftValue.significantFigures, rightValue.significantFigures].filter(isNotUndefined));
+			const maxAC = a < c ? c : a;
```
OliverGerstl: Minor: Maybe rather `maxLowerBound` for `maxAC` and `minUpperBound` for `minBD` or so?

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -0,0 +1,325 @@
...
+	const argValue = visitor.getAbstractValue(argNodeId, state);
+
+	if(isUndefined(argValue)) {
+		return state;
+	}
+
+	return state.bottom();
+}
+
+function intervalNegatedIsNaFn(_argNodeId: NodeId, state: MutableStateAbstractDomain<IntervalDomain> | undefined) {
```
OliverGerstl: Maybe rename this to `intervalIdentity` or so to reuse this for other conditions later where you also cannot further constrain the state?

src/abstract-interpretation/interval/eval.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Replace `workingDir` content with all files from `folder` and instrument all .R files by inserting
+ * logging-statements for all POIs.
+ *
+ * Current POIs: Assignment calls (`<-`, `<<-`, `=`)
+ * @param folder - The source folder.
+ * @param workingDir - The target folder for the instrumented files.
+ * @param outputFolder - The folder where the csv files should be written to.
+ * @returns A promise that resolves when the instrumentation is complete.
+ */
+async function instrument(folder: string, workingDir: string, outputFolder: string) {
```
OliverGerstl, line 28: This is fine for know but later I would prefer this to be in a separate repo that includes flowR as dependency :)

schuler-henry: Sure thing, this is only a temporary solution, as the current changes are not part of any release, so I didn't know how to create a dependency to a not published flower branch. This file is not intended to ever be merged to main.

src/abstract-interpretation/interval/condition-semantics.ts
schuler-henry: Top is the neutral element of the meet.
If we know one side of the equals, we know that the other side must meet these constraints too, so if one side is undefined and the other is [1,1], we know that the undefined value must at least be [1,1].
So we do need to update the states, but I will refactor it so that this becomes a little bit clearer.

OliverGerstl: Ah, yes, you are right. I probably shouldn't do code reviews at 10 PM.
Yes, I think some refactoring and/or comments would make this clearer :)

PR #2398, opened 2026-03-19
Linter: Mismatch Roxygen Arguments

EagleoutIce:
recognize overdocumented and underdocumented parameters (only if there is a roxygen comment for the function)


src/linter/rules/roxygen-arguments.ts
```diff
@@ -0,0 +1,145 @@
...
+	if(comment === undefined){
+		return undefined;
+	}
+	return Array.isArray(comment) ? comment : [comment] as readonly RoxygenTag[];
+}
+
+/**
+ * Please note that this actually modifies the parameters to only leave those that differ!
+ */
+function calculateArgumentDiff(inheritedParams: string[], remainingFunctionParam: string[], remainingRoxygenParam: string[]): boolean{
```
EagleoutIce: Generally, I'd prefer to return new lists of over and underdocumented names.
+ we could use `Set` with difference and `intersect` to calculate these differences.
Do you want to try out and change that?

src/linter/rules/roxygen-arguments.ts
```diff
@@ -39,50 +39,28 @@ function getDocumentation(id: NodeId, idMap: AstIdMap): readonly RoxygenTag[] |
...
- * Please note that this actually modifies the parameters to only leave those that differ!
- */
-function calculateArgumentDiff(inheritedParams: string[], remainingFunctionParam: string[], remainingRoxygenParam: string[]): boolean{
-	const notOverdocumented = remainingFunctionParam.includes('...');
-	remainingFunctionParam.sort();
-	remainingRoxygenParam.sort();
-	inheritedParams.sort();
-	let i = 0;
-	let j = 0;
+function calculateArgumentDiff(inheritedParams: string[], functionParam: string[], roxygenParam: string[]): [boolean, string[], string[]]{
```
EagleoutIce: You can rewrite the return type also to something like this:

```ts
false | [under: string[], over: string[]]
```

Or even:

```ts
false | { under: string[], over: string[] }
```

src/linter/rules/roxygen-arguments.ts
```diff
@@ -39,50 +39,28 @@ function getDocumentation(id: NodeId, idMap: AstIdMap): readonly RoxygenTag[] |
...
-	//case: '...'
-	if(notOverdocumented){
+	let underDocumented = new Set(functionParam);
+	let overDocumented = new Set(roxygenParam);
+	const commonParams = underDocumented.intersection(overDocumented);
+	underDocumented = underDocumented.difference(commonParams);
+	overDocumented = overDocumented.difference(commonParams);
+
+	//case: '...', overdocumentation not possible
+	if(functionParam.includes('...')){
```
EagleoutIce: Given that you have a set you can also make this check after lien 45 with `underDocumented.has('...')` which would be even sipler

src/linter/rules/roxygen-arguments.ts
EagleoutIce: given how you use the function signature now you can now also request `ReadonlySet` as inputs or at least mark the arrays as `readonly` to signal that they no longer modify their inputs

PR #2416, opened 2026-03-31
[AbsInt] Fix several issues with absint framework and data frame shape inference

OliverGerstl:
- Correct over-approximation for data frames resulting from too large CSV files
- Improve visitation order of absint visitor in case of `break` and `next`
- Handle unsupported environment-modifying functions
- Better support unknown cases of functions returning data frames
- Fix representation of bottom of the state abstract domain


src/abstract-interpretation/data-frame/mappers/access-mapper.ts
```diff
@@ -172,17 +164,13 @@ function mapDataFrameIndexColRowAccess(
...
 function getAccessArgs(
 	operator: RIndexAccess['operator'],
 	args: readonly RFunctionArgument<ParentInformation>[]
 ): readonly RFunctionArgument<ParentInformation>[] {
-	const specialArgs = SpecialAccessArgumentsMapper[operator];
-
-	return args.filter(arg => arg === EmptyArgument || arg.name === undefined || !specialArgs.includes(unquoteArgument(
-		arg.name.content
-	)));
+	return args.filter(arg => arg === EmptyArgument || arg.name === undefined);
```
EagleoutIce: Hier könnte man auch prüfen ob das Function Argument unnamed ist - mit dem helper object

src/abstract-interpretation/data-frame/mappers/arguments.ts
```diff
@@ -177,9 +177,10 @@ export function getUnresolvedSymbolsInExpression(
 	RNode.visitAst(expression, node => {
 		if(node.type === RType.Symbol) {
 			const vertex = dfg.get(node.info.id);
+			const symbolName = Identifier.mapName(node.content, unquoteArgument);
 
-			if(isUseVertex(vertex?.[0]) && vertex[1].size === 0) {
-				unresolvedSymbols.push(Identifier.mapName(node.content, unquoteArgument));
+			if(isUseVertex(vertex?.[0]) && vertex[1].size === 0 && symbolName !== '.') {
```
EagleoutIce, line 182: irgendwo dokumentieren warum das special is?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -31,44 +31,49 @@ enum DataFrameType {
...
-	'read.csv':      { mapper: mapDataFrameRead,      library: 'utils', returnType: DataFrameType.DataFrame },
-	'read.csv2':     { mapper: mapDataFrameRead,      library: 'utils', returnType: DataFrameType.DataFrame },
-	'read.delim':    { mapper: mapDataFrameRead,      library: 'utils', returnType: DataFrameType.DataFrame },
-	'read.delim2':   { mapper: mapDataFrameRead,      library: 'utils', returnType: DataFrameType.DataFrame },
-	'read_table':    { mapper: mapDataFrameRead,      library: 'readr', returnType: DataFrameType.Tibble    },
-	'read_csv':      { mapper: mapDataFrameRead,      library: 'readr', returnType: DataFrameType.Tibble    },
-	'read_csv2':     { mapper: mapDataFrameRead,      library: 'readr', returnType: DataFrameType.Tibble    },
-	'read_tsv':      { mapper: mapDataFrameRead,      library: 'readr', returnType: DataFrameType.Tibble    },
-	'read_delim':    { mapper: mapDataFrameRead,      library: 'readr', returnType: DataFrameType.Tibble    },
+	'data.frame':    { mapper: mapDataFrameCreate,    library: 'base',  returnType: DataFrameType.DataFrame, alwaysDataFrame: true },
```
EagleoutIce, line 35: i think it would be beneficial to document the reason for this flag especially wrt. returning tibbles etc.

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -31,44 +31,49 @@ enum DataFrameType {
...
+	'arrange':       { mapper: mapDataFrameIdentity,  library: 'dplyr', returnType: DataFrameType.DataFrame, alwaysDataFrame: true }
 } as const satisfies Record<string, DataFrameFunctionMapperInfo<never>>;
 
 /**
  * List of other data frame functions that are not explicitly supported but may return data frames.
  */
 const OtherDataFrameFunctions = [
 	{
+		type:       'entry_point',
+		names:      ['as.data.frame.matrix'],
```
EagleoutIce: wobei es hier ja schon in s3 dispatches geht

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -121,7 +121,7 @@ const OtherDataFrameFunctions = [
 		dataFrame:  { pos: 0, name: 'object' }
 	}, {
 		type:       'transformation',
-		names:      ['unique', 't'],
+		names:      ['unique', 'droplevels'],
```
EagleoutIce, line 125: have we lost `t`?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
```diff
@@ -1386,23 +1409,24 @@ function getRequestFromRead(
...
 /**
  * Gets all entries from a line of a CSV file using a custom separator char, quote char, and comment char
  */
 function getEntriesFromCsvLine(line: string, sep: string = ',', quote: string = '"', comment: string = '', trim: boolean = true): (string | undefined)[] {
+	sep = sep.length > 0 ? sep : '\\s';  // default to whitespace separator
 	sep = escapeRegExp(sep, true);  // only allow tokens like `\s`, `\t`, or `\n` in separator, quote, and comment chars
 	quote = escapeRegExp(quote);
 	comment = escapeRegExp(comment);
-	const quantifier = sep === '\\s' ? '+' : '*';  // do not allow unquoted empty entries in whitespace-sparated files
+	const quantifier = sep === '\\s' ? '+' : '*';  // do not allow unquoted empty entries in whitespace-separated files
```
EagleoutIce, line 1440: adding this fixed feils weird if `sep` is pre-set is this from `r` itself?

src/abstract-interpretation/domains/product-domain.ts
```diff
@@ -68,7 +74,7 @@ export abstract class ProductDomain<Product extends AbstractProduct>
 		for(const key in result.value) {
 			result._value[key] = result.value[key].join(other.value[key]);
 		}
-		return result;
+		return result.refine();
```
EagleoutIce, line 78: refine on bottom?

src/abstract-interpretation/domains/product-domain.ts
```diff
@@ -143,15 +149,37 @@ export abstract class ProductDomain<Product extends AbstractProduct>
...
 
+	public isBottom(): boolean;
+	public isBottom(): this is this;
 	public isBottom(): this is this {
 		return Object.values(this.value).every(value => value.isBottom());
 	}
 
+	public isValue(): boolean;
+	public isValue(): this is this;
 	public isValue(): this is this {
```
EagleoutIce, line 167: hmmm?

src/abstract-interpretation/domains/state-abstract-domain.ts
```diff
@@ -1,158 +1,299 @@
...
+import { AbstractDomain, domainElementToString, type AnyAbstractDomain, type ConcreteDomain } from './abstract-domain';
+import { Bottom, BottomSymbol, Top } from './lattice';
+
+/** The type of the concrete state of the concrete domain of a state abstract domain that maps keys to a concrete value in the concrete domain */
+export type ConcreteState<Domain extends AnyAbstractDomain> = ReadonlyMap<NodeId, ConcreteDomain<Domain>>;
+
+/** The type of the actual values of the state abstract domain as map of keys to domain values */
+type StateDomainValue<Domain extends AnyAbstractDomain> = ReadonlyMap<NodeId, Domain>;
+/** The type of the Top element of the state abstract domain as (empty) map of keys to domain values */
+type StateDomainTop<Domain extends AnyAbstractDomain> = ReadonlyMap<NodeId, Domain>;
```
EagleoutIce: any reason why Value and Top have the same type alias?

src/abstract-interpretation/domains/state-abstract-domain.ts
```diff
@@ -1,158 +1,299 @@
...
+		this.domain = domain;
+	}
+
+	public create(value: StateDomainLift<Domain>): this;
+	public create(value: StateDomainLift<Domain>): StateAbstractDomain<Domain> {
+		return new StateAbstractDomain(value, this.domain);
+	}
+
+	public static top<Domain extends AnyAbstractDomain>(domain: Domain): StateAbstractDomain<Domain, StateDomainTop<Domain>> {
+		return new StateAbstractDomain(new Map(), domain) as StateAbstractDomain<Domain, StateDomainTop<Domain>>;
```
EagleoutIce: given that this is readonly, do we need a new map instantiation for every top?

src/abstract-interpretation/domains/state-abstract-domain.ts
```diff
@@ -1,158 +1,299 @@
...
 	public toJson(): unknown {
-		if(this._isBottom) {
-			return BottomSymbol;
-		} else if(this.value.size === 0) {
-			return TopSymbol;
+		if(this.value === Bottom) {
+			return this.value.description;
 		}
-		return super.toJson();
+		return Object.fromEntries(this.value.entries().map(([key, value]) => [key, value.toJson()]));
```
EagleoutIce, line 257: i think we could also do this with the default toJSON property check, but this is fine as well

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -143,7 +149,7 @@ export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDo
 		const exitNodes = exitPoints.map(CfgVertex.getRootId).filter(isNotUndefined);
 		const states = exitNodes.map(node => this.trace.get(node)).filter(isNotUndefined);
 
-		return AbstractDomain.joinAll(states, this._currentState.top());
+		return AbstractDomain.joinAll(states, this._currentState.bottom());
```
EagleoutIce, line 152: it still feels wird to access this iwth underscore :P

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -211,12 +219,12 @@ export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDo
...
 			// continue visiting after widening point if visited for the first time or the state changed
-			return visitedCount === 0 || !oldState.equals(this._currentState);
+			return visitedCount === 0 || !oldState?.equals(this._currentState);
 		} else {
 			this.onVisitNode(vertexId);
 
-			// discard the inferred abstract state when encountering functions with unknown side effects (e.g. `eval`)
-			if(this.config.dfg.unknownSideEffects.has(nodeId)) {
+			// discard the inferred abstract state when encountering unsupported (environment-changing) functions (e.g. `eval`, `load`, `attach`, `rm`, ...)
+			if(isUnsupportedFunctionCall(this.getDataflowGraph(nodeId))) {
```
EagleoutIce: why is unsupported function call not part of a helper object here?

src/abstract-interpretation/unsupported-functions.ts
```diff
@@ -0,0 +1,35 @@
+import { Identifier } from '../dataflow/environments/identifier';
+import { type DataflowGraphVertexArgument, isFunctionCallVertex } from '../dataflow/graph/vertex';
+
+export const UnsupportedFunctions = {
```
EagleoutIce: this is partially domain specific isn't it?

src/abstract-interpretation/domains/product-domain.ts
```diff
@@ -18,6 +19,11 @@ export type ConcreteProduct<Product extends AbstractProduct> = {
 export abstract class ProductDomain<Product extends AbstractProduct>
 	extends AbstractDomain<ConcreteProduct<Product>, Product, Product, Product> {
 
+	constructor(value: Product) {
+		super(Object.fromEntries(Object.entries(value).map(([key, value]) => [key, value.create(value.value)])) as Product);
```
EagleoutIce: you/we quite frequently use this object map-on values pattern which is, in a way , I functor :D (btw. what about `value.clone()`?),
so why not have an `fmap` on the object?

`mapProperties(value, (k, v) => [k, v.clone()])`

or something similar, if we just want a conventional fmap (nothing co-crazy)
`mapProperties(value, v => v.clone())`

with the added benefit for type safety

src/abstract-interpretation/domains/product-domain.ts
```diff
@@ -68,7 +74,7 @@ export abstract class ProductDomain<Product extends AbstractProduct>
 		for(const key in result.value) {
 			result._value[key] = result.value[key].join(other.value[key]);
```
EagleoutIce, line 76: the way of setting on underscore but retrieving without feels weird

src/abstract-interpretation/absint-visitor.ts
```diff
@@ -30,27 +31,32 @@ export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDo
 	/**
 	 * The abstract trace of the abstract interpretation visitor mapping node IDs to the abstract state at the respective node.
 	 */
-	protected readonly trace: Map<NodeId, MutableStateAbstractDomain<Domain>> = new Map();
+	private readonly trace: Map<NodeId, MutableStateAbstractDomain<Domain>> = new Map();
```
EagleoutIce, line 34: why hide the traces this way?

package.json
```diff
@@ -210,7 +209,7 @@
     "gray-matter": "^4.0.3",
     "joi": "^18.0.1",
     "lz-string": "^1.5.0",
-    "n-readlines": "^1.0.3",
+    "n-readlines": "^3.4.1",
```
EagleoutIce, line 212: big bump, are types included now?

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: The result of `t` is coerced to a matrix which is why I had to remove it

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Yes, if you set the separator for `read.csv` (explicitly) to an empty string "", whitespace separation is used (which is also the default). This resulted in a few wrongly parsed CSV files in the eval :)

src/abstract-interpretation/domains/product-domain.ts
OliverGerstl: This is to fix the type constraint, so that I can do something like `if (domain isTop()) ... else if (domain.isValue()) ...` without TypeScript inferring `never` for the second occurence of `domain` after `isValue`

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: The problem is that `_currentState` is a `MutableStateAbstractDomain` and the getter `currentState` is a non-mutable `StateAbtractDomain`, and here I need the mutable version

src/abstract-interpretation/unsupported-functions.ts
OliverGerstl: Yes, it is domain-specific to abstract interpretation / value inference, if that is what you mean

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: The elements in the trace are mutable which results in hard to find errors when they are mutated without copying/cloning first. Therefore, the trace should only be accessed/modified via `updateState`, `removeState`, and `getAbstractState`.

package.json
OliverGerstl: Yes, types are included. This was mainly necessary, as the old version did not handle all types of newlines correctly

src/abstract-interpretation/domains/state-abstract-domain.ts
OliverGerstl: The state domain is readonly from the outside, but it can be modified by subclasses via `set` and `remove` or as mutable state abstract domain

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
OliverGerstl: Oh yes, I didn't realize that. But I will leave it there for now, as I saw it a few times in the scripts

src/abstract-interpretation/domains/product-domain.ts
OliverGerstl: I added a check to only refine if the value if neither top nor bottom

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: fine, but we should make a marker about this. while they can be called directly, this can be also called by `as.data.frame`

src/abstract-interpretation/domains/state-abstract-domain.ts
OliverGerstl: No, I changed `StateDomainTop` to `ReadonlyMap<NodeId, never>` :)

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: ok

src/abstract-interpretation/data-frame/mappers/function-mapper.ts
EagleoutIce: oh what an amazing feature ._.

src/abstract-interpretation/domains/product-domain.ts
EagleoutIce: (re)fine

src/abstract-interpretation/domains/state-abstract-domain.ts
EagleoutIce: okay, but I meant as a seed for top :sweat_smile: but that is still fine

src/abstract-interpretation/absint-visitor.ts
EagleoutIce: I have a personal vendetta against such patterns xD but fine

src/abstract-interpretation/unsupported-functions.ts
EagleoutIce: Yeah but not just this, for a domain like the string domain we may want to have our own handling for `eval` and decide whether it is unsupported -> I would make this object the default and move it into the domain to avoid having a global function that just 'decides' this

package.json
EagleoutIce: great (:sob:)

src/abstract-interpretation/domains/product-domain.ts
OliverGerstl: I think this is more something general which is not specific to the product domain

EagleoutIce: yes it is, but i just noticed it, and I think the pattern would be a useful readability addition.

src/abstract-interpretation/unsupported-functions.ts
OliverGerstl: Ok, I moved the function into a overwritable method of the absint visitor

src/abstract-interpretation/domains/product-domain.ts
OliverGerstl: Unfortunately, as `value` is a (public) getter, I get an error when using `value` instead of `_value` in the property assignments

PR #2455, opened 2026-04-23
[AbsInt] Generalize abstract interpretation visitor to other state-like domains

### LINKED ISSUE #2454: [AbsInt] Generalize abstract interpretation visitor to other state-like domains
Generalize the abstract interpretation visitor to allow not only a `StateAbstractDomain` but also other relational abstract domains (e.g., a pentagon domain)

OliverGerstl:


src/abstract-interpretation/data-frame/shape-inference.ts
```diff
@@ -123,13 +124,13 @@ export class DataFrameShapeInferenceVisitor extends AbstractInterpretationVisito
 			const constraintType = type ?? getConstraintType(operation);
 
 			if(operand !== undefined && constraintType === ConstraintType.OperandModification) {
-				this.updateState(operand, value);
+				this.currentState.set(operand, value);
```
EagleoutIce, line 127: why is this no longer done via a method call?

src/abstract-interpretation/domains/abstract-domain.ts
```diff
@@ -113,13 +113,41 @@ implements Lattice<Abstract, Top, Bot, Value> {
...
+	public static toString(this: void, value: AnyAbstractDomain | unknown): string {
+		if(value instanceof Map) {
+			return `{${value.entries().map(([key, value]) => `${AbstractDomain.toString(key)} -> ${AbstractDomain.toString(value)}`).toArray().join(', ')}}`;
+		} else if(value instanceof Set) {
+			return `{${value.values().map(AbstractDomain.toString).toArray().join(', ')}}`;
+		} else if(typeof value === 'object' && value !== null && value.toString !== Object.prototype.toString) {
+			// eslint-disable-next-line @typescript-eslint/no-base-to-string
+			return value.toString();
+		} else if(Array.isArray(value)) {
+			return `[${value.map(AbstractDomain.toString).join(', ')}]`;
```
EagleoutIce, line 129: why does this *not* dispatch to the respective toString of the element (maybe I am too tired still xD)? shouldn't it call `.map(v => v.toString())` to allow for custom implementations to kick in?
Is this tested?

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
+import type { Writable } from 'ts-essentials';
+import { type  AbstractValue, type AnyAbstractDomain, AbstractDomain } from './abstract-domain';
+import { Top } from './lattice';
+import { Record } from '../../util/record';
+
+/** The type of an abstract product of a product domain mapping named properties of the product to abstract domains */
+export type AbstractProduct = { [key in string]?: AnyAbstractDomain };
```
EagleoutIce: Any reason you do not provide a generic to limit products to, e.g., numerical only domains?

EagleoutIce: + These are not documented yet are they?

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/** The type of the concrete product of an abstract product mapping each property to a concrete value in the respective concrete domain */
+export type ConcreteProduct<Product extends AbstractProduct> = {
```
EagleoutIce: Maybe call such unlifter something like `To...`? or `ConcreteProductOf` because otherwise it feels weird to have `AbstractProduct` to build a new one and then have `ConcreteProduct<Abstract...>` to map

src/abstract-interpretation/domains/multi-value-state-domain.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	public readonly reductions: readonly Reduction<Product>[];
+
+	constructor(value: Product, domain: Required<Product>, reductions: readonly Reduction<Product>[] = []) {
+		super(value, domain);
+		this.reductions = reductions;
+	}
+
+	public create(value: Product): this;
+	public create(value: Product): MultiValueDomain<Product> {
```
EagleoutIce, line 65: is this were you meant `this` is required? i see nothing hindering us from using `this` directly as we do not even narrow

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+	[Key in keyof Product]: Product[Key] extends AbstractDomain<infer Concrete, unknown, unknown, unknown> ? Concrete : never;
+};
+
+/**
+ * The type of the abstract product values of a product abstract domain.
+ * @template Product - The abstract product to get the product value type for
+ */
+export type ProductValue<Product extends AbstractProduct> = {
```
EagleoutIce: Generally, I'd argue that all of these lifts could be generalized into a

`LiftRecordByKey<AbstractValue[, never]>` - for all type helpers with the product domain

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+	public meet(other: this): this {
+		const result = this.create(this.value);
+
+		for(const key in result.value) {
+			if(other.value[key] === undefined) {
+				result._value[key] = undefined as Product[Extract<keyof Product, string>];
+			}
+		}
+		for(const key in other.value) {
+			if(result.value[key] !== undefined && other.value[key] !== undefined) {
```
EagleoutIce: so this logic feels inconsistent because you iterate twice:

* if you set result._value to undefined if other is undefined before, you do not need to check again here 
* otherwise: why not move it into a ternary and safe the other iteration:

```ts
if(result.value[key] !== undefined) {
    				result._value[key] = other.value[key] === undefined ? ...
}
```

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+				result._value[key] = result.value[key].widen(other.value[key]) as Product[Extract<keyof Product, string>];
+			}
+		}
+		return result;
+	}
+
+	public narrow(other: this): this {
+		const result = this.create(this.value);
+
+		for(const key in result.value) {
```
EagleoutIce: same as with the key logic

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+			if(other.value[key] === undefined) {
+				result._value[key] = undefined as Product[Extract<keyof Product, string>];
```
EagleoutIce: shouldnt `as typeof result._...` be more precise?

src/abstract-interpretation/domains/partial-product-domain.ts
```diff
@@ -0,0 +1,229 @@
...
+				result._value[key] = other.value[key];
+			} else {
+				result._value[key] = result.value[key].widen(other.value[key]) as Product[Extract<keyof Product, string>];
+			}
+		}
+		return result;
+	}
+
+	public narrow(other: this): this {
+		const result = this.create(this.value);
```
EagleoutIce: unsure whether the create dispatch might be too expensive?

EagleoutIce (approved):

src/abstract-interpretation/data-frame/shape-inference.ts
OliverGerstl: As the new state-like domain can be any domain, it does not have to be a state domain of a value domain.
Hence, other state-like domains (like a Pentagon domain) will require additional update methods to update, for example, relational parts of the state-like domain.

EagleoutIce: fine :D

src/abstract-interpretation/domains/abstract-domain.ts
OliverGerstl: If the array is, for example, an array of objects (e.g., sets or maps), I don't want to get "[[object Set], [object Set]". The part
```typescript
if(typeof value === 'object' && value !== null && value.toString !== Object.prototype.toString) {
	// eslint-disable-next-line @typescript-eslint/no-base-to-string
	return value.toString();
}
```
allows for custom implementations of `toString`.

src/abstract-interpretation/domains/partial-product-domain.ts
OliverGerstl: This was not necessary until now, but could be easily implemented via something like `type NumericProduct = Record<string, NumericAbstractDomain>`.
The documentation for the domains is automatically generated from the doc comments.

src/abstract-interpretation/domains/abstract-domain.ts
EagleoutIce: Ok, fair hack although this feels... weird

src/abstract-interpretation/domains/multi-value-state-domain.ts
OliverGerstl: When I omit `this` here, I get an inheritance error, and when I omit the `MultiValueDomain<Product>`, I get a type error

src/abstract-interpretation/domains/partial-product-domain.ts
OliverGerstl: Sorry, but I don't understand this: `AbstractValue[, never]` produces a type error as `AbstractValue` requires a type parameter, and I am not sure what the semantic of `[, never]` is here. Maybe we should discuss this

src/abstract-interpretation/domains/partial-product-domain.ts
EagleoutIce: Yes I've seen the automated generation (although for the first point I'd argue that a generic would make it more explicit as it hides the fact that it is an Record under the hood and could allow us to change this later)

wrt. products I'd argue that these meta/compositional domains should have at least a small block in the documentation in which they are highlighted as they serve a combinatory purpose

src/abstract-interpretation/domains/multi-value-state-domain.ts
EagleoutIce: Jeah i think we talked about this once already, if I find the time I'll have a look at the base definitions, this shouldnt be needed :sweat:

src/abstract-interpretation/domains/partial-product-domain.ts
EagleoutIce: Jeah let's discuss this, the idea was for the RHS

src/abstract-interpretation/domains/partial-product-domain.ts
OliverGerstl: Yes, I completely restructured the join, meet, widening, and narrowing in the (partial) product domain

src/abstract-interpretation/domains/partial-product-domain.ts
OliverGerstl: Mmh, I don't really like type helpers named with `To...` or `...Of`, I prefer the style without `To`/`Of` which is also used in `linter-rules.ts`

src/abstract-interpretation/domains/partial-product-domain.ts
EagleoutIce:
```ts
export type AbstractProduct<D extends AnyAbstractDomain = AnyAbstractDomain> = { [key in string]?: D };

type NumericProduct = AbstractProduct<NumericDomain>;
```

EagleoutIce:
_This pull request is included in v2.10.5 (see [Release v2.10.5 (Extend Noteboook Support, Default Plugins, AbsInt fixes)](https://github.com/flowr-analysis/flowr/releases/tag/v2.10.5))._

PR #2461, opened 2026-04-24
[AbsInt] Fix abstract interpretation visiting order at replacement calls

### LINKED ISSUE #2459: [AbsInt] CFG visitation at replacement calls is in wrong order

OliverGerstl:


src/abstract-interpretation/absint-visitor.ts
```diff
@@ -240,7 +244,36 @@ export abstract class AbstractInterpretationVisitor<Domain extends AnyAbstractDo
...
+				}
+			}
+		}
+	}
+
 	protected override onDispatchFunctionCallOrigin(call: DataflowGraphVertexFunctionCall, origin: BuiltInProcName) {
+		if(origin === BuiltInProcName.Replacement) {
+			const node = this.getNormalizedAst(call.id);
+			const assignment = RNode.iterateParents(node, this.config.normalizedAst.idMap)
+				.find(parent => this.getDataflowGraph(parent.info.id) === undefined);
```
EagleoutIce, line 234: Why do we check for a node no longer present in the DFG? why not just guard on the role? :D

EagleoutIce (approved):

src/abstract-interpretation/absint-visitor.ts
OliverGerstl: The problem is that I need to identify the assignment node belonging to a replacement call to defer the processing of the replacement call until this assignment.
The only way I found to identify this assignment is by finding the first parent node of the replacement call that has no corresponding data flow vertex (as the assignment node of replacement calls is not mapped to a DFG vertex).
The role of this assignment is just `ExpressionListChild` which does not always have to be the case (e.g., if the replacement calls is wrapped in another expression)

EagleoutIce: if it works its fine for now but there should be a more robust way checking for role + origin but that is fine here

OliverGerstl: Yeah, the problem is that the assignment has no origin (as it is not represented in the DFG) and the role doesn't tell us anything about the fact that the assignment belongs to a replacement call (as it does not incorporate DFG information) :/

EagleoutIce:
_This pull request is included in v2.10.5 (see [Release v2.10.5 (Extend Noteboook Support, Default Plugins, AbsInt fixes)](https://github.com/flowr-analysis/flowr/releases/tag/v2.10.5))._

PR #2465, opened 2026-04-28
Pentagon as value product

schuler-henry:


src/abstract-interpretation/domains/state-abstract-domain.ts
```diff
@@ -40,45 +41,42 @@ export class StateAbstractDomain<Domain extends AnyAbstractDomain, Value extends
 		return new StateAbstractDomain(value, this.domain);
 	}
 
-	public static top<Domain extends AnyAbstractDomain>(domain: Domain): StateAbstractDomain<Domain, StateDomainTop> {
-		return new StateAbstractDomain(new Map<NodeId, never>(), domain);
+	public static top<Domain extends AnyAbstractDomain, T extends StateAbstractDomain<Domain, StateDomainLift<Domain>>>(this: new (value: StateDomainTop, domain: Domain) => T, domain: Domain): T {
```
OliverGerstl: Maybe a more readable name for `T`, e.g. `StateDomain`?
```suggestion
	public static top<Domain extends AnyAbstractDomain, StateDomain extends StateAbstractDomain<Domain>>(this: new (value: StateDomainTop, domain: Domain) => StateDomain, domain: Domain): StateDomain {
```

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+	public override bottom(): this & StateAbstractDomain<ClosedPentagonValueDomain, StateDomainBottom>;
+	public override bottom(): ClosedPentagonDomain {
+		return ClosedPentagonDomain.bottom(this.domain);
 	}
 
-	public static top(): ClosedPentagonDomain {
-		return new ClosedPentagonDomain({ interval: StateAbstractDomain.top(IntervalDomain.top()), upperBounds: UpperBoundsDomain.top() });
+	public override create(value: StateDomainLift<ClosedPentagonValueDomain>): this;
+	public override create(value: StateDomainLift<ClosedPentagonValueDomain>): ClosedPentagonDomain {
+		return new ClosedPentagonDomain(value, ClosedPentagonValueDomain.top());
```
OliverGerstl:
```suggestion
		return new ClosedPentagonDomain(value, this.domain);
```

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
 	}
 
-	public static bottom(): ClosedPentagonDomain {
-		return new ClosedPentagonDomain({ interval: StateAbstractDomain.bottom(IntervalDomain.bottom()), upperBounds: UpperBoundsDomain.bottom() });
+	public override set(node: NodeId, value: ClosedPentagonValueDomain) {
+		if(this._value !== Bottom) {
+			const valueWithoutNodeItself = value.create(value.value);
+			valueWithoutNodeItself.value.upperBounds.remove(node);
+			(this._value as Map<NodeId, ClosedPentagonValueDomain>).set(node, valueWithoutNodeItself);
```
OliverGerstl: Maybe `super.set(node, valueWithoutNodeItself)`?

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
-		return new ClosedPentagonDomain({ interval: StateAbstractDomain.top(IntervalDomain.top()), upperBounds: UpperBoundsDomain.top() });
+	public override create(value: StateDomainLift<ClosedPentagonValueDomain>): this;
+	public override create(value: StateDomainLift<ClosedPentagonValueDomain>): ClosedPentagonDomain {
+		return new ClosedPentagonDomain(value, ClosedPentagonValueDomain.top());
 	}
 
-	public static bottom(): ClosedPentagonDomain {
-		return new ClosedPentagonDomain({ interval: StateAbstractDomain.bottom(IntervalDomain.bottom()), upperBounds: UpperBoundsDomain.bottom() });
+	public override set(node: NodeId, value: ClosedPentagonValueDomain) {
+		if(this._value !== Bottom) {
```
OliverGerstl: If possible, prefer to use the getter `this._value` if you don't need to modify it
```suggestion
		if(this.value !== Bottom) {
```

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+			const valueWithoutNodeItself = value.create(value.value);
+			valueWithoutNodeItself.value.upperBounds.remove(node);
```
OliverGerstl: Can't you just put this directly into the reduction `ClosedPentagonDomain.reduce` so that you do not have to do this separately. If not, I would at least outsource this into a helper function

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
-	protected override reduce(value: AbstractClosedPentagon): AbstractClosedPentagon {
-		if(value.interval.isTop() && value.upperBounds.isTop()) {
-			return value;
+	public override get(node: NodeId): ClosedPentagonValueDomain | undefined {
+		if(this._value === Bottom) {
+			return this.domain.bottom();
 		}
-		if(value.interval.isBottom() || value.upperBounds.isBottom()) {
-			return { interval: StateAbstractDomain.bottom(IntervalDomain.bottom()), upperBounds: UpperBoundsDomain.bottom() };
+		const value = this._value.get(node);
```
OliverGerstl: Is there a reason why you don't use just `const value = super.get(node)`?

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+		return this.create(super.meet(other).value);
+	}
+
+	public override widen(other: this): this {
+		return this.create(super.widen(other).value);
+	}
+
+	public override narrow(_other: this): this {
+		throw new Error('Not Implemented');
+	}
```
OliverGerstl: I would at least call `super.narrow` so that we do not have to adapt this code when narrowing is supported later

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+		return this.create(super.join(other).value);
+	}
+
+	public override meet(other: this): this {
+		return this.create(super.meet(other).value);
+	}
+
+	public override widen(other: this): this {
+		return this.create(super.widen(other).value);
+	}
```
OliverGerstl, line 64: This is currently a bit inefficient, as we call `create` two times every time we are joining two states which copies the whole mapping

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+	public override narrow(_other: this): this {
+		throw new Error('Not Implemented');
+	}
+
+	public override concretize(_limit: number): ReadonlySet<ConcreteState<ClosedPentagonValueDomain>> | typeof Top {
+		if(this.value === Bottom) {
+			return new Set();
+		}
+		return Top;
+	}
```
OliverGerstl: Do you really need to override this?

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+			}
+			const pentagon = new ClosedPentagonValueDomain({ interval: interval, upperBounds: upperBounds });
+			result.set(nodeIdA, pentagon);
+		}
+		return result;
+	}
 
-				(value.upperBounds.value as Map<NodeId, ReadonlySet<NodeId>>).set(key, newUpperBounds);
+	public override isTop(): this is this & StateAbstractDomain<ClosedPentagonValueDomain, StateDomainTop> {
+		return this.value !== Bottom && this.value.values().every(pentagon => pentagon.value.interval.isTop() && pentagon.value.upperBounds.isTop());
```
OliverGerstl: I am not sure if this is the correct Top representation here.
Because, if you have a node for which you inferred Top in the inferval/upper bounds domain, you at least inferred that it is a number.
This contains more information than an empty map where you have no information about any node.
I think you don't need to override the `isTop` check here (your implementation of `this.top` is also the same as for the state domain)

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,164 @@
...
+	}
+
+	public static reduce(value: StateDomainLift<ClosedPentagonValueDomain>): StateDomainLift<ClosedPentagonValueDomain> {
```
OliverGerstl: Does this need to be `public`?

src/abstract-interpretation/pentagon/closed-pentagon-value-domain.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export class ClosedPentagonValueDomain extends ProductDomain<AbstractClosedPentagon> {
+	public create(value: AbstractClosedPentagon): this;
+	public create(value: AbstractClosedPentagon): ClosedPentagonValueDomain {
+		return new ClosedPentagonValueDomain(value);
+	}
+
+	public static top(): ClosedPentagonValueDomain {
+		return new ClosedPentagonValueDomain({ interval: IntervalDomain.top(), upperBounds: UpperBoundsValueDomain.top() as UpperBoundsValueDomain });
```
EagleoutIce: significant why?

EagleoutIce: Bottom carries them but top does not yet carry the sign-figures

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -1,96 +1,111 @@
...
- * @param visitor - The numeric inference visitor performing the analysis used to resolve argument intervals.
+ * @param set - Setter function to update an interval in the provided state.
+ * @param getInterval - Retrieves the inferred abstract interval from the visitor for the provided node.
+ * @param getVariableOrigins - Retrieves the origins of a node from the visitor.
  * @param dfg - The dataflow graph containing the vertex and its arguments.
  * @returns The filtered state after applying the semantics.
  */
-export function applyIntervalConditionSemantics(argNodeId: NodeId | undefined, state: StateAbstractDomain<IntervalDomain> | undefined, visitor: NumericIntervalInferenceVisitor, dfg: DataflowGraph): StateAbstractDomain<IntervalDomain> | undefined {
+export function applyIntervalConditionSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(argNodeId: NodeId | undefined, state: StateDomain, set: SetIntervalState<StateDomain>, getInterval: GetInterval<StateDomain>, getVariableOrigins: GetVariableOrigins, dfg: DataflowGraph): StateDomain {
```
OliverGerstl: I am not sure, but shouldn't you be able to just provide a `AbstractInterpretationVisitor<StateDomain>` where the `StateDomain` is a state domain of intervals (i.e. interval state domain or pentagon domain), and use the getter, setter and origin retrieval of that visitor (but maybe I am wrong)?
```suggestion
export function applyIntervalConditionSemantics<Domain extends IntervalDomain, StateDomain extends AnyStateDomain<Domain>>(argNodeId: NodeId | undefined, state: StateDomain, inference: AbstractInterpretationVisitor<StateDomain>): StateDomain {
```

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -186,7 +189,7 @@ function intervalAddOp(left: IntervalDomain | undefined, right: IntervalDomain |
  * @param arg - The interval to negate (undefined meaning no information).
  * @returns The resulting interval after negation. If the interval is undefined, the result is also undefined.
  */
-function intervalNegativeOp(arg: IntervalDomain | undefined): IntervalDomain | undefined {
+export function intervalNegativeOp(arg: IntervalDomain | undefined): IntervalDomain | undefined {
```
OliverGerstl: Minor: Maybe `intervalNegateOp`?

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -237,15 +240,71 @@ function intervalMultiplyOp(left: IntervalDomain | undefined, right: IntervalDom
...
+		if(0 < c || d < 0) {
+			return intervalMultiplyOp(left, right.create([1/d, 1/c]));
+		} else {
+			return intervalMultiplyOp(left, right.create([Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]));
+		}
+	}
+	return undefined;
+}
+
+function intervalPowerOp(left: IntervalDomain | undefined, right: IntervalDomain | undefined): IntervalDomain | undefined {
```
OliverGerstl, line 255: Just a note: Make sure that you have test cases for all edge cases (negative base/exponent, non-integer base/exponent, ...) :)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,297 @@
...
+ * @param leftNodeId - The node id of the left argument of the binary operator.
+ * @param rightNodeId - The node id of the right argument of the binary operator.
+ * @param state - The state to retrieve the argument values and apply the semantics to.
+ * @param set - Setter function to update an upper-bound in the provided state.
+ * @param getUb - Retrieves the inferred abstract upper-bound from the visitor for the provided node.
+ * @param getVariableOrigins - Retrieves the origins of a node from the visitor.
+ * @param dfg - The dataflow graph containing the vertex and its arguments.
+ * @returns The filtered state after applying the semantics.
+ */
+type UbBinaryOperatorSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>> = (leftNodeId: NodeId, rightNodeId: NodeId, state: StateDomain, set: SetUbState<StateDomain>, getUb: GetUb<StateDomain>, getVariableOrigins: GetVariableOrigins, dfg: DataflowGraph) => StateDomain;
```
OliverGerstl: Mmh, I am still not so happy with all these duplicate (type) definitions for the interval, upper bounds, and pentagon domain. Maybe you can consider unifying them?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,297 @@
...
+ * Applies the abstract condition semantics of the provided function with respect to the upper-bounds domain to the provided args.
+ * @param argNodeId - The node id representing the condition to which the semantics should be applied.
+ * @param state - The state before applying the semantics, which can be used to determine the resulting state after applying the semantics.
+ * @param set - Setter function to update an upper-bound in the provided state.
+ * @param getUb - Retrieves the inferred abstract upper-bound from the visitor for the provided node.
+ * @param getVariableOrigins - Retrieves the origins of a node from the visitor.
+ * @param dfg - The dataflow graph containing the vertex and its arguments.
+ * @returns The filtered state after applying the semantics.
+ */
+export function applyUpperBoundsConditionSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(argNodeId: NodeId | undefined, state: StateDomain, set: SetUbState<StateDomain>, getUb: GetUb<StateDomain>, getVariableOrigins: GetVariableOrigins, dfg: DataflowGraph): StateDomain {
```
OliverGerstl: Maybe outsource the common part of `applyUpperBoundsConditionSemantics` and `applyNegatedUpperBoundsConditionSemantics` (I think they could also be generalized with `applyIntervalConditionSemantics` and `applyNegatedIntervalConditionSemantics`)?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/**
+ * Guard for unary operators, filtering all calls with more/less than 1 argument or with undefined argument.
+ * If the call has exactly 1 defined argument, the provided unary operator semantics is applied to it.
+ * Otherwise, the state is returned unmodified and a warning is logged.
+ * @param unaryOperatorSemantics - The semantics to apply if the call has exactly 1 defined argument.
+ * @returns The semantics to apply for a unary operator call, which includes the guard for the number of arguments.
+ */
+function unaryCondOpSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(unaryOperatorSemantics: UbUnaryOperatorSemantics<StateDomain>): UbNAryFnSemantics<StateDomain> {
```
OliverGerstl: These could be generic (with respect to the domain) as well :)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,297 @@
...
+function upperBoundsEqualsOp<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(leftNodeId: NodeId, rightNodeId: NodeId, state: StateDomain, set: SetUbState<StateDomain>, getUb: GetUb<StateDomain>, getVariableOrigins: GetVariableOrigins): StateDomain {
+	const leftValue = getUb(leftNodeId, state);
+	const rightValue = getUb(rightNodeId, state);
+
+	// We need to include the origin ids instead of the reference ids if the variable has an origin.
+	// We can only include the origin if there is only one origin, as we only then know that this is the only possible origin.
+	const originNodeIds = new Set<NodeId>();
+
+	const leftOrigins = getVariableOrigins(leftNodeId);
+	const leftOrigin = leftOrigins.length === 1 ? leftOrigins[0] : leftNodeId;
```
OliverGerstl: Minor: Maybe use a helper function for this origin handling for better readability, as you are using this a couple of times?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,297 @@
...
+
+	return leftState.meet(rightState);
+}
+
+function upperBoundsNegatedAndOp<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(leftNodeId: NodeId, rightNodeId: NodeId, state: StateDomain, set: SetUbState<StateDomain>, getUb: GetUb<StateDomain>, getVariableOrigins: GetVariableOrigins, dfg: DataflowGraph): StateDomain {
+	const leftState = applyNegatedUpperBoundsConditionSemantics(leftNodeId, state.create(state.value), set, getUb, getVariableOrigins, dfg);
+	const rightState = applyNegatedUpperBoundsConditionSemantics(rightNodeId, state.create(state.value), set, getUb, getVariableOrigins, dfg);
+
+	return leftState.join(rightState);
+}
```
OliverGerstl: I think these can also be independent of the used domain :)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-domain.ts
```diff
@@ -0,0 +1,100 @@
...
+import type {
+	ConcreteState,
+	StateDomainLift,
+	StateDomainTop,
+	StateDomainValue
+} from '../../domains/state-abstract-domain';
+import { StateAbstractDomain } from '../../domains/state-abstract-domain';
+import { UpperBoundsValueDomain } from './upper-bounds-value-domain';
+
+export class UpperBoundsDomain extends StateAbstractDomain<UpperBoundsValueDomain> {
```
OliverGerstl: This domain is currently only used for testing purposes? (Maybe add a comment or so?)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
```diff
@@ -0,0 +1,146 @@
...
+import { setEquals } from '../../../util/collections/set';
+
+const UpperBoundsTop: UpperBoundsValue = new Set<NodeId>();
+
+type UpperBoundsValue = ReadonlySet<NodeId>;
+type UpperBoundsTop = typeof UpperBoundsTop;
+type UpperBoundsBottom = typeof Bottom;
+export type UpperBoundsLift = UpperBoundsValue | UpperBoundsBottom;
+
+export class UpperBoundsValueDomain<Value extends UpperBoundsLift = UpperBoundsLift, Concrete = unknown> extends AbstractDomain<Concrete, UpperBoundsValue, UpperBoundsTop, UpperBoundsBottom, Value> {
```
OliverGerstl: Here, some basic documentation of what the purpose of this domain is would be nice, as these doc comments are also used to generate the wiki page for the abstract interpretation framework

OliverGerstl: I think you can just remove the `Concrete` type parameter here and set it directly to `unknown` (or `UpperBoundLift`). This is not really relevant for this domain and removing the type parameter simplifies the notation a bit :)

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,149 @@
...
 				}
+			}
+			const pentagon = new ClosedPentagonValueDomain({ interval: interval, upperBounds: upperBounds });
+			result.set(nodeIdA, pentagon);
+		}
+		return result;
+	}
+
+	private static reduce(value: StateDomainLift<ClosedPentagonValueDomain>): StateDomainLift<ClosedPentagonValueDomain> {
+		if(value === Bottom || value.values().some((value) => value.value.interval.isBottom() || value.value.upperBounds.isBottom())) {
```
OliverGerstl: For performance reasons, you could move the `isBottom` check for each value into the `for` loop, such that you do not have to iterate `value.values()`/`value.entries()` twice here

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -1,72 +1,149 @@
...
+			return Bottom;
+		}
 
-				const prevUpperBounds = value.upperBounds.value.get(key) ?? new Set();
-				const newUpperBounds = prevUpperBounds.union(inferredUpperBounds);
+		for(const [key, pentagon] of value.entries()) {
```
OliverGerstl, line 125: Here are more performance optimization opportunities:
- You could directly perform the first `isValue` check after the first `for` loop
- Instead of iterating over all pairs twice as `([k1, v1], [k2, v2])` and `([k2, v2], [k1, v1])`, you could use a for-i loop to iterate over all pairs only once and directly exclude the `([k, v], [k, v])` pairs

But I am not 100% sure if they really improve performance, and there are probably more options

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,278 @@
...
+		const rightOrigins = visitor.getVariableOrigins(rightNodeId);
+		const rightOrigin = rightOrigins.length === 1 ? rightOrigins[0]: rightNodeId;
+
+		if(rightOrigins.length <= 1) {
+			if(a >= 0) {
+				const rightPentagon = currentState.get(rightOrigin);
+				if(isNotUndefined(rightPentagon)) {
+					rightPentagon.value.upperBounds.add(target);
+					currentState.set(rightOrigin, rightPentagon);
+				}
```
OliverGerstl: Maybe this can also be outsourced in a helper function, as you use it multiple times :)

src/abstract-interpretation/pentagon/numeric-pentagon-inference.ts
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		if(Number.isNaN(node.content.num)) {
+			// NaN is part of the general Top, which is represented as undefined in the state abstract domain, so we can just skip it here.
+			return;
+		}
+
+		const pentagon = new ClosedPentagonValueDomain({
+			interval:    IntervalDomain.scalar(node.content.num, this.config.ctx.config.abstractInterpretation.numeric.significantFigures),
+			upperBounds: new UpperBoundsValueDomain(new Set())
```
OliverGerstl: Minor: Isn't this just `UpperBoundsValueDomain.top()`
```suggestion
			upperBounds: UpperBoundsValueDomain.top()
```

test/functionality/abstract-interpretation/pentagon/closed-pentagon-domain.test.ts
```diff
@@ -1,34 +1,41 @@
...
-		return {
-			interval:    new StateAbstractDomain(new Map(entries.map(([key, [interval, _]]) => [key, new IntervalDomain(interval)])), IntervalDomain.top()),
-			upperBounds: new UpperBoundsDomain(new Map(entries.map(([key, [_, upperBounds]]) => [key, new Set(upperBounds)])))
-		};
+	const pentagon = (entries: [NodeId, [interval: [l: number, r: number], upperBounds: NodeId[]]][]): StateDomainLift<ClosedPentagonValueDomain> => {
+		return new Map(entries.map(([key, [interval, upperBounds]]): [NodeId, ClosedPentagonValueDomain] => [key, new ClosedPentagonValueDomain({
+			interval:    new IntervalDomain(interval),
+			upperBounds: new UpperBoundsValueDomain(new Set(upperBounds))
+		})]));
 	};
```
OliverGerstl: Minor: You can also directly combine `createDomain` and `pentagon`, then you don't have to call `pentagon` for each test case element :)
```suggestion
	const createDomain = (entries: [NodeId, [interval: AbstractValue<IntervalDomain>, upperBounds: NodeId[]]][]) => {
		return new ClosedPentagonDomain(new Map(
			entries.map(([key, [interval, upperBounds]]) => [key, new ClosedPentagonValueDomain({
				interval:    new IntervalDomain(interval),
				upperBounds: new UpperBoundsValueDomain(new Set(upperBounds))
			})])
		), ClosedPentagonValueDomain.top());
	};
```

test/functionality/abstract-interpretation/pentagon/interim-presentation-example.test.ts
```diff
@@ -0,0 +1,39 @@
+import { describe } from 'vitest';
+import { testPentagonDomain, UpperBoundsTests } from './pentagon';
+import { IntervalTests } from '../interval/interval';
+
+describe('Interim Presentation Example', () => {
+	testPentagonDomain(`
+			ifelse(c, a <- -Inf, a <- Inf)
```
OliverGerstl: Minor: Maybe `a <- ifelse(c, -Inf, +Inf)` is more readable?

test/functionality/abstract-interpretation/pentagon/pentagon.ts
```diff
@@ -0,0 +1,188 @@
...
+			ctx:           ctx
+		});
+
+		visitor.start();
+	});
+
+	test.each(
+		// Append the test name manually because we need to access a property of criterionExpected, which cannot be done using vitest's test.each syntax.
+		(Object.entries(expected) as [criterion: SlicingCriterion, criterionExpected: PentagonSlicingCriterionExpected][])
+			.map(([criterion, criterionExpected]) => [`should infer ${criterionExpected.intervalMatching ?? DomainMatchingType.Exact}: ${criterionExpected.interval?.toString()} and include upper-bounds {${criterionExpected.upperBounds === Bottom ? BottomSymbol : criterionExpected.upperBounds.includes.values().toArray().join(', ')}}/exclude {${criterionExpected.upperBounds === Bottom ? BottomSymbol : criterionExpected.upperBounds.notIncludes.values().toArray().join(', ')}} for ${criterion} at ${code.trim().replaceAll('\n', ' \\n ')}`, criterion, criterionExpected] as const)
```
OliverGerstl: This line is hard to read :)
Maybe you can outsource the message construction to a helper?

test/functionality/abstract-interpretation/pentagon/pentagon.ts
```diff
@@ -0,0 +1,188 @@
...
+			controlFlow:   cfg,
+			ctx:           ctx
+		});
+
+		visitor.start();
+	});
+
+	test.each(
+		// Append the test name manually because we need to access a property of criterionExpected, which cannot be done using vitest's test.each syntax.
+		(Object.entries(expected) as [criterion: SlicingCriterion, criterionExpected: PentagonSlicingCriterionExpected][])
```
OliverGerstl: Minor: You can try `Record.entries` to omit the casting :)

test/functionality/abstract-interpretation/pentagon/pentagon.ts
```diff
@@ -0,0 +1,188 @@
...
+			const targetId: NodeId = SlicingCriterion.parse(criterion, ast.idMap);
+
+			const inferredPentagonDomain = visitor.getAbstractValue(targetId);
+
+			if(isUndefined(criterionExpected.intervalMatching)) {
+				criterionExpected.intervalMatching = DomainMatchingType.Exact;
+			}
+
+			const intervalErrorContext = `expected inferred value ${inferredPentagonDomain?.value.interval.toString()} to be ${criterionExpected.intervalMatching} 
+			match for ${criterionExpected.interval?.toString()} in final state ${visitor.getEndState().toString()} for ${code.trim().replaceAll('\n', ' \\n ')}`;
```
OliverGerstl: Technically, you are not checking for the `criterion` in `visitor.getEndState()` but in `visitor.getAbstractState(targetId)`, as `visitor.getAbstractValue()` returns the abstract value at the abstract state of the node and not at the end state

test/functionality/abstract-interpretation/pentagon/pentagon.ts
```diff
@@ -0,0 +1,188 @@
...
+
+			if(isNotUndefined(inferredPentagonDomain?.value.interval) && isNotUndefined(criterionExpected.interval)) {
+				if(criterionExpected.intervalMatching === DomainMatchingType.Exact) {
```
OliverGerstl: Minor: I would use `DomainMatchingType.Equal` instead of `DomainMatchingType.Exact` (for me, "exact" means that the lower bound is equal to the upper bound)

OliverGerstl (approved):

src/abstract-interpretation/interval/condition-semantics.ts
schuler-henry: The problem is, that the pentagon domain does not extend the interval domain, and the getter must return an interval (for the pentagon it would return a pentagon I assume).

We could however extract the getter and setter into an Interface, that needs to be implemented by the visitor and simply provide the visitor as an argumetn instead of the three arguments getInterval, set and getVariableOrigins.

And the interval and pentagon visitor then simply implement this interface that tells the condition semantics how to retrive an interval/set an interval.

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-domain.ts
schuler-henry: Or should we rather just remove the domain as we do (and at least I will probably) not need it rn?

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
schuler-henry: The first `isValue` check is only for the type system, as we already checked that it is not bottom before, so it has to be a value at this point. (It is required as the isBottom check does not provide the info to the type system that the bottom value is no longer possible if we return after a true isBottom check)

For the second point - do you simply mean two for i/j loops and accessing the values via their index?

src/abstract-interpretation/interval/condition-semantics.ts
OliverGerstl: Yes, I think this would be cleaner

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-domain.ts
OliverGerstl: Yes, I think if you don't need it anymore, you can just remove it.
It might be a bit confusing to have two different versions of the upper bounds domain, but only one of them is used :)

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
OliverGerstl: I thought of something like
```ts
for (let i = 0; i < value.size; i++) {
	const [key, pentagon] = value.entries()[i];

	for (let j = i + 1; j < value.size; j++) {
		const [otherKey, otherPentagon] = value.entries()[j];
		...
	}
}
```
But this does not work as `value.entries()` is just an iterator and creating an array would not be computationally cheap either.
So you can just ignore this, I am not sure if this would improve the performance anyway.

schuler-henry: Okay 😇

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: The test cases for this operator exist.

PR #2468, opened 2026-04-29
feat(dataflow): track ellmer chat backends and live helpers as unknown side effects

### LINKED ISSUE #2467: Track `ellmer` and co as unknown side effects
Similar to `eval` or `load` which we cannot handle fully - the functions around the `ellmer` package should be marked as such!

tmchow:
Closes #2467

Adds an `EllmerFunctions` named array (matching the convention used for `GgPlotCreate`, `TinyPlotCrate`, etc.) and a new row in `DefaultBuiltinConfig` that marks every ellmer top-level entry point as `hasUnknownSideEffects: true`. Same treatment as the existing `load` / `setwd` / `set.seed` row, just one line below it.

## Why

ellmer's `chat_*` family opens a network connection to a third-party LLM provider (or local Ollama / vLLM endpoint), and `live_console` / `live_browser` start an interactive REPL or browser session. flowr's dataflow analysis can't model the side effects of any of those calls, so they should be treated the same way as `eval` and `load` — as the issue body says, "Similar to `eval` or `load` which we cannot handle fully".

## Provider list

The chat backends mirror ellmer's [reference index](https://ellmer.tidyverse.org/reference/index.html#chat):

`chat_anthropic`, `chat_aws_bedrock`, `chat_azure_openai`, `chat_cloudflare`, `chat_cortex_analyst`, `chat_databricks`, `chat_deepseek`, `chat_github`, `chat_google_gemini`, `chat_google_vertex`, `chat_groq`, `chat_huggingface`, `chat_mistral`, `chat_ollama`, `chat_openai`, `chat_openrouter`, `chat_perplexity`, `chat_portkey`, `chat_snowflake`, `chat_vllm`

Plus `live_console` and `live_browser` for the interactive helpers.

## `forceArgs: 'all'`

I picked `forceArgs: 'all'` (same as the `cat` row a few lines down) rather than `[true]` (the `load` style) because ellmer chat calls take a system prompt as positional arg 1 plus tools / params as named args — every argument can carry data the analysis shouldn't drop. Happy to switch to `[true]` if you'd rather match the `load` row exactly; let me know.

## What's intentionally NOT in this PR

- The `Chat$chat()` / `Chat$stream()` S7 methods on the resulting Chat object — those are tracked through the object's data flow, and the call-site is what gets flagged here. Adding S7 method names here would also flag unrelated namespace collisions.
- `params(...)`, `type_*`, `content_*`, `register_tool` — pure data constructors / registry helpers; no side effect on their own.
- ggplot2-style explicit add functions or wrapper packages around ellmer (chatlas, etc.) — out of scope per "ellmer and co" referring to ellmer itself, not third-party wrappers. Easy to extend later if you want.

## Verification

- `npx tsc --project . --noEmit` — clean.
- `git diff --stat upstream/main` — 17 lines added in one file.

### DIFF
--- a/src/dataflow/environments/default-builtin-config.ts
+++ b/src/dataflow/environments/default-builtin-config.ts
@@ -92,6 +92,22 @@ export const GgPlotAddons = [
 ];
 const PlotAddons = GraphicsPlotAddons.concat(GgPlotImplicitAddons, ...PlotFunctionsWithAddParam);
 
+/**
+ * Top-level entry points of the {@link https://ellmer.tidyverse.org/ ellmer} package.
+ * Every chat backend opens a network connection to a third-party LLM provider, and
+ * `live_console` / `live_browser` start an interactive REPL or browser session, so
+ * we treat every one of these as having unknown side effects (similar to `eval` or
+ * `load`). The `chat_*` family below mirrors the providers documented at
+ * {@link https://ellmer.tidyverse.org/reference/index.html#chat}.
+ */
+export const EllmerFunctions = [
+	'chat_anthropic', 'chat_aws_bedrock', 'chat_azure_openai', 'chat_cloudflare', 'chat_cortex_analyst',
+	'chat_databricks', 'chat_deepseek', 'chat_github', 'chat_google_gemini', 'chat_google_vertex',
+	'chat_groq', 'chat_huggingface', 'chat_mistral', 'chat_ollama', 'chat_openai', 'chat_openrouter',
+	'chat_perplexity', 'chat_portkey', 'chat_snowflake', 'chat_vllm',
+	'live_console', 'live_browser'
+];
+
 const RegexConvIn = /[-/\\^$*+?.()|[\]{}]/g;
 function toRegex(n: readonly string[]): RegExp {
 	return new RegExp(`^(${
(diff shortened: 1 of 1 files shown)


EagleoutIce:
I am not sure how to handle this... This is definitely AI generated and you appear to do this to... what? Why target good first issues and so quickly; these are intended for people to get into the project and not for AIs to solve low hanging fruits - I could have solved this issue in the time it took me to write it but it allows others an easier entry. 
The solution in and of itself is also vastly incomplete. 
So mostly because this is blatantly ai generated for a feature that was intended to get people into the project I am going to reject this.

tmchow:
> I am not sure how to handle this... This is definitely AI generated and you appear to do this to... what? Why target good first issues and so quickly; these are intended for people to get into the project and not for AIs to solve low hanging fruits - I could have solved this issue in the time it took me to write it but it allows others an easier entry. The solution in and of itself is also vastly incomplete. So mostly because this is blatantly ai generated for a feature that was intended to get people into the project I am going to reject this.

?

AI assisted not AI generated.  The AI generated part if you're thinking is due to the PR description?  when i run a pr description skill it generated the pr desc which you think this?  happy to move on if the solution didn't fit!

EagleoutIce:
I think this because it took you 6 minutes to 
1) see an issue that was just opened and marked as good first issue, 
2) read and understand it, 
3) fork the repository and setup the development by installing the dependencies etc. (assuming you even bothered to think about running the test suite - which you didn't extend as well), 
4) do the necessary research to even notice which functions you would need to support, 
5) find the correct location in the source code and understand the structure to add your changes
6) commit with message, open the PR with the message

This is not AI assisted. From your channel and recent activity you appear to be changelog mining in as many different projects as you can. I cannot vouch for however useful your other contributions are, but the "good first issue" is, according to its own description intended **for newcomers** and you, even if I am willing to believe that you did the steps above AI assisted and not fully automated - which makes you some kind of 10X engineer, are definitely not a newcomer that needs to take these issues. 
Sorry but at least for this repository I think you should move on.

PR #2470, opened 2026-05-05
Validity prototype

MaxAtoms:
I allowed myself to just experiment for a bit. It's a very crude untested prototype right now.
~~Demo available via `npm run taint-test`.~~


src/taint-analysis/scale-domain.ts
```diff
@@ -0,0 +1,101 @@
+import { AbstractDomain } from '../abstract-interpretation/domains/abstract-domain';
+import { Bottom, Top } from '../abstract-interpretation/domains/lattice';
+
+export const Unscaled = Symbol('Unscaled');
+export const Scaled = Symbol('Scaled');
+
+type Lift = typeof Scaled | typeof Unscaled | typeof Top | typeof Bottom;
+
+// TODO Builder pattern in the DSL for the definition of finite taint lattices
+export class ScaleDomain<Value extends Lift = Lift> extends AbstractDomain<Lift, Lift, typeof Top, typeof Bottom, Value> {
```
MaxAtoms: Using the `AbstractInterpretationVisitor` has been fine so far. My biggest gripe is the need to define the type of concrete values and abstraction/concretization functions

src/taint-analysis/predefined/predefined.ts
```diff
@@ -0,0 +1,12 @@
+import { scaleAnalysis } from './scale-analysis';
+import type { TaintAnalysisDefinition, TaintAnalysisName } from '../builder/taint-analysis-definition';
+
+export const predefinedTaintAnalyses = {
+	'scale': scaleAnalysis,
+} as const satisfies AnalysisMap<['scale']>;
```
MaxAtoms, line 6: How can I actually make this type safe so that the key name has to be the same literal string as the `Name` of the respective `TaintAnalysisDefinition<Name>`?

MaxAtoms:
Plz roast when you find the time, @EagleoutIce 😇

src/taint-analysis/predefined/predefined.ts
```diff
@@ -0,0 +1,12 @@
...
+
+type AnalysisMap<Defs extends readonly string[]> = {
+	[key in TaintAnalysisName<TaintAnalysisDefinition<Defs[number]>>]: TaintAnalysisDefinition<Defs[number]>;
```
EagleoutIce: i think you can make it type safe here, if you reuse `key` on the rhs?

PR #2472, opened 2026-05-06
Trigger on Packages with Known Information

### LINKED ISSUE #2072: Trigger known packages!
Consider:

```r
library(ggplot2)

ggplot()
```

I want a `reads | calls` edge from `ggplot` to the `ggplot2` library **if** we know that `ggplot2` exports the `ggplot` function. Additionally, I want to add the information to `ggplot`, that it comes from the `ggplot2` package.
So in other words, we want to link calls from package functions to the package that provides them.

This information is provided by the [dependencies context](https://github.com/flowr-analysis/flowr/wiki/Analyzer#Dependencies_Context).

**Please note:** Right now, flowR does not have dependencies for packages loaded (this is the step *after* this one with which we include the results from crawlR) so whenever you want to test this, you have to *tell* flowR what every package is exporting (but just for the tests, for this you can have a look at the namespace file tests which we already have.

It should mainly work like this: in the built-in-library handler:

https://github.com/flowr-analysis/flowr/blob/613d91fad67723f6ce9621c121f50775ab182a76/src/dataflow/internal/process/functions/call/built-in/built-in-library.ts#L19-L25

I want access knowledge that we have about this package (with the dependencies context of the flowr analyzer context). Similar to `assignments`
https://github.com/flowr-analysis/flowr/blob/613d91fad67723f6ce9621c121f50775ab182a76/src/dataflow/internal/process/functions/call/built-in/built-in-assignment.ts#L410

 this should add the exported variables to the current environment. In other words, if we know `ggplot2` exports `ggplot`, the built-in library handler with `library(ggplot2)` should assign `ggplot` to point to the package (i.e., add an according Identifier Definition). Then, every variable use of `ggplot` etc. should link to this definition (you can set `definedAt` to `built-in:ggplot` when defining it in the environment).

EagleoutIce:
Still open:

* [x] Correctly identify the name of the package that is loaded, not just the lexeme (e.g., with: `x <- "foo"; library(x, character.only = TRUE)` should recognize that we load `foo` and not `x`, likewise with the `lapply` etc. pattern)
* [x] Not every exported symbol is a function call => we need information for each symbol, whether it is (not-)callable!
* [x] We need to follow "re-exports" with `import` etc. (`import(grid)`); but we have to check whether already loaded
* [x] We need to support `exportPattern...` 

In the future, we want to include the information and we have to respect versions.


src/dataflow/environments/environment.ts
```diff
@@ -49,6 +49,8 @@ export class Environment implements IEnvironment {
 	readonly id: number;
 	/** Optional name for namespaced/non-anonymous environments, please only set if you know what you are doing */
 	n?:          string;
+	/** to keep track if/whether environment was added as package/namespace/imports environment */
+	t?:          'package' |'namespace' |'imports';
```
EagleoutIce: please make this an enum :)

```ts
enum EnvType {
  Package = 'pkg',
  Namespace = 'ns',
  Imports = `imp`
}
```

src/dataflow/internal/process/functions/call/built-in/built-in-library.ts
```diff
@@ -23,33 +34,188 @@ export function processLibrary<OtherInfo>(
...
+}*/
+
+function linkLibrary<OtherInfo>(dependency: Package, info: DataflowInformation, rootId: NodeId, data: DataflowProcessorInformation<OtherInfo & ParentInformation>) {
+	if(info.environment.level < 0|| isUndefined(dependency.namespaceInfo)){
+		return;
+	}
+	const currentEnv = info.environment.current;
+	const pack = dependency.name;
+	const functions = dependency.namespaceInfo.callable;
+	/*
```
EagleoutIce: please remove comment blocks / clean up

test/functionality/dataflow/main/libraries/link-lib-character-only.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import { EdgeType } from '../../../../../src/dataflow/graph/edge';
+import { FlowrNamespaceFile, setCallable } from '../../../../../src/project/plugins/file-plugins/files/flowr-namespace-file';
+import { FlowrInlineTextFile } from '../../../../../src/project/context/flowr-file';
+import { emptyGraph } from '../../../../../src/dataflow/graph/dataflowgraph-builder';
+import { NodeId } from '../../../../../src/r-bridge/lang-4.x/ast/model/processing/node-id';
+import { label } from '../../../_helper/label';
+
+const ggplot2Callable = ['+', 'ggplot', 'aes', 'geom_point', 'geom_line', 'theme_bw', 'coord_cartesian', 'ggsave', 'fortify', 'scale_type'];
+
+describe('Link libraries with character.only', withTreeSitter(ts => {
```
EagleoutIce: I'd prefer at least two more tests:

1) loading pkgA and pkgB with the code needing variables and functions from both packages
2) loading pkgA and pkgB with pkgB overwriting some of the bindings of pkgA, just to check that:
   - `pkgA::var` still works
   - `var` resolves to `pkgB::var`

test/functionality/dataflow/main/libraries/link-lib-character-only.test.ts
```diff
@@ -0,0 +1,155 @@
...
+	assertDataflow(label('Link with variable'), ts, 'x <- "ggplot2"\nlibrary(x, character.only = TRUE)\nggplot()\nggplot()',
+		emptyGraph()
+		//reads character.only
+			.addEdge(9, 8, EdgeType.Argument)
+			.addEdge(8, 7, EdgeType.Reads)
+		//ggplot links to library
+			.addEdge(11, NodeId.toBuiltIn('ggplot'), EdgeType.Reads |EdgeType.Calls)
```
EagleoutIce: please make the bit-or spacing consistent ;)

test/functionality/dataflow/main/libraries/link-lib-character-only.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import { NodeId } from '../../../../../src/r-bridge/lang-4.x/ast/model/processing/node-id';
+import { label } from '../../../_helper/label';
+
+const ggplot2Callable = ['+', 'ggplot', 'aes', 'geom_point', 'geom_line', 'theme_bw', 'coord_cartesian', 'ggsave', 'fortify', 'scale_type'];
+
+describe('Link libraries with character.only', withTreeSitter(ts => {
+	assertDataflow(label('Link with variable'), ts, 'x <- "ggplot2"\nlibrary(x, character.only = TRUE)\nggplot()\nggplot()',
+		emptyGraph()
+		//reads character.only
+			.addEdge(9, 8, EdgeType.Argument)
```
EagleoutIce: You could have used:


```suggestion
			.addEdge('2@library', '2@character.only', EdgeType.Argument)
```

test/functionality/dataflow/main/libraries/link-lib-character-only.test.ts
```diff
@@ -0,0 +1,155 @@
...
+		//correctly links
+			.addEdge(8, NodeId.toBuiltIn('ggplot'), EdgeType.Reads |EdgeType.Calls)
+			.addEdge(10, NodeId.toBuiltIn('ggplot'), EdgeType.Reads |EdgeType.Calls)
+			.addEdge(NodeId.toBuiltIn('ggplot'), 6, EdgeType.Reads | EdgeType.Calls),
+		{
+			modifyAnalyzer: a => {
+				a.context().deps.addDependency(new Package({
+					name:          'ggplot2',
+					namespaceInfo: setCallable(FlowrNamespaceFile.from(new FlowrInlineTextFile('NAMESPACE', `S3method(fortify,data.frame)
+S3method(fortify,lm)
```
EagleoutIce: I'd prefer it if you would soursource the `FlowrInlineTextFile` namespace object into a constant and not copy-paste it.

test/functionality/dataflow/main/libraries/link-lib-character-only.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Package } from '../../../../../src/project/plugins/package-version-plugins/package';
+import { EdgeType } from '../../../../../src/dataflow/graph/edge';
+import { FlowrNamespaceFile, setCallable } from '../../../../../src/project/plugins/file-plugins/files/flowr-namespace-file';
+import { FlowrInlineTextFile } from '../../../../../src/project/context/flowr-file';
+import { emptyGraph } from '../../../../../src/dataflow/graph/dataflowgraph-builder';
+import { NodeId } from '../../../../../src/r-bridge/lang-4.x/ast/model/processing/node-id';
+import { label } from '../../../_helper/label';
+
+const ggplot2Callable = ['+', 'ggplot', 'aes', 'geom_point', 'geom_line', 'theme_bw', 'coord_cartesian', 'ggsave', 'fortify', 'scale_type'];
+
```
EagleoutIce:
```ts
const namespaceContent = `S3method(fortify,data.frame)
S3method(fortify,lm)
S3method(fortify,default)
S3method(fortify,map)
S3method(scale_type,Date)
S3method(scale_type,POSIXt)
S3method(scale_type,character)
S3method(scale_type,numeric)
S3method(scale_type,factor)
S3method(scale_type,default)
S3method(ggplot,default)
S3method(print,ggproto)
S3method(print,rel)
export("+")
export(ggplot)
export(aes)
export(geom_point)
export(geom_line)
export(theme_bw)
export(coord_cartesian)
export(ggsave)
export(fortify)
export(scale_type)
exportPattern("^[^\\\\.]\\\\.*$")
import(grid)
import(rlang)
importFrom(scales,alpha)
importFrom(stats,setNames)
`
```

test/functionality/dataflow/main/libraries/link-lib-imports.test.ts
```diff
@@ -0,0 +1,259 @@
...
+
+		const df = await analyzer.dataflow();
+		let env = df.environment.current;
+		const ggplotSymbols = ['+', 'ggplot', 'aes', 'geom_point', 'geom_line', 'theme_bw', 'coord_cartesian', 'ggsave', 'fortify', 'scale_type'];
+		const rP1symbols = ['test1', 'test2'];
+		const rP2Symbols = ['test1', 'test2', 'test3'];
+		const rP3Symbols = ['a', 'b'];
+		// random_placeholder3 namespace -> random_placeholder3 imports -> random_placeholder2 namespace -> random_placeholder2 imports -> ggplot2 namespace -> ggplot2 imports -> global environment
+		const imp3 = ['random_placeholder2:test3'];
+		const imp2 = ['random_placeholder3:a'].concat(ggplotSymbols.map(v => 'ggplot2:' + v)).concat(rP1symbols.map(v => 'random_placeholder:' + v));
```
EagleoutIce: your id constructions are technically fragile; maybe a function?

src/dataflow/internal/process/functions/call/built-in/built-in-library.ts
```diff
@@ -23,33 +34,188 @@ export function processLibrary<OtherInfo>(
...
+		const values = valueSetGuard(resolveIdToValue(charId[0], resolveArgs));
+		if(values?.type === 'set' && values.elements.length === 1 && values.elements[0].type === 'logical') {
+			isCharacterOnly = values.elements[0].value;
 		}
-	};
-	return processKnownFunctionCall({
-		name, args:                 wrapArgumentsUnnamed([newArg], data.completeAst.idMap), rootId, data,
-		hasUnknownSideEffect: true,
+	}
+	if(isCharacterOnly !== false){
```
EagleoutIce: if it is not known to be true (if it could be maybe) we must also consider that it could be the name of the symbol! 

```R
x <- "foo"
library(x, character.only=u)
```

=> we must assume that it loads either `x` or `foo`

src/dataflow/internal/process/functions/call/built-in/built-in-library.ts
```diff
@@ -23,33 +34,188 @@ export function processLibrary<OtherInfo>(
...
+		const values = valueSetGuard(resolveIdToValue(nameToLoad.info.id, resolveArgs));
+		if(values?.type === 'set' && values.elements.length !== 0){
+			if(values.elements[0].type === 'string' && 'str' in values.elements[0].value){
+				packetName =  values.elements[0].value.str;
```
EagleoutIce: if there are multiple options we shouldn't just lock in on the first and throw away the rest. 

1. we could log this (`log.warn(...)`)
2. we could load *all* these libraries similar to if we would have:
   ```R
   if(u) {
     library(a)
   } else {
     library(b)
   }
   # ------
   if(u) {
     x <- "a"
   } else {
     x <- "b"
   }
   library(x, character.only=TRUE)
   ```

=> please with tests.

src/dataflow/internal/process/functions/call/built-in/built-in-library.ts
```diff
@@ -23,33 +35,214 @@ export function processLibrary<OtherInfo>(
...
+			}
+			if(hasMaybe > 0){
+				isCharacterOnly = 'maybe';
+			} else if(hasTrue === 0 && hasFalse > 0){
+				isCharacterOnly = false;
+			} else if(hasTrue > 0 && hasFalse === 0){
+				isCharacterOnly = true;
+			} else {
+				isCharacterOnly = 'maybe';
+			}
```
EagleoutIce: bitte lager das aus das ist ja nichts anderes als ein `ternaryAnd` auf sets das kannst du in die tenray datei packen :D

src/dataflow/environments/environment.ts
```diff
@@ -32,6 +32,11 @@ export interface IEnvironment {
 	builtInEnv?: true | undefined
 }
 
+export enum EnvType{
+	Namespace = 'ns',
```
EagleoutIce, line 33: wieso ist package verschwunden?

test/functionality/dataflow/main/libraries/link-libraries.test.ts
```diff
@@ -111,6 +111,32 @@ describe('Link libraries', withTreeSitter(ts => {
 		}
 	);
 
+	assertDataflow(label('pkgB overwrites bindings of pkgA'), ts, 'library(pkgA)\nlibrary(pkgB)\nx()\npkgA::x()\ny()',
```
EagleoutIce: kannst du das noch so anpassen dass wir vor librarypkGB auch mal `x` laden? damit dass da sicher auf pkgA gelinkt wrid?

EagleoutIce: mit `` ` `` kannst du das auch direkt multiline schreiben :D

test/functionality/dataflow/main/libraries/link-libraries.test.ts
```diff
@@ -203,6 +203,42 @@ describe('Link libraries', withTreeSitter(ts => {
 		expect(env.n === 'ggplot2' && env.t === EnvType.Imports).toBeTruthy();
 		expect(new Set(env.memory.keys()).size === 0).toBeTruthy();
 	});
+
+	/*test('Multiple options for library', async() => {
```
EagleoutIce: wurde das hier noch entfernt? blick gerade bei den changes nich mehr ganz durch

EagleoutIce (commented):
supii, dann eigentlich nur noch die konflikte lösen und wr können das mergen :)

EagleoutIce:
_This pull request is included in v2.11.0 (see [Release v2.11.0 (Package Database Support)](https://github.com/flowr-analysis/flowr/releases/tag/v2.11.0))._

PR #2475, opened 2026-05-07
[AbsInt] Provide a generic testing framework for abstract interpretation

OliverGerstl:

PR #2480, opened 2026-05-14
2451 Numeric Pentagon Inference (Iteration 2)

schuler-henry:
Implemented the closed pentagon domain as state domain of a value product with the same expression/condition semantics as the interval inference implementation.
Added support for divide, power, and, and or semantics to both domains.

This concludes the changes for iteration 2.


src/abstract-interpretation/interval/eval.ts
```diff
@@ -45,7 +46,15 @@ async function instrument(folder: string, workingDir: string, outputFolder: stri
 		ctx:           analyzer.inspectContext()
 	});
 
+	const pentagonVisitor = new NumericPentagonInferenceVisitor({
+		normalizedAst: await analyzer.normalize(),
```
EagleoutIce, line 50: ein bisschen wild das zweimal zu awaiten aber das passt für mich :D

src/abstract-interpretation/interval/eval.ts
```diff
@@ -90,6 +102,7 @@ async function instrument(folder: string, workingDir: string, outputFolder: stri
 								`paste(length(${lhs}), collapse=""), ` +
 								`paste0('"', gsub('"', '""', gsub("\\n", " ", ifelse(is.numeric(${lhs}), paste(${lhs}, collapse=","), ""))), '"'), ` +
 								`paste('"', "${inferredValue?.toString() ?? 'undefined'}", '"', sep=""), ` +
```
EagleoutIce: why sometimes paste0 but then use paste with an empty sep?

src/abstract-interpretation/interval/eval.ts
```diff
@@ -82,6 +93,7 @@ async function instrument(folder: string, workingDir: string, outputFolder: stri
 				after:  '; ' +
 							'cat(' +
 								`"${resolved.lhs.info.id}", ` +
+								`paste('"', "[${lhsOrigins.join(', ')}]", '"', sep=""), ` +
```
EagleoutIce: das recorded dann die ids oder? an sich fein ig aber man könnte die locations noch adden dann weiß man wos passiert.

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -4,15 +4,18 @@ import { isNotUndefined, isUndefined } from '../../util/assert';
...
+import type { NodeId } from '../../r-bridge/lang-4.x/ast/model/processing/node-id';
 
 /**
  * Maps function/operator names to the semantic functions.
  */
-const IntervalExpressionSemanticsMapper = [
-	[Identifier.make('+'), unaryBinaryExprOpSemantics(intervalPositiveOp, intervalAddOp)],
-	[Identifier.make('-'), unaryBinaryExprOpSemantics(intervalNegativeOp, intervalSubtractOp)],
+export const IntervalExpressionSemanticsMapper = [
+	[Identifier.make('+'), unaryBinaryExprOpSemantics(intervalUnaryIdentityOp, intervalAddOp)],
```
EagleoutIce, line 21: bei unary könnte man uu. noch was über den Ergebnis typ inferren, aber das reicht als FW

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -38,20 +41,20 @@ type BinaryOpSemantics = (left: IntervalDomain | undefined, right: IntervalDomai
...
 /**
  * Semantics definition function n-ary functions/operators.
  * @param args - The raw arguments of the function/operator.
- * @param visitor - The numeric inference visitor performing the analysis used to resolve argument intervals.
+ * @param getInterval - A function to retrieve the current inferred interval for the provided node id.
  * @param significantFigures - The number of significant figures used to create new intervals.
  * @returns The resulting interval after applying the semantics.
  */
-type NaryFnSemantics = (args: readonly FunctionArgument[], visitor: NumericIntervalInferenceVisitor, significantFigures: number | undefined) => IntervalDomain | undefined;
+type NaryFnSemantics = (args: readonly FunctionArgument[], getInterval: (node: NodeId) => IntervalDomain | undefined, significantFigures: number | undefined) => IntervalDomain | undefined;
```
EagleoutIce: wenn du das immer wieder nutzt kannst du auch einen `type GetInterval = (...` machen :)

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -71,7 +74,7 @@ export function applyIntervalExpressionSemantics(functionIdentifier: Identifier,
 
 	const [_, semantics] = match;
 
-	return semantics(args, visitor, significantFigures);
+	return semantics(args, (node: NodeId ) => visitor.getAbstractValue(node), significantFigures);
```
EagleoutIce: spacing :C

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -84,19 +87,19 @@ export function applyIntervalExpressionSemantics(functionIdentifier: Identifier,
  * @returns A semantics function for n-ary functions that applies the correct provided operator based on the provided arguments.
  */
 function unaryBinaryExprOpSemantics(unaryOperatorSemantics: UnaryOpSemantics, binaryOperatorSemantics: BinaryOpSemantics): NaryFnSemantics {
-	return (args: readonly FunctionArgument[], visitor: NumericIntervalInferenceVisitor, significantFigures: number | undefined): IntervalDomain | undefined => {
+	return (args: readonly FunctionArgument[], getInterval: (node: NodeId) => IntervalDomain | undefined, significantFigures: number | undefined): IntervalDomain | undefined => {
 		// Usage as unary operator
 		if(args.length === 1) {
 			if(FunctionArgument.isEmpty(args[0])) {
 				numericInferenceLogger.warn('Called unary operator with an empty argument, which is not supported.');
 				return IntervalDomain.bottom(significantFigures);
```
EagleoutIce, line 102: reiterating, aber wollen wir unsupported operations mit bottom muten? und nich mit top?

src/abstract-interpretation/interval/numeric-interval-inference.ts
```diff
@@ -62,15 +63,33 @@ export class NumericIntervalInferenceVisitor extends AbstractInterpretationVisit
...
 		let result: StateAbstractDomain<IntervalDomain> | undefined;
 
+		const { applyConditionSemantics: intervalPositiveSemantics, applyNegatedConditionSemantics: intervalNegativeSemantics } = getIntervalConditionSemantics<StateAbstractDomain<IntervalDomain>, this>();
+
 		if(trueBranch) {
-			result = applyIntervalConditionSemantics(conditionNodeId, state, this, this.config.dfg);
+			result = intervalPositiveSemantics(
+				conditionNodeId,
+				state,
+				this,
```
EagleoutIce, line 93: es fühlt sich ein wenig seltsam an this zu übergeben und dann immer auch den `this.config.cfg`

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+import type { AbstractInterpretationVisitor } from '../../absint-visitor';
+
+/**
+ * Interface that needs to be implemented by any {@link AbstractInterpretationVisitor} that applies upper bounds
+ * condition semantics.
+ */
+export interface UpperBoundsDomainAccess<StateDomain extends AnyStateDomain<AnyAbstractDomain>> {
+	setUpperBounds(state: StateDomain): (node: NodeId, value: UpperBoundsValueDomain) => void;
+	getUpperBounds(node: NodeId, state?: StateDomain): UpperBoundsValueDomain;
+	getUniqueOrigin(node: NodeId): NodeId | undefined;
```
EagleoutIce: please document the semantics of `unique` in this setup maybe more a `getOriginIfUnique`

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	return [
+		[Identifier.make('=='), binaryConditionSemanticsGuard(upperBoundsEqualsOp), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp)],
+		[Identifier.make('!='), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp), binaryConditionSemanticsGuard(upperBoundsEqualsOp)],
+		[Identifier.make('>'), binaryConditionSemanticsGuard(upperBoundsGreaterOp), binaryConditionSemanticsGuard(upperBoundsLessEqualOp)],
+		[Identifier.make('>='), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp), binaryConditionSemanticsGuard(upperBoundsLessOp)],
+		[Identifier.make('<'), binaryConditionSemanticsGuard(upperBoundsLessOp), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp)],
+		[Identifier.make('<='), binaryConditionSemanticsGuard(upperBoundsLessEqualOp), binaryConditionSemanticsGuard(upperBoundsGreaterOp)],
+		[Identifier.make('is.na'), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
```
EagleoutIce: braucht es das satisfies? der return type carried ja die selben Constraints hier

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+function getUpperBoundsSemanticsMapper<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends UpperBoundsConditionSemanticsVisitor<StateDomain>>(): ConditionSemanticsMapperInfo<StateDomain, Visitor>[] {
+	return [
+		[Identifier.make('=='), binaryConditionSemanticsGuard(upperBoundsEqualsOp), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp)],
+		[Identifier.make('!='), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp), binaryConditionSemanticsGuard(upperBoundsEqualsOp)],
+		[Identifier.make('>'), binaryConditionSemanticsGuard(upperBoundsGreaterOp), binaryConditionSemanticsGuard(upperBoundsLessEqualOp)],
+		[Identifier.make('>='), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp), binaryConditionSemanticsGuard(upperBoundsLessOp)],
+		[Identifier.make('<'), binaryConditionSemanticsGuard(upperBoundsLessOp), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp)],
+		[Identifier.make('<='), binaryConditionSemanticsGuard(upperBoundsLessEqualOp), binaryConditionSemanticsGuard(upperBoundsGreaterOp)],
```
EagleoutIce: maybe explain why `<=` (less equal uses `GreaterOp`) somewhere

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	getUniqueOrigin(node: NodeId): NodeId | undefined;
+}
+
+type UpperBoundsConditionSemanticsVisitor<StateDomain extends AnyStateDomain<AnyAbstractDomain>> = AbstractInterpretationVisitor<StateDomain> & UpperBoundsDomainAccess<StateDomain>;
+
+/**
+ * Wrapper for the upper bounds condition semantics that adds all upper bounds specific condition semantics and returns
+ * the applyConditionSemantics and applyNegatedConditionSemantics functions.
+ */
+export function getUpperBoundsConditionSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends UpperBoundsConditionSemanticsVisitor<StateDomain>>(): { applyConditionSemantics: UnaryConditionSemantics<StateDomain, Visitor>, applyNegatedConditionSemantics: UnaryConditionSemantics<StateDomain, Visitor> } {
```
EagleoutIce: maybe an interface for the semantics ret value? or make the dispatch transparent with `ReturnType<typeof create...>`

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+		[Identifier.make('!='), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp), binaryConditionSemanticsGuard(upperBoundsEqualsOp)],
+		[Identifier.make('>'), binaryConditionSemanticsGuard(upperBoundsGreaterOp), binaryConditionSemanticsGuard(upperBoundsLessEqualOp)],
+		[Identifier.make('>='), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp), binaryConditionSemanticsGuard(upperBoundsLessOp)],
+		[Identifier.make('<'), binaryConditionSemanticsGuard(upperBoundsLessOp), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp)],
+		[Identifier.make('<='), binaryConditionSemanticsGuard(upperBoundsLessEqualOp), binaryConditionSemanticsGuard(upperBoundsGreaterOp)],
+		[Identifier.make('is.na'), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
+}
+
+// Semantics
```
EagleoutIce: personally I'd refrain from using segment comments like this 😅  they become outdated easily

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	if(resultingUpperBounds.isBottom()) {
+		return state.bottom();
+	}
+
+	if(isNotUndefined(leftOrigin)) {
+		visitor.setUpperBounds(state)(leftOrigin, resultingUpperBounds);
+	}
+	if(isNotUndefined(rightOrigin)) {
+		visitor.setUpperBounds(state)(rightOrigin, resultingUpperBounds);
+	}
```
EagleoutIce, line 62: Shouldn't you cache the setUpperBounds?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	if(leftValue.isBottom() || rightValue.isBottom()) {
+		return state.bottom();
+	}
+
+	// We can only argue if both sides have exactly one origin (or are the origin).
+	// In that case, if both sides include each others origins, they must be equal and therefore cannot be inequal.
+	const leftOrigin = visitor.getUniqueOrigin(leftNodeId);
+	const rightOrigin = visitor.getUniqueOrigin(rightNodeId);
+	if(isNotUndefined(leftOrigin) && isNotUndefined(rightOrigin )&& leftValue.has(rightOrigin) && rightValue.has(leftOrigin)) {
```
EagleoutIce: spacing :D

EagleoutIce: + if you ensure that neither left nor right value has undefined, the first checks are redundant

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+
+function upperBoundsGreaterEqualOp<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends UpperBoundsConditionSemanticsVisitor<StateDomain>>(leftNodeId: NodeId, rightNodeId: NodeId, state: StateDomain, visitor: Visitor): StateDomain {
+	const leftValue = visitor.getUpperBounds(leftNodeId, state);
+	const rightValue = visitor.getUpperBounds(rightNodeId, state);
+
+	const leftOrigin = visitor.getUniqueOrigin(leftNodeId);
+	const rightOrigin = visitor.getUniqueOrigin(rightNodeId);
+
+	// We want to upper bounds of the right side, which we can only do if there is exactly on origin that we can update.
+	if(isUndefined(rightOrigin)) {
```
EagleoutIce, line 92: why get the left origin and all the other boudns before the check then? :D

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
```diff
@@ -0,0 +1,152 @@
...
+		} else if(this.value === Bottom || other.value === Bottom) {
+			return false;
+		}
+		return setEquals(this.value, other.value);
+	}
+
+	public leq(other: this): boolean {
+		if(this.isBottom()) {
+			return true;
+		} else if(other.isBottom()) {
```
EagleoutIce: given there is an upper bounds top, why not include this? `this.isBot() || otherwise.isTop()`

EagleoutIce: this could then probably reuse `SetDomain` semantics

src/abstract-interpretation/pentagon/closed-pentagon-domain.ts
```diff
@@ -0,0 +1,157 @@
...
+			}
+
+			if(pentagon.value.upperBounds.has(key)) {
+				removeBounds.add(key);
+			}
+
+			if(newInferredBounds.size > 0 || removeBounds.size > 0) {
+				const reducedPentagon = pentagon.create(pentagon.value);
+				newInferredBounds.values().forEach(bound => reducedPentagon.value.upperBounds.add(bound));
+				removeBounds.values().forEach(bound => reducedPentagon.value.upperBounds.remove(bound));
```
EagleoutIce: i'd evoid using `.forEach`

src/abstract-interpretation/pentagon/closed-pentagon-value-domain.ts
```diff
@@ -0,0 +1,48 @@
...
+/** The type of the abstract product representing the closed pentagon */
+export type AbstractClosedPentagon = {
+	interval:    IntervalDomain;
+	upperBounds: UpperBoundsValueDomain;
+};
+
+/**
+ * The weakly relational closed pentagon domain is product combining the interval state domain and the upper-bounds domain.
+ * To remove the duplicate mapping from NodeId to values
+ * (interval state domain: NodeId -&gt; interval, upper bounds: NodeId -&gt; Set(NodeId)),
```
EagleoutIce: if you use backticks you do not have to do html escapes

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,270 @@
...
+ * @param target - Node ID of the result node.
+ * @param functionIdentifier - The {@link Identifier} of the function/operator.
+ * @param args - The arguments of the function/operator.
+ * @param visitor - The pentagon inference visitor performing the analysis used to resolve nodes.
+ * @param currentState - The current state in the inference process, to update the upper bounds of other nodes.
+ * @param significantFigures - The number of significant figures used to create new intervals.
+ * @returns The resulting pentagon after applying the semantics.
+ */
+export function applyPentagonExpressionSemantics(target: NodeId, functionIdentifier: Identifier, args: readonly FunctionArgument[], visitor: NumericPentagonInferenceVisitor, currentState: ClosedPentagonDomain, significantFigures?: number): ClosedPentagonValueDomain | undefined {
+	const match = PentagonExpressionSemanticsMapper.find(([id]) => Identifier.matches(id, functionIdentifier));
```
EagleoutIce, line 48: if you add a lot more fns in the third iter we can discuss how to speed this up!

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,270 @@
...
+		} else {
+			const [_, semantics] = intervalMatch;
+
+			const interval = semantics(args, (node: NodeId) => visitor.getAbstractValue(node)?.value.interval, significantFigures);
+
+			if(isUndefined(interval)) {
+				return undefined;
+			} else {
+				const pentagon = ClosedPentagonValueDomain.top(significantFigures);
+				pentagon.value.interval = interval;
```
EagleoutIce: is the only reason you set it like this to avoid a refinement step?

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		const left = visitor.getAbstractValue(args[0].nodeId);
+		const right = visitor.getAbstractValue(args[1].nodeId);
+
+		return binaryOperatorSemantics(target, [args[0].nodeId, left], [args[1].nodeId, right], currentState, visitor, significantFigures);
+	};
+}
+
+function _unaryExprFnSemantics(unaryFunctionSemantics: UnaryFnSemantics): NaryFnSemantics {
```
EagleoutIce: maybe remove if unneded?

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,270 @@
...
+				addTargetAsUpperBoundToNodeInState(target, argOrigin, currentState);
+			}
+		}
+
+		return targetPentagon;
+	}
+}
+
+function pentagonSubtractOp(target: NodeId, left: [NodeId, ClosedPentagonValueDomain | undefined], right: [NodeId, ClosedPentagonValueDomain | undefined], currentState: ClosedPentagonDomain, visitor: NumericPentagonInferenceVisitor): ClosedPentagonValueDomain | undefined {
+	const [leftNodeId, leftValue] = left;
```
EagleoutIce: you could do the destructuring directly in the parameters

src/abstract-interpretation/pentagon/numeric-pentagon-inference.ts
```diff
@@ -0,0 +1,179 @@
...
+
+	protected override applyConditionSemantics(state: ClosedPentagonDomain, conditionNodeId: NodeId, trueBranch: boolean): ClosedPentagonDomain {
+		const reducePentagon = (state: ClosedPentagonDomain) => state.create(state.value);
+
+		const { applyConditionSemantics: intervalPositiveSemantics, applyNegatedConditionSemantics: intervalNegativeSemantics } = getIntervalConditionSemantics<ClosedPentagonDomain, this>();
+		const { applyConditionSemantics: upperBoundsPositiveSemantics, applyNegatedConditionSemantics: upperBoundsNegativeSemantics } = getUpperBoundsConditionSemantics<ClosedPentagonDomain, this>();
+
+		if(trueBranch) {
+			return reducePentagon(
+				upperBoundsPositiveSemantics(
```
EagleoutIce: the fn you call is the only difference right? we could also merge this with `const fn = trueBranch ? upperBoundsPos.. : upperBoundsNeg...

src/abstract-interpretation/absint-condition-semantics.ts
```diff
@@ -0,0 +1,174 @@
...
+import type { AnyStateDomain } from './domains/state-domain-like';
+import type { AnyAbstractDomain } from './domains/abstract-domain';
+import type { AbstractInterpretationVisitor } from './absint-visitor';
+
+export type NAryConditionSemantics<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends AbstractInterpretationVisitor<StateDomain>> =
+	(
+		argNodeIds: readonly (NodeId | undefined)[],
+		state: StateDomain,
+		visitor: Visitor,
+		dfg: DataflowGraph
```
EagleoutIce, line 19: you could also spread in the last three if there always supposed to be there

test/functionality/abstract-interpretation/interval/interval.ts
```diff
@@ -58,13 +58,13 @@ export const IntervalTests = {
 };
 
 export enum DomainMatchingType {
-	Exact = 'exact',
+	Equal = 'equal',
```
EagleoutIce, line 61: why the weakening?

test/functionality/abstract-interpretation/pentagon/closed-pentagon-domain.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+	assertAbstractDomain(createDomain, ClosedPentagonBottom, ClosedPentagonTop, {
+		equal: false, leq: true, join: ClosedPentagonTop, meet: ClosedPentagonBottom, concrete: [], abstract: ClosedPentagonBottom, widen: ClosedPentagonTop
+	});
+
+	assertAbstractDomain(createDomain, ClosedPentagonBottom, ClosedPentagonBottom, {
+		equal: true, leq: true, join: ClosedPentagonBottom, meet: ClosedPentagonBottom, concrete: [], abstract: ClosedPentagonBottom, widen: ClosedPentagonBottom
+	});
+
+	// Compare {V ∈ [0, 10], V <= {V, W}} and {V ∈ [0, 10], V <= {W}}
```
EagleoutIce, line 48: are self bounds not ignored automatically?

test/functionality/abstract-interpretation/pentagon/interim-presentation-example.test.ts
```diff
@@ -0,0 +1,39 @@
...
+			}
+			print(z)
+		`, {
+		'6@a':  { interval: IntervalTests.interval(-Infinity, Infinity), upperBounds: UpperBoundsTests.top() },
+		'6@x':  { interval: IntervalTests.interval(1, Infinity), upperBounds: UpperBoundsTests.top() },
+		'7@b':  { interval: IntervalTests.interval(-Infinity, Infinity), upperBounds: UpperBoundsTests.top() },
+		'7@y':  { interval: IntervalTests.interval(1, Infinity), upperBounds: UpperBoundsTests.top() },
+		'10@x': { interval: IntervalTests.bottom(), upperBounds: UpperBoundsTests.bottom() },
+		'10@y': { interval: IntervalTests.bottom(), upperBounds: UpperBoundsTests.bottom() },
+		'20@z': { interval: IntervalTests.bottom(), upperBounds: UpperBoundsTests.bottom() },
```
EagleoutIce, line 36: why not smth like Pentagons.bot?

test/functionality/abstract-interpretation/pentagon/pentagon.ts
```diff
@@ -0,0 +1,200 @@
...
+		analyzer.addRequest(code.trim());
+
+		ast = await analyzer.normalize();
+		const dfg = (await analyzer.dataflow()).graph;
+		const cfg = await analyzer.controlflow();
+		const ctx = analyzer.inspectContext();
+
+		visitor = new NumericPentagonInferenceVisitor({
+			normalizedAst: ast,
+			dfg:           dfg,
```
EagleoutIce: in these cases you can also just write `dfg, `

test/functionality/abstract-interpretation/interval/thesis-example.test.ts
EagleoutIce, line 37: does this test not exist for pentagons currently?

EagleoutIce (approved):

src/abstract-interpretation/interval/eval.ts
schuler-henry: Me no R guy!

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -1,177 +1,97 @@
...
- * @param argNodeId - The node id of the argument of the unary operator.
- * @param state - The state to retrieve the argument values and apply the semantics to.
- * @param visitor - The numeric inference visitor performing the analysis used to resolve argument intervals.
- * @param dfg - The dataflow graph containing the vertex and its arguments.
- * @returns The filtered state after applying the semantics.
+ * Interface that needs to be implemented by any {@link AbstractInterpretationVisitor} that applies interval condition
+ * semantics.
  */
-type UnaryOpSemantics = (argNodeId: NodeId, state: StateAbstractDomain<IntervalDomain> | undefined, visitor: NumericIntervalInferenceVisitor, dfg: DataflowGraph) => StateAbstractDomain<IntervalDomain> | undefined;
+export interface IntervalValueDomainAccess<StateDomain extends AnyStateDomain<AnyAbstractDomain>> {
```
OliverGerstl: Minor: This name is a bit confusing to me, maybe something like `IntervalInference` or so?

```suggestion
export interface IntervalInference<StateDomain extends AnyStateDomain<AnyAbstractDomain>> {
```

test/functionality/abstract-interpretation/interval/interval.ts
OliverGerstl: That was a suggestion from me to clarify that `DomainMatchingType.Equal` checks for equality rather than, for example, whether the lower bound equals the upper bound (the term “Exact” was a bit misleading here).
I also changed this in #2475

src/abstract-interpretation/domains/mapped-abstract-domain.ts
OliverGerstl, line 1: This file should have been removed :)

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -1,177 +1,97 @@
...
+	return [
+		[Identifier.make('=='), binaryConditionSemanticsGuard(intervalEqualsOp), binaryConditionSemanticsGuard(intervalNotEqualsOp)],
+		[Identifier.make('!='), binaryConditionSemanticsGuard(intervalNotEqualsOp), binaryConditionSemanticsGuard(intervalEqualsOp)],
+		[Identifier.make('>'), binaryConditionSemanticsGuard(intervalGreaterOp), binaryConditionSemanticsGuard(intervalLessEqualOp)],
+		[Identifier.make('>='), binaryConditionSemanticsGuard(intervalGreaterEqualOp), binaryConditionSemanticsGuard(intervalLessOp)],
+		[Identifier.make('<'), binaryConditionSemanticsGuard(intervalLessOp), binaryConditionSemanticsGuard(intervalGreaterEqualOp)],
+		[Identifier.make('<='), binaryConditionSemanticsGuard(intervalLessEqualOp), binaryConditionSemanticsGuard(intervalGreaterOp)],
+		[Identifier.make('is.na'), unaryConditionSemanticsGuard(intervalIsNaFn), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
+}
```
OliverGerstl: Idea: Maybe define this as arrow function to highlight that this is const?

```suggestion
const IntervalSemanticsMapper = <StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends IntervalConditionSemanticsVisitor<StateDomain>>() => [
	[Identifier.make('=='), binaryConditionSemanticsGuard(intervalEqualsOp), binaryConditionSemanticsGuard(intervalNotEqualsOp)],
	[Identifier.make('!='), binaryConditionSemanticsGuard(intervalNotEqualsOp), binaryConditionSemanticsGuard(intervalEqualsOp)],
	[Identifier.make('>'), binaryConditionSemanticsGuard(intervalGreaterOp), binaryConditionSemanticsGuard(intervalLessEqualOp)],
	[Identifier.make('>='), binaryConditionSemanticsGuard(intervalGreaterEqualOp), binaryConditionSemanticsGuard(intervalLessOp)],
	[Identifier.make('<'), binaryConditionSemanticsGuard(intervalLessOp), binaryConditionSemanticsGuard(intervalGreaterEqualOp)],
	[Identifier.make('<='), binaryConditionSemanticsGuard(intervalLessEqualOp), binaryConditionSemanticsGuard(intervalGreaterOp)],
	[Identifier.make('is.na'), unaryConditionSemanticsGuard(intervalIsNaFn), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
```

src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -1,177 +1,97 @@
...
+
+function onUnknownPositiveFunctionCall<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends IntervalConditionSemanticsVisitor<StateDomain>>(argNodeId: NodeId | undefined, state: StateDomain, visitor: Visitor) {
```
OliverGerstl: Minor: Maybe `applyUnknownPositiveCondition` and `applyUnknownNegativeCondition` or so, since this changes the abstract state?

src/abstract-interpretation/absint-condition-semantics.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Returns a callback, that extracts exactly one argument and calls the provided unary condition semantics function.
+ * If more or less than one argument is provided, no semantics are applied and a warning is logged.
+ * @param unaryConditionSemantics - The condition semantics to call.
+ */
+export function unaryConditionSemanticsGuard<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends AbstractInterpretationVisitor<StateDomain>>(
+	unaryConditionSemantics: UnaryConditionSemantics<StateDomain, Visitor>
+): NAryConditionSemantics<StateDomain, Visitor> {
+	return (argNodeIds: readonly (NodeId | undefined)[], state: StateDomain, visitor: Visitor, dfg: DataflowGraph) => {
+		if(argNodeIds.length !== 1 || isUndefined(argNodeIds[0])) {
+			numericInferenceLogger.warn('Called unary condition operator with more/less than 1 argument or with undefined argument.');
```
OliverGerstl: Maybe you shouldn't use the `numericInferenceLogger` here if this is domain-/analysis-independent :)

src/abstract-interpretation/interval/eval.ts
```diff
@@ -45,7 +46,15 @@ async function instrument(folder: string, workingDir: string, outputFolder: stri
 		ctx:           analyzer.inspectContext()
 	});
 
+	const pentagonVisitor = new NumericPentagonInferenceVisitor({
+		normalizedAst: await analyzer.normalize(),
+		dfg:           (await analyzer.dataflow()).graph,
+		controlFlow:   await analyzer.controlflow(),
+		ctx:           analyzer.inspectContext()
+	});
```
OliverGerstl: Maybe you can just construct a common visitor config?
```ts
	const visitorConfig = {
		normalizedAst: await analyzer.normalize(),
		dfg:           (await analyzer.dataflow()).graph,
		controlFlow:   await analyzer.controlflow(),
		ctx:           analyzer.inspectContext()
	};
	const visitor = new NumericIntervalInferenceVisitor(visitorConfig);
	const pentagonVisitor = new NumericPentagonInferenceVisitor(visitorConfig);
```

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -38,20 +41,20 @@ type BinaryOpSemantics = (left: IntervalDomain | undefined, right: IntervalDomai
 /**
  * Semantics definition function for unary functions.
  * @param arg - The raw argument of the function.
- * @param visitor - The numeric inference visitor performing the analysis used to resolve the argument interval.
+ * @param getInterval - A function to retrieve the current inferred interval for the provided node id.
  * @param significantFigures - The number of significant figures used to create new intervals.
  * @returns The resulting interval after applying the semantics.
  */
-type UnaryFnSemantics = (arg: FunctionArgument, visitor: NumericIntervalInferenceVisitor, significantFigures: number | undefined) => IntervalDomain | undefined;
+type UnaryFnSemantics = (arg: FunctionArgument, getInterval: (node: NodeId) => IntervalDomain | undefined, significantFigures: number | undefined) => IntervalDomain | undefined;
```
OliverGerstl: I thought you added the `IntervalValueDomainAccess` to get rid of the `getInterval`/`setInterval` functions?

src/abstract-interpretation/interval/numeric-interval-inference.ts
```diff
@@ -62,15 +63,33 @@ export class NumericIntervalInferenceVisitor extends AbstractInterpretationVisit
 		return this.currentState.set(call.id, result);
 	}
 
-	protected override applyConditionSemantics(state: StateAbstractDomain<IntervalDomain> | undefined, conditionNodeId: NodeId, trueBranch: boolean): StateAbstractDomain<IntervalDomain> | undefined {
+	setInterval(state: StateAbstractDomain<IntervalDomain>): (node: NodeId, value: (IntervalDomain | undefined)) => void {
+		return (node: NodeId, interval: IntervalDomain | undefined) => isUndefined(interval) ? state.remove(node) : state.set(node, interval);
+	}
```
OliverGerstl: I don't really understand this setter-like higher-order function.
Why can't this just be `setInterval(state: StateAbstractDomain<IntervalDomain>, node: NodeId, value: IntervalDomain | undefined): void`?

```suggestion
	setInterval(state: StateAbstractDomain<IntervalDomain>, node: NodeId, value: IntervalDomain | undefined): void {
		isUndefined(value) ? state.remove(node) : state.set(node, value);
	}
```

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	unaryIdentityConditionSemantics
+} from '../../absint-condition-semantics';
+import type { AbstractInterpretationVisitor } from '../../absint-visitor';
+
+/**
+ * Interface that needs to be implemented by any {@link AbstractInterpretationVisitor} that applies upper bounds
+ * condition semantics.
+ */
+export interface UpperBoundsDomainAccess<StateDomain extends AnyStateDomain<AnyAbstractDomain>> {
+	setUpperBounds(state: StateDomain): (node: NodeId, value: UpperBoundsValueDomain) => void;
```
OliverGerstl: Same as above: Why does this have to be a higher order function and not just `setUpperBounds(state: StateDomain, node: NodeId, value: UpperBoundsValueDomain): void`?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
```diff
@@ -0,0 +1,141 @@
...
+	return [
+		[Identifier.make('=='), binaryConditionSemanticsGuard(upperBoundsEqualsOp), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp)],
+		[Identifier.make('!='), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp), binaryConditionSemanticsGuard(upperBoundsEqualsOp)],
+		[Identifier.make('>'), binaryConditionSemanticsGuard(upperBoundsGreaterOp), binaryConditionSemanticsGuard(upperBoundsLessEqualOp)],
+		[Identifier.make('>='), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp), binaryConditionSemanticsGuard(upperBoundsLessOp)],
+		[Identifier.make('<'), binaryConditionSemanticsGuard(upperBoundsLessOp), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp)],
+		[Identifier.make('<='), binaryConditionSemanticsGuard(upperBoundsLessEqualOp), binaryConditionSemanticsGuard(upperBoundsGreaterOp)],
+		[Identifier.make('is.na'), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
+}
```
OliverGerstl: Maybe again as arrow function?
```suggestion
const UpperBoundsSemanticsMapper = <StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends UpperBoundsConditionSemanticsVisitor<StateDomain>>() => [
	[Identifier.make('=='), binaryConditionSemanticsGuard(upperBoundsEqualsOp), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp)],
	[Identifier.make('!='), binaryConditionSemanticsGuard(upperBoundsNotEqualsOp), binaryConditionSemanticsGuard(upperBoundsEqualsOp)],
	[Identifier.make('>'), binaryConditionSemanticsGuard(upperBoundsGreaterOp), binaryConditionSemanticsGuard(upperBoundsLessEqualOp)],
	[Identifier.make('>='), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp), binaryConditionSemanticsGuard(upperBoundsLessOp)],
	[Identifier.make('<'), binaryConditionSemanticsGuard(upperBoundsLessOp), binaryConditionSemanticsGuard(upperBoundsGreaterEqualOp)],
	[Identifier.make('<='), binaryConditionSemanticsGuard(upperBoundsLessEqualOp), binaryConditionSemanticsGuard(upperBoundsGreaterOp)],
	[Identifier.make('is.na'), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
] as const satisfies ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
```

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
```diff
@@ -0,0 +1,152 @@
...
+			return false;
+		}
+		return (other.value as UpperBoundsValue).isSubsetOf(this.value as UpperBoundsValue);
```
OliverGerstl: Minor: You can also use direct comparisons with `Bottom` here to omit the type casts :)

```suggestion
		if(this.value === Bottom) {
			return true;
		} else if(other.value === Bottom) {
			return false;
		}
		return other.value.isSubsetOf(this.value);
```

src/abstract-interpretation/pentagon/closed-pentagon-value-domain.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	public override top(significantFigures?: number): this & ProductDomain<AbstractClosedPentagon>;
+	public override top(significantFigures?: number): ClosedPentagonValueDomain {
+		return ClosedPentagonValueDomain.top(significantFigures);
+	}
+
+	public override bottom(significantFigures?: number): this & ProductDomain<AbstractClosedPentagon>;
+	public override bottom(significantFigures?: number): ClosedPentagonValueDomain {
+		return ClosedPentagonValueDomain.bottom(significantFigures);
+	}
```
OliverGerstl: I think `top` and `bottom` should just carry on the `significantFigures` from the current value here :)

```suggestion
	public override top(): this & ProductDomain<AbstractClosedPentagon>;
	public override top(): ClosedPentagonValueDomain {
		return ClosedPentagonValueDomain.top(this.value.interval.significantFigures);
	}

	public override bottom(): this & ProductDomain<AbstractClosedPentagon>;
	public override bottom(): ClosedPentagonValueDomain {
		return ClosedPentagonValueDomain.bottom(this.value.interval.significantFigures);
	}
```

src/abstract-interpretation/pentagon/expression-semantics.ts
```diff
@@ -0,0 +1,270 @@
...
+			numericInferenceLogger.warn('Called unary function with more/less than 1 argument, which is not supported.');
+			return ClosedPentagonValueDomain.bottom(significantFigures);
+		}
+
+		return unaryFunctionSemantics(target, args[0], visitor, currentState, significantFigures);
+	};
+}
+
+function pentagonUnaryIdentityOp(_target: NodeId, arg: [NodeId, ClosedPentagonValueDomain | undefined]): ClosedPentagonValueDomain | undefined {
+	return arg[1];
```
OliverGerstl: Why `arg[1]`? Is this a bug? Otherwise, please add a short comment to explain this :)

src/abstract-interpretation/pentagon/numeric-pentagon-inference.ts
```diff
@@ -0,0 +1,179 @@
...
+
+			const sourcePentagon = this.currentState.get(sourceOrigin);
+			const targetPentagon = this.currentState.get(target);
+
+			if(sourcePentagon?.isValue() && targetPentagon?.isValue()) {
+				sourcePentagon.value.upperBounds.add(target);
+				targetPentagon.value.upperBounds.add(sourceOrigin);
+				// To every upper-bounds that contains source, also add target: a small reduction step to increase precision.
+				if(this.currentState.isValue()) {
+					this.currentState.value.forEach(closedPentagonValue => {
```
OliverGerstl: A for loop is most of the time more performant than calling `forEach` :)

src/abstract-interpretation/pentagon/numeric-pentagon-inference.ts
```diff
@@ -0,0 +1,179 @@
...
+		return (node: NodeId, value: UpperBoundsValueDomain) => {
+			const pentagon = state.get(node);
+			if(isUndefined(pentagon)) {
+				// As we currently cannot describe that we have upper bounds-info but not know whether it is a numeric scalar value, we cannot infer upper-bounds values for non-numeric scalar values.
+				return;
+			}
+			pentagon.value.upperBounds = value;
+			state.set(node, pentagon);
+		};
+	}
```
OliverGerstl: Same as above: Why the higher-order functions and not just `setInterval(state: ClosedPentagonDomain, node: NodeId, value: IntervalDomain | undefined): void {...}` and `setUpperBounds(state: ClosedPentagonDomain, node: NodeId, value: UpperBoundsValueDomain): void {...}`, respectively? :)

OliverGerstl (approved):

src/abstract-interpretation/interval/eval.ts
schuler-henry: Da die left side eh keinen Origin haben sollte ist das aktuell hardcore unnötig, deshlab lohnt es sich auch nicht die location mit dazu zu packen.

EagleoutIce: ok fair, aber ich dachte jetzt auch an nen general purpose dump :) aber non-urgent!

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: Wie genau meinst du das? Wir können ja nur sicher wissen, dass das Ergebnis eine Zahl ist wenn das argument eine Zahl ist, oder? Oder bezieht sich das auf typescript?

EagleoutIce: Naja das weißt du ja nich generell weil es ja auch ein S3 overload ist, aber du kannst theoretisch infos von anderen domänen einfließen lassen weil man mit "+" dinge converted. Das war mehr eine mention dass wir da wegen Future work mal drüber reden können; Sorry das hätte ich klarer machen können

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: Zumindest für die Condition hab ich das ja schon in den Visitor rein gezogen über ein interface. Wahrscheinlich ist es sinnvoll das dann analog auch für die expression semantics zu benutzen. Ich schau dass ich das noch mit umsetz, dass das konsequent ist.

EagleoutIce: tyy

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: Ah, jetzt kapier ich auch was FW bedeutet....

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: Not supported in terms of R throws an error if no argument (meaning an empty argument) is provided to a unary operator. So an error means the program will stop so I guess bottom should be fine here.

src/abstract-interpretation/interval/numeric-interval-inference.ts
schuler-henry: Ja, das stimmt. Ich würde das als improvement für Iteration 3 mal notieren.

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
schuler-henry: What do you mean with that?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
schuler-henry: Explain that to the type system

src/abstract-interpretation/absint-condition-semantics.ts
schuler-henry: Please elaborate.

src/abstract-interpretation/interval/expression-semantics.ts
EagleoutIce: But it does not have to, just the default implementations do that :see_no_evil: (and technically for some older ones the parser still requires there to be either one or to). But I think this is fine for us.

test/functionality/abstract-interpretation/pentagon/interim-presentation-example.test.ts
schuler-henry: We could do that but we do not have many use cases for that yet, so I do not think that it is so relevant yet.

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
EagleoutIce: u potentially invoke `visitor.setUpperBounds(state)` twice

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
EagleoutIce: That is more an issue with `has` then :P

src/abstract-interpretation/absint-condition-semantics.ts
EagleoutIce: You take the state, visistor, and dfg as a batch for basically every semantic handler, so either you group them in an interface `argNodeIds: ..., ctx: { state.... }` or you spread the types into the arguments (similar to a `fn(...Parameters<typeof anotherFn>)`

test/functionality/abstract-interpretation/interval/interval.ts
EagleoutIce: Hmm Okay, I am still not quite convinced whether we want a semantic equal and not a structural equal (which was my understanding of the `exact` option) but this is nothing that henry has to suffer through :D

test/functionality/abstract-interpretation/pentagon/interim-presentation-example.test.ts
EagleoutIce: ok!

src/abstract-interpretation/absint-condition-semantics.ts
EagleoutIce: oh jeah, good catch, I agree :D
Additionally, a lot of the error messaging/general handling is so similar to other N*ary fn handlers so maybe we can outsource this as well?

src/abstract-interpretation/pentagon/numeric-pentagon-inference.ts
EagleoutIce: Jeah, although newer versions of v8 are watering this down (sadly, in the wrong direction by also slowing down for), so I'd argue the reason to change this is more for consistency than guaranteed perf benefits.

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
schuler-henry: @OliverGerstl in how far should we reuse the SetDomain semantics, or is that something that you will do after i completed my thesis?

src/abstract-interpretation/domains/mapped-abstract-domain.ts
schuler-henry: Yeah, this is what is happening in this pr (I accidentally comitted an empty file during the merge so this should fix that).

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: Jokes on me, should be fixed now 😇

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
schuler-henry: Is this the correct solution to the fact that typescript cannot infer this using the info provided by `isBottom()`? Seems a bit odd to have an isBottom function but then not use it... @EagleoutIce

src/abstract-interpretation/pentagon/expression-semantics.ts
schuler-henry: Well we simply carry on the closed pentagon from the argument and since we provide the nodeid + pentagon as a list to the function, we need to return the pentagon part which is at 1.

schuler-henry: But if you want to we could use destruction of the array in the fn signature and simply return the pentagon part, so that it is more obvious?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
EagleoutIce: `isBottom` needs a type assertion for it to work without the `=== Bottom`, I've already mentioned that I am not a fan of the current type f..ery going on with the AbsInt `is*` guards :P (looking at you `this is this` na na na nana)

src/abstract-interpretation/interval/eval.ts
```diff
@@ -156,7 +169,7 @@ const folder = process.argv[2];
 const outputFolder = process.argv[3];
 const script = process.argv[4];
 
-void instrument(folder, path.join('/tmp', folder), outputFolder).then(() => execute(path.join('/tmp', folder), script)).catch(err => {
+instrument(folder, path.join('/tmp', folder), outputFolder).then(() => execute(path.join('/tmp', folder), script)).catch(err => {
```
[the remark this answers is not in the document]
schuler-henry: As the eval script will not be part of the final main merge, this is irrelevant

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
OliverGerstl: We can do it however you like. In any case, it would make sense later on to implement the upper bounds domain as a reverse set domain (also, so that the maximum number of inferred bounds is limited). But whether you want to do that or I do it later is up to you (you can also delegate this to a new issue for now) :)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
OliverGerstl: This was just a suggestion based on your implementation of the other domain operators where you used comparisons with `Bottom`.
Another solution would be to use `isValue()`. I can also introduce a `isNotBottom` function :)
```ts
public leq(other: this): boolean {
	if (this.isValue() && other.isValue()) {
		return other.value.isSubsetOf(this.value);
	}
	return this.isBottom();
}
```

src/abstract-interpretation/pentagon/expression-semantics.ts
OliverGerstl: Ah ok, I thought that `arg` is an array, sorry xD
Yes, I think destructuring the tuple in the parameter list would make this clearer :)

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
schuler-henry: If I get that correctly, with the change from higher order function to setUpperBounds(state, node, value) this is no longer an issue?!?

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-value-domain.ts
schuler-henry: Okay, so I will note that for the third iteration as a potential will do.

src/abstract-interpretation/pentagon/upper-bounds/upper-bounds-condition-semantics.ts
EagleoutIce: probably not

src/abstract-interpretation/interval/eval.ts
schuler-henry: I think that is rather minor as this is only for eval purpose and not a real contribution to flowr

test/functionality/abstract-interpretation/pentagon/closed-pentagon-domain.test.ts
schuler-henry: yes, they are

src/abstract-interpretation/pentagon/closed-pentagon-value-domain.ts
```diff
@@ -0,0 +1,48 @@
+import { ProductDomain } from '../domains/product-domain';
+import { UpperBoundsValueDomain } from './upper-bounds/upper-bounds-value-domain';
+import { IntervalDomain } from '../domains/interval-domain';
+
+/** The type of the abstract product representing the closed pentagon */
+export type AbstractClosedPentagon = {
+	interval:    IntervalDomain;
+	upperBounds: UpperBoundsValueDomain;
+};
```
OliverGerstl, line 9: You could make `interval` and `upperBounds` explicitly readonly here to enforce immutability :)
(You can also just do this in the next iteration)

OliverGerstl (approved):
Looks good to me, can be merged :D

src/abstract-interpretation/pentagon/closed-pentagon-value-domain.ts
schuler-henry: Lets do this in the next iteration and save this fun for later 😄

PR #2485, opened 2026-05-15
[AbsInt] Improve implementation of abstract domains

### LINKED ISSUE #2484: [AbsInt] Improve implementation of abstract domains
- Remove the `concretize` and `abstract` method of abstract domains
- Provide a `from` function for abstract value domains to abstract concrete values
- Make implementation of domain operators more consistent
- Introduce interfaces for value domains (i.e., numeric domains, string domains, boolean domains, set domains, ...)
- Provide default handling for top and bottom cases of abstract domains

OliverGerstl:


src/abstract-interpretation/data-frame/dataframe-domain.ts
```diff
@@ -4,18 +4,17 @@ import { SetRangeDomain } from '../domains/set-range-domain';
 
 /** The type of the abstract product representing the shape of data frames */
 export type AbstractDataFrameShape = {
```
EagleoutIce, line 6: why not have it as an interface?

src/abstract-interpretation/domains/interval-domain.ts
```diff
@@ -38,149 +39,114 @@ export class IntervalDomain<Value extends IntervalLift = IntervalLift>
...
+			this.upper === +Infinity ? other.upper : this.upper
+		]);
 	}
 
 	public satisfies(value: number, comparator: NumericalComparator = NumericalComparator.Equal): Ternary {
 		switch(comparator) {
 			case NumericalComparator.Equal: {
-				if(this.isValue() && this.value[0] <= value && value <= this.value[1]) {
-					return this.value[0] === this.value[1] ? Ternary.Always : Ternary.Maybe;
+				if(this.isValue() && this.lower <= value && value <= this.upper) {
```
EagleoutIce, line 120: so we consistently consider top to be a value but bottom to be no value?

src/abstract-interpretation/domains/value-abstract-domain.ts
```diff
@@ -0,0 +1,140 @@
...
+	Less,
+	LessOrEqual,
+	Greater,
+	GreaterOrEqual
+}
+
+/**
+ * Represents the different types of set comparators for satisfiability checks for an abstract domain.
+ */
+export enum SetComparator {
```
EagleoutIce, line 136: maybe also an interest of Intersect and NoIntersect? Especially if we want to blacklis some results?

EagleoutIce (commented):
Just a quick pass for now, i have not yet checked whether all changes are applied consistently but the gneeral change looks clear :)

src/abstract-interpretation/data-frame/dataframe-domain.ts
OliverGerstl: Unfortunately, an interface does not work here

src/abstract-interpretation/domains/interval-domain.ts
OliverGerstl: Yes, but you can override this any time in child classes if you want a different behavior :)

src/abstract-interpretation/data-frame/dataframe-domain.ts
EagleoutIce: i fail to understand why :see_no_evil: because some of the subtypes are unsafe?

src/abstract-interpretation/domains/interval-domain.ts
EagleoutIce: as long as this is documented i am fine :D

src/abstract-interpretation/data-frame/dataframe-domain.ts
OliverGerstl: This seems to be a known "issue" in TypeScript
https://github.com/microsoft/TypeScript/issues/42825
https://github.com/microsoft/TypeScript/issues/15300#issuecomment-332366024
The suggested fix is to just use types instead of interfaces here

EagleoutIce: Lol okay xD

EagleoutIce:
_This pull request is included in v2.13.0 (see [Release v2.13.0 (Guessing Dep. Versions, Specialization, RDA)](https://github.com/flowr-analysis/flowr/releases/tag/v2.13.0))._

PR #2488, opened 2026-05-18
[CFG]: Support for origin 

gigalasr:
very work in progress


EagleoutIce:
with the new integration of the cfg into the dtaflow we no longer need this!!

PR #2522, opened 2026-06-01
2481 interval pentagon ai increase num of function semantics

schuler-henry:
Added coarse semantics for the top 50 (or so) functions and operators, to check whether more semantics increase the number of inferred values in the ground truth evaluation.

### DIFF
--- a/src/abstract-interpretation/interval/condition-semantics.ts
+++ b/src/abstract-interpretation/interval/condition-semantics.ts
@@ -37,7 +37,11 @@ export const IntervalSemanticsMaper = <StateDomain extends AnyStateDomain<AnyAbs
 	[Identifier.make('is.na'), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
 	[Identifier.make('is.null'), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
 	[Identifier.make('is.character'), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	[Identifier.make('is.factor'), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
 	[Identifier.make('is.numeric'), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval)],
+	[Identifier.make('is.nan'), unaryConditionSemanticsGuard(intervalBottomIfArgIsInterval), unaryConditionSemanticsGuard(unaryIdentityConditionSemantics)],
+	[Identifier.make('is.infinite'), unaryConditionSemanticsGuard(intervalIsInfinite), unaryConditionSemanticsGuard(intervalIsFinite)],
+	[Identifier.make('is.finite'), unaryConditionSemanticsGuard(intervalIsFinite), unaryConditionSemanticsGuard(intervalIsInfinite)]
 ] as const satisfies readonly ConditionSemanticsMapperInfo<StateDomain, Visitor>[];
 
 function applyUnknownPositiveCondition<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends IntervalConditionSemanticsVisitor<StateDomain>>(argNodeId: NodeId | undefined, state: StateDomain, visitor: Visitor) {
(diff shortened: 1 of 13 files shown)


src/abstract-interpretation/interval/condition-semantics.ts
```diff
@@ -201,12 +208,65 @@ function intervalLessEqualOp<StateDomain extends AnyStateDomain<AnyAbstractDomai
...
+
+function intervalIsInfinite<StateDomain extends AnyStateDomain<AnyAbstractDomain>, Visitor extends IntervalConditionSemanticsVisitor<StateDomain>>(argNodeId: NodeId, state: StateDomain, visitor: Visitor): StateDomain {
+	const argValue = visitor.getInterval(argNodeId, state);
+
+	// Only if we can resolve the argument to an interval, and it has no infinit bounds we can safely return bottom.
+	// Otherwise, the condition might be satisfiable.
+	if(isNotUndefined(argValue)) {
+		if(argValue.isValue()) {
+			const [a, b] = argValue.value;
+			if(Number.isFinite(a) && Number.isFinite(b)) {
```
OliverGerstl, line 229: Minor: You could also use `IntervalDomain::isFinite` :)

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -324,4 +386,466 @@ function intervalLengthFn<StateDomain extends AnyStateDomain<AnyAbstractDomain>>
...
+	let rmNa = false;
+	if(isNotUndefined(rmNaArg)) {
+		if(isNotUndefined(rmNaArg.valueId)) {
+			// Resolve the value to "TRUE" or "FALSE"
+			const valueNode = dfg.idMap?.get(rmNaArg.valueId);
+			if(RLogical.is(valueNode) && RLogical.isTrue(valueNode)) {
+				rmNa = true;
+			}
+		}
+	}
```
OliverGerstl, line 423: For resolving values, you should use `resolveIdToValue` in `src/dataflow/eval/resolve/alias-tracking.ts` (or `resolveIdToArgValue` in `src/abstract-interpretation/data-frame/resolve-args.ts`)

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -324,4 +386,466 @@ function intervalLengthFn<StateDomain extends AnyStateDomain<AnyAbstractDomain>>
...
+
+	const allArgumentIntervals = args
+		.filter(arg => FunctionArgument.isNamed(arg) && arg.name !== 'na.rm' || FunctionArgument.isPositional(arg))
+		.map(arg => visitor.getInterval((arg as NamedFunctionArgument | PositionalFunctionArgument).nodeId));
```
OliverGerstl: Maybe outsource these repeating patterns into helper functions?

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -324,4 +386,466 @@ function intervalLengthFn<StateDomain extends AnyStateDomain<AnyAbstractDomain>>
...
+	}
+
+	const [a, b] = operand.value as [number, number];
+
+	if(a < 0) {
+		return undefined;
+	}
+	if(a == b) {
+		if(isUndefined(base)) {
+			return operand.scalar(Math.log(a));
```
OliverGerstl, line 607: I think these are cases where you should check how the significant figures are affected by the log?

src/abstract-interpretation/interval/expression-semantics.ts
```diff
@@ -324,4 +386,466 @@ function intervalLengthFn<StateDomain extends AnyStateDomain<AnyAbstractDomain>>
...
+	return IntervalDomain.top();
+}
+
+function intervalRound<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(args: readonly FunctionArgument[], visitor: IntervalExpressionSemanticsVisitor<StateDomain>, significantFigures: number | undefined): IntervalDomain | undefined {
+	if(args.length === 0 || args.length > 2 || args.some(FunctionArgument.isEmpty)) {
+		return IntervalDomain.bottom(significantFigures);
+	}
+
+	let x: IntervalDomain | undefined | 'unknown' = 'unknown';
+	let digits: IntervalDomain | undefined | 'unknown' = 'unknown';
```
OliverGerstl, line 649: Maybe you can define helper functions for retrieving values for arguments, similar to the helpers in `src/abstract-interpretation/data-frame/mappers/arguments.ts`, as you need to resolve arguments quite often?

test/functionality/abstract-interpretation/interval/interval.ts
```diff
@@ -59,7 +59,8 @@ export const IntervalTests = {
 
 export enum DomainMatchingType {
 	Equal = 'equal',
-	Overapproximation = 'overapproximation'
+	Overapproximation = 'overapproximation',
+	Underapproximation = 'underapproximation'
```
OliverGerstl: I don't understand the `Underapproximation` option here.
Do we have test where we want the inference to be an under-approximation?
Shouldn't these cases be "assert fails" tests (e.g., with `{ fails: true }`) to make explicit that something is not supported yet?

OliverGerstl (approved):

src/abstract-interpretation/interval/condition-semantics.ts
schuler-henry: Wenn ich das nutzen würde, sagt der mir, dass argValue nach dem if type 'never' ist...

EagleoutIce: gotta love that absint type safety :P

src/abstract-interpretation/interval/expression-semantics.ts
schuler-henry: As discussed will note that as future work.

test/functionality/abstract-interpretation/interval/interval.ts
schuler-henry: Mhm the only way we can currently support nrow without using knowledge from the df-shape inference would be to assume that nrow always returns a positive number (although it might also return null if the provided element is not a valid datastructure that has nrow defined on it). So this is currently a thing where we underapproximate, hence the extra option as it is a known underapproximation.

The other option (which is the safe way) would be that we can only infer 'undefined' stating that we don't know if nrow produces a number, if you (both) think this is better, we can switch back to being perfectly sound 🤷🏻 .

(Of course there is still the assumption with S3/S4 dispatching ... - i know, but thats nothing were gonna support anytime soon)

Impl of nrow for reference:
```ts
function intervalNrowNcolFn<StateDomain extends AnyStateDomain<AnyAbstractDomain>>(arg: FunctionArgument, visitor: IntervalExpressionSemanticsVisitor<StateDomain>, significantFigures: number | undefined): IntervalDomain | undefined {
	if(FunctionArgument.isEmpty(arg)) {
		numericInferenceLogger.warn('Called unary "nrow/ncol" with an empty argument, which is not supported.');
		return IntervalDomain.bottom(significantFigures);
	}

	const inferredInterval = visitor.getInterval(arg.nodeId);
	if(isNotUndefined(inferredInterval)) {
		if(inferredInterval.isValue()) {
			// nrow/ncol cannot process scalar values and therefore returns null
			return undefined;
		}
		if(inferredInterval.isBottom()) {
			return IntervalDomain.bottom(significantFigures);
		}
	}

	numericInferenceLogger.warn('Applying the nrow/ncol semantics can possibly lead to an underapproximation, as we neglect the fact that they can return null.');
	return new IntervalDomain([0, Infinity], significantFigures);
}
```

EagleoutIce: I am not sure I understand both sides, but Id argue that we use Equal, etc. to signal which values are also *correct* but potentially not as precise. So I'd also prefer to have no Underapproximation option (indicating that being wrong here is something we accept) and reuse the existing system also used with the DFA etc. to mark tests in which we currently *know* that we are (slightly) unsound.

OliverGerstl: Yes, it is okay that there are some cases where we are just soundy and may under-approximate (and we should definitely document them with tests).
But I would not add an "Underapproximatition" option in the tests but rather flag them as "this test should fail, as we are not sound in this special case" with assert-fails tests via the option `{ fails: true }`

schuler-henry: I have removed the Underapproximation option and just added tests with the fails flag as discussed.

PR #2535, opened 2026-06-03
feat: added benchmarking script

schuler-henry:
@OliverGerstl @EagleoutIce schaut mal ob ich die Daten so richtig raus zieh aus flowr und ob die festgehaltenen Daten so sinnvoll/ausreichend sind.
Falls ihr noch punkte habt die man ergänzen kann, gerne schreiben.

In Olis Arbeit hab ich gesehen, dass er die einzelnen base-flowr steps separat gemessen hat, also dfg, cfg, ast, ...., wäre das hier auch sinnvoll? muss ich dafür im Zweifel was spezielles beachten wie die Reihenfolge in der ich das mach (weil sonst evtl. vorherige Graphen bereits erzeugt werden)?


src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+interface InferredValue {
+	nodeId:             NodeId;
+	inferredValue:      string;
+	significantFigures: string;
+	sourceLocation:     string;
+}
+
+void async function() {
+	fs.mkdirSync(outputDirectory, { recursive: true });
+	const fileName = filePath.split('/').reverse()[0];
```
EagleoutIce: what is going on here?? please use the getBasename or whatever their name is funcion

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+import { SourceLocation } from '../util/range';
+import { NumericPentagonInferenceVisitor } from './pentagon/numeric-pentagon-inference';
+import path from 'path';
+
+if(process.argv.length < 4) {
+	console.error('Usage: ts-node src/abstract-interpretation/benchmarking.ts <file-to-analyze> <output-folder>');
+	process.exit(1);
+}
+
+const filePath = process.argv[2];
```
EagleoutIce: you could also pattern match to make all of them clear
```ts
const [node,prg,filepath,outDir] = process.argv
```

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	nodeId:             NodeId;
+	inferredValue:      string;
+	significantFigures: string;
+	sourceLocation:     string;
+}
+
+void async function() {
+	fs.mkdirSync(outputDirectory, { recursive: true });
+	const fileName = filePath.split('/').reverse()[0];
+	const metadataOutputFile = fileName.slice(0, -2) + '.csv';
```
EagleoutIce: again, a command that cuts of the extension like.. say.. path.parse would be appreciated for readability (<3)

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const intervalResultFile = fileName.slice(0, -2) + '-interval.csv';
+	const pentagonResultFile = fileName.slice(0, -2) + '-pentagon.csv';
+
+	const baselineStart = performance.now();
+	const analyzer = await new FlowrAnalyzerBuilder()
```
EagleoutIce, line 81: you could disable some plugins if you like but you do not have to. 
Likewise we could add a plugini that dos some instrumentation to your analysis

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+		.setEngine('tree-sitter')
+		.build();
+	analyzer.addRequest(fileProtocol + filePath);
+	const dfg = (await analyzer.dataflow()).graph;
+	const ast = await analyzer.normalize();
+
+	const visitorContext = {
+		normalizedAst: ast,
+		dfg:           dfg,
+		controlFlow:   await analyzer.controlflow(),
```
EagleoutIce: so dfg and ast are initialized prior, but cfg and context are set inline?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const intervalResults: InferredValue[] = [];
+	for(const nodeId of intervalVisitor.getAbstractTrace().keys()) {
+		const value = intervalVisitor.getAbstractValue(nodeId);
+		if(isNotUndefined(value)) {
+			const node = ast.idMap.get(nodeId);
+			const location = node ? SourceLocation.fromNode(node) : undefined;
+
+			intervalResults.push({
+				nodeId:             nodeId,
+				inferredValue:      value?.toString() ?? 'undefined',
```
EagleoutIce: why `value?` if you assert it is not undefined?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+				significantFigures: value?.significantFigures?.toString() ?? 'unknown',
+				sourceLocation:     isUndefined(location) ? 'unknown' : SourceLocation.getRange(location).slice(0, 4).toString(),
```
EagleoutIce: why slice for 0,4 if you get the range? you could also use `??` to have `SourceLocation.getRange(location) ?? 'unknown'`, also, why do you use `SourceLocation.fromNode` and not just `SourceRange.fromNode` which safes you the get (and why drop filenames :tear:)

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const intervalResultStart = performance.now();
+	const intervalResults: InferredValue[] = [];
+	for(const nodeId of intervalVisitor.getAbstractTrace().keys()) {
+		const value = intervalVisitor.getAbstractValue(nodeId);
+		if(isNotUndefined(value)) {
+			const node = ast.idMap.get(nodeId);
+			const location = node ? SourceLocation.fromNode(node) : undefined;
+
+			intervalResults.push({
+				nodeId:             nodeId,
```
EagleoutIce, line 112: you could just write `nodeId,`

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	const intervalStart = performance.now();
+	intervalVisitor.start();
+	const intervalEnd = performance.now();
+
+	const intervalResultStart = performance.now();
+	const intervalResults: InferredValue[] = [];
+	for(const nodeId of intervalVisitor.getAbstractTrace().keys()) {
+		const value = intervalVisitor.getAbstractValue(nodeId);
+		if(isNotUndefined(value)) {
```
EagleoutIce: I am not a fan of these functions for anything besides iterator fn compliance (due to their broken type contracts). So I'd prefer something like `if(value !== undefined)` for readability. Anything who has another opinion is wrong (:side-eye:).

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+			const node = ast.idMap.get(nodeId);
+			const location = node ? SourceLocation.fromNode(node) : undefined;
```
EagleoutIce: fromNode also takes in `undefined` (at least on main) so it should be possible to elicitate the ternary if

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+				sourceLocation:     isUndefined(location) ? 'unknown' : SourceLocation.getRange(location).slice(0, 4).toString(),
+			});
+		}
+	}
+	const intervalResultEnd = performance.now();
+
+	// Dump Interval Results
+	const intervalPath = path.join(outputDirectory, intervalResultFile);
+	if(fs.existsSync(intervalPath)) {
+		fs.rmSync(intervalPath, { recursive: true, force: true });
```
EagleoutIce: given that force does not fail, why even the exist check?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+	fs.writeFileSync(intervalPath, ['nodeId', 'inferredValue', 'significantFigures', 'sourceLocation'].join(',') + '\n');
+	for(const result of intervalResults) {
+		fs.appendFileSync(intervalPath, Object.values(result).map(value => '"' + value + '"').join(',') + '\n');
+	}
```
EagleoutIce: generally I'd strinigfy the value to avoid breaking the csv escaping 

btw: why use `toString` to get the value contents instead of just dumping json right away, this makes it easier to parse the values again, safes you on operations, and I'd argue simplifies the eval.
small performance: moving the loop into the string build and then calling `appendFileSync` only once.... your disk will thank you

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const pentagonResultStart = performance.now();
+	const pentagonResults: InferredValue[] = [];
+	for(const nodeId of pentagonVisitor.getAbstractTrace().keys()) {
+		const value = pentagonVisitor.getAbstractValue(nodeId);
+		if(isNotUndefined(value)) {
+			const node = ast.idMap.get(nodeId);
+			const location = node ? SourceLocation.fromNode(node) : undefined;
+
+			pentagonResults.push({
+				nodeId:             nodeId,
```
EagleoutIce, line 143: i mean the node id is fine for matching and the source location is also fine to find wherever it happened, but theoretically you could also dump the lexeme to get more information in case of a failure (and also dataflow type etc.)

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+				inferredValue:      value?.toString() ?? 'undefined',
+				significantFigures: value?.value.interval.significantFigures?.toString() ?? 'unknown',
+				sourceLocation:     isUndefined(location) ? 'unknown' : SourceLocation.getRange(location).slice(0, 4).toString(),
```
EagleoutIce: repeated pattern, maybe outsource the dump logic?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const pentagonResultEnd = performance.now();
+
+	// Dump Pentagon Results
+	const pentagonPath = path.join(outputDirectory, pentagonResultFile);
+	if(fs.existsSync(pentagonPath)) {
+		fs.rmSync(pentagonPath, { recursive: true, force: true });
+	}
+	fs.writeFileSync(pentagonPath, ['nodeId', 'inferredValue', 'significantFigures', 'sourceLocation'].join(',') + '\n');
+	for(const result of pentagonResults) {
+		fs.appendFileSync(pentagonPath, Object.values(result).map(value => '"' + value + '"').join(',') + '\n');
```
EagleoutIce: please outsource the logic for the writing these duplications are hurting me

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// Create File and Runtime Metadata
+	const metadataStart = performance.now();
+	const constants = dfg.verticesOfType(VertexType.Value).toArray();
+	const numberConstants = constants.filter(value => ast.idMap.get(value[0])?.type === RType.Number);
```
EagleoutIce, line 162: if you filter on the ast type you have to be more careful wrt. symbols that resolve to numeric constants (which you miss this way), like `NA` etc. or `pi`

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+		pentagonAnalysisInMs:                   pentagonEnd - pentagonStart,
+		pentagonResultGatheringInMs:            pentagonResultEnd - pentagonResultStart
+	};
+
+	// Dump File and Runtime Metadata
+	const fileMetadataPath = path.join(outputDirectory, metadataOutputFile);
+	if(fs.existsSync(fileMetadataPath)) {
+		fs.rmSync(fileMetadataPath, { recursive: true, force: true });
+	}
+	fs.writeFileSync(fileMetadataPath, Object.keys(fileMetadata).join(',') + '\n' +
```
EagleoutIce: write should just overwrite? no?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	const functionCalls = dfg.verticesOfType(VertexType.FunctionCall).toArray();
+	const supportedExpressionFunctionCalls = functionCalls.filter(([_, dfgCall]) =>
+		isNotUndefined(IntervalExpressionSemanticsMapper().find(([id]) => Identifier.matches(id, dfgCall.name))) ||
+		isNotUndefined(PentagonExpressionSemanticsMapper.find(([id]) => Identifier.matches(id, dfgCall.name)))
+	);
+	const supportedConditionFunctionCalls = functionCalls.filter(([_, dfgCall]) =>
+		isNotUndefined(IntervalSemanticsMaper().find(([id]) => Identifier.matches(id, dfgCall.name))) ||
+		isNotUndefined(UpperBoundsSemanticsMapper().find(([id]) => Identifier.matches(id, dfgCall.name))) ||
+		['&&', '!', '||'].some(id => Identifier.matches(id, dfgCall.name))
```
EagleoutIce: ouch. but fine :P

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+		loc:                                    analyzer.inspectContext().files.getAllFiles()[0].content().toString().split('\n').length,
+		numOfConstants:                         constants.length,
+		numOfRNumberConstants:                  numberConstants.length,
+		RNumberConstantNodeIds:                 numberConstants.map(([node]) => node),
+		numOfFunctionCalls:                     functionCalls.length,
+		numOfSupportedExpressionFunctionCalls:  supportedExpressionFunctionCalls.length,
+		SupportedExpressionFunctionCallNodeIds: supportedExpressionFunctionCalls.map(([node]) => node),
+		numOfSupportedConditionFunctionCalls:   supportedConditionFunctionCalls.length,
+		SupportedConditionFunctionCalls:        supportedConditionFunctionCalls.map(([node]) => node),
+		numOfConditions:                        conditions.length,
```
EagleoutIce, line 206: also numOfUnknown side effects etc. and why not count ifs and whiles separate? especially for `while` i'd argue the chance for dead code should be lower as 1) `if(FALSE)` is the more frequent pattern and we are weaker in tracking monotony in whiles.

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+		metadataGatheringInMs:                  metadataEnd - metadataStart,
```
EagleoutIce, line 212: how about size of total graph, number of control flow nodes....?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+		baselineInMs:                           baselineEnd - baselineStart,
+		intervalAnalysisInMs:                   intervalEnd - intervalStart,
+		intervalResultGatheringInMs:            intervalResultEnd - intervalResultStart,
```
EagleoutIce, line 218: you could also rely on `IStoppableStopwatch` for this similar to the benchmark slicer to make this more readable
+ why not also store reported dfg, cfg and normalize timings?

src/abstract-interpretation/benchmarking.ts
EagleoutIce: write should just overwrite? no?

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	const supportedConditionFunctionCalls = functionCalls.filter(([_, dfgCall]) =>
+		isNotUndefined(IntervalSemanticsMaper().find(([id]) => Identifier.matches(id, dfgCall.name))) ||
+		isNotUndefined(UpperBoundsSemanticsMapper().find(([id]) => Identifier.matches(id, dfgCall.name))) ||
+		['&&', '!', '||'].some(id => Identifier.matches(id, dfgCall.name))
+	);
+	const conditions = functionCalls.filter(([_, dfgCall]) =>
+		dfgCall.origin.includes(BuiltInProcName.IfThenElse) || dfgCall.origin.includes(BuiltInProcName.WhileLoop)
+	);//.map(([node, _]) => cfg.graph.ingoingEdges(node)?.keys().toArray()[0]);
+	const metadataEnd = performance.now();
+
```
EagleoutIce, line 188: why not also use the dead code analysis flowR has so far to give meaningful insights about the influence of bottom findings?

src/abstract-interpretation/benchmarking.ts
EagleoutIce: ouch. but fine :P

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,198 @@
...
+	for(const result of pentagonResults) {
+		fs.appendFileSync(pentagonPath, Object.values(result).map(value => '"' + value + '"').join(',') + '\n');
+	}
+
+	// Create File and Runtime Metadata
+	const metadataStart = performance.now();
+	const constants = dfg.verticesOfType(VertexType.Value).toArray();
+	const numberConstants = constants.filter(value => ast.idMap.get(value[0])?.type === RType.Number);
+
+	const functionCalls = dfg.verticesOfType(VertexType.FunctionCall).toArray();
```
EagleoutIce, line 164: pleaes note that this includes clalls within function definitions. generally why not also count fdefs to have indications for unsupported parts?

EagleoutIce (changes requested):

src/abstract-interpretation/benchmarking.ts
schuler-henry: Luckily I used these functions everywhere haha lol 🤷🏻‍♂️ I don't mind not using them though

EagleoutIce: You are wrong
\- Oliver Gerstl

EagleoutIce: (ich kann das nur schreiben weil bis er in Github bei dem Internet antworten kann bestimmt Jahre vergehen)

OliverGerstl: Ey, ich finde `isUndefined` und `isNotUndefined` **SCHRECKLICH**, Flo hat Recht und ist ein super Mensch. Diese Nachricht ist sicher von mir, Olivar Gestl.

schuler-henry: At this point würde ich einfach wild mischen nach lust und laune 😇

src/abstract-interpretation/benchmarking.ts
schuler-henry: What woud be ur suggested solution?

src/abstract-interpretation/benchmarking.ts
schuler-henry: In that case just json dump everything rather than csv I guess, so just one JSON.stringify() at the end without the need for any loops whatsoever...?

src/abstract-interpretation/benchmarking.ts
EagleoutIce: you could resolve the value with alias tracking and then check on the unpacked value in case of symbols etc. but generally I'd argue this is fine 1) you do not support pi etc. yourself right now 2) the other counts are very low 3) you do not care about NA(_integer_) do you?

It was more a comment of something that should be noted wrt. statements made on the result.

src/abstract-interpretation/benchmarking.ts
EagleoutIce: I'd say that is better! but you may need bigStringify if the output gets too big

src/abstract-interpretation/benchmarking.ts
schuler-henry: Okay 😄

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,227 @@
...
+interface InferredValue {
+	nodeId:             NodeId;
+	inferredValue:      string;
+	significantFigures: string;
+	sourceLocation:     string;
+	lexeme:             string;
+	dfgTag:             VertexType | 'unknown';
+}
+
+void async function() {
```
OliverGerstl, line 72: I think this function would be more readable if you split parts of it into separate functions :)

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,227 @@
...
+		intervalResultGatheringInMs:            intervalResultEnd - intervalResultStart,
+		pentagonAnalysisInMs:                   pentagonEnd - pentagonStart,
+		pentagonResultGatheringInMs:            pentagonResultEnd - pentagonResultStart
+	};
+
+	// Dump File and Runtime Metadata
+	const fileMetadataPath = path.join(outputDirectory, metadataOutputFile);
+	fs.rmSync(fileMetadataPath, { recursive: true, force: true });
+	fs.writeFileSync(fileMetadataPath, JSON.stringify(fileMetadata, null, 2));
+}();
```
OliverGerstl, line 227: Also, for comprehensibility, giving this function a name (e.g. `main` or so) and calling it by name would be better

src/abstract-interpretation/benchmarking.ts
```diff
@@ -0,0 +1,227 @@
...
+	fs.rmSync(pentagonPath, { recursive: true, force: true });
+	superBigJsonStringify(pentagonResults, '\n', (msg) => fs.appendFileSync(pentagonPath, msg));
+
+	// Create File and Runtime Metadata
+	const metadataStart = performance.now();
+	const constants = dfg.verticesOfType(VertexType.Value).toArray();
+	const numberConstants = constants.filter(value => ast.idMap.get(value[0])?.type === RType.Number);
+
+	const functionCalls = dfg.verticesOfType(VertexType.FunctionCall).toArray();
+	const supportedExpressionFunctionCalls = functionCalls.filter(([_, dfgCall]) =>
```
OliverGerstl, line 165: Maybe it would also be interesting to filter these supported function calls for those calls for which you inferred anything?

OliverGerstl (approved):

EagleoutIce:
We will do the clean up on the staging branch

PR #2536, opened 2026-06-05
Add rudimentary violation detection

### LINKED ISSUE #2502: [taint-analysis] Violation Detection
For tainting problems for which we know certain (possible) violations we want to detect these and surface them in the API.

MaxAtoms:
Closes #2502

### DIFF
--- a/src/queries/catalog/taint-query/taint-query-executor.ts
+++ b/src/queries/catalog/taint-query/taint-query-executor.ts
@@ -1,14 +1,14 @@
 import { log } from '../../../util/log';
 import type { BasicQueryData } from '../../base-query-format';
 import type { TaintQuery, TaintQueryResult } from './taint-query-format';
-import type {
-	AllPredefinedTaintAnalyses,
-} from '../../../taint-analysis/builder/taint-analysis';
+import type { AllPredefinedTaintAnalysisNames } from '../../../taint-analysis/predefined/predefined';
+
+
 
 /**
  * Executes the given taint queries using the provided analyzer.
  */
-export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: readonly TaintQuery[]): Promise<TaintQueryResult> {
+export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: readonly TaintQuery[]): Promise<TaintQueryResult<AllPredefinedTaintAnalysisNames>> {
 	const flattened = queries.flatMap(q => q.defs);
 
 	if(flattened.length == 0) {
@@ -16,18 +16,14 @@ export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: r
 	}
 
 	const start = Date.now();
-	const analysis = analyzer.taint() as unknown as AllPredefinedTaintAnalyses;
+
+	const analysis = analyzer.taint<AllPredefinedTaintAnalysisNames>();
 	for(const def of flattened) {
 		analysis.addPredefined(def);
 	}
 
-	const visitors = await analysis.run();
-	const results = new Map(
-		Array.from(visitors, ([k, r]) => [k, r])
-	);
-
 	return {
-		results: results,
+		results: await analysis.run(),
 		'.meta': {
 			timing: Date.now() - start
 		},
--- a/src/queries/catalog/taint-query/taint-query-executor.ts
+++ b/src/queries/catalog/taint-query/taint-query-executor.ts
@@ -1,6 +1,9 @@
 import { log } from '../../../util/log';
 import type { BasicQueryData } from '../../base-query-format';
 import type { TaintQuery, TaintQueryResult } from './taint-query-format';
+import type {
+	AllPredefinedTaintAnalyses,
+} from '../../../taint-analysis/builder/taint-analysis';
 
 /**
  * Executes the given taint queries using the provided analyzer.
@@ -13,14 +16,14 @@ export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: r
 	}
 
 	const start = Date.now();
-	const analysis = analyzer.taint();
+	const analysis = analyzer.taint() as unknown as AllPredefinedTaintAnalyses;
 	for(const def of flattened) {
 		analysis.addPredefined(def);
 	}
 
 	const visitors = await analysis.run();
 	const results = new Map(
-		Array.from(visitors, ([k, v]) => [k, v.getEndState()])
+		Array.from(visitors, ([k, r]) => [k, r])
 	);
 
 	return {
--- a/src/queries/catalog/taint-query/taint-query-format.ts
+++ b/src/queries/catalog/taint-query/taint-query-format.ts
@@ -4,7 +4,9 @@ import type { FlowrConfig } from '../../../config';
 import type { ParsedQueryLine, QueryResults, SupportedQuery } from '../../query';
 import Joi from 'joi';
 import { executeTaintQuery } from './taint-query-executor';
-import type { AnyPredefinedAnalysisName } from '../../../taint-analysis/predefined/predefined';
+import type {
+	AnyPredefinedTaintAnalysisName
+} from '../../../taint-analysis/predefined/predefined';
 import { predefinedTaintAnalyses } from '../../../taint-analysis/predefined/predefined';
 import type { AnyAbstractDomain } from '../../../abstract-interpretation/domains/abstract-domain';
 import { Bottom, BottomSymbol } from '../../../abstract-interpretation/domains/lattice';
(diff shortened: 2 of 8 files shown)


src/queries/catalog/taint-query/taint-query-executor.ts
```diff
@@ -13,14 +16,14 @@ export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: r
 	}
 
 	const start = Date.now();
-	const analysis = analyzer.taint();
+	const analysis = analyzer.taint() as unknown as AllPredefinedTaintAnalyses;
```
EagleoutIce: I mean maybe we could at least hide this cast in an explicit generic?

`.taint<AllPredefinedTaint...>` + this type could be considered the default initializer?

src/queries/catalog/taint-query/taint-query-executor.ts
```diff
@@ -13,14 +16,14 @@ export async function executeTaintQuery({ analyzer }: BasicQueryData, queries: r
...
 	for(const def of flattened) {
 		analysis.addPredefined(def);
 	}
 
 	const visitors = await analysis.run();
 	const results = new Map(
-		Array.from(visitors, ([k, v]) => [k, v.getEndState()])
+		Array.from(visitors, ([k, r]) => [k, r])
```
EagleoutIce: is this mapper still required then? :D this feels idempotent

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -4,24 +4,24 @@ import type { FlowrConfig } from '../../../config';
...
 import type { CommandCompletions } from '../../../cli/repl/core';
 import { fileProtocol } from '../../../r-bridge/retriever';
-import type { AnyStateDomain } from '../../../abstract-interpretation/domains/state-domain-like';
 import type { StateDomainLift } from '../../../abstract-interpretation/domains/state-abstract-domain';
+import type { TaintInferenceResult } from '../../../taint-analysis/builder/taint-analysis';
 
 export interface TaintQuery extends BaseQueryFormat {
 	readonly type: 'taint';
-	readonly defs: PredefinedTaintAnalysis[]
+	readonly defs: AnyPredefinedAnalysisName[]
```
EagleoutIce: my main quarrel is with the act that now the `Taint` term is no longer part of the type which might make this confusing with all the other analysis passes that we support?

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -4,24 +4,24 @@ import type { FlowrConfig } from '../../../config';
...
 }
 
 export interface TaintQueryResult extends BaseQueryResult {
-	readonly results: Map<string, AnyStateDomain>
+	readonly results: Map<string, TaintInferenceResult>
```
EagleoutIce: I'd say the taint result could also be a generic based on the input taint analyses; but this is not critical

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -57,21 +57,21 @@ function taintQueryCompleter(line: readonly string[], startingNewArg: boolean, _
 	return { completions: [] };
 }
 
-function defsInInput(defsPart: readonly string[]): { valid: (PredefinedTaintAnalysis)[], invalid: string[] } {
+function defsInInput(defsPart: readonly string[]): { valid: (AnyPredefinedAnalysisName)[], invalid: string[] } {
```
EagleoutIce:
```suggestion
function defsInInput(defsPart: readonly string[]): { valid: AnyPredefinedAnalysisName[], invalid: string[] } {
```

Again I'd argue that you could also just use a generic for this :D this way you could also enforce that the invalids are a type subset of the `defsPart` - but again not critical for now :D just an API consideration.

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -57,21 +57,21 @@ function taintQueryCompleter(line: readonly string[], startingNewArg: boolean, _
...
 	return defsPart
 		.reduce((acc, name) => {
 			name = name.trim();
 			if(name in predefinedTaintAnalyses) {
```
EagleoutIce, line 70: this could also be a Object.hasOwn... check to avoid anyone requesting a predefined analysis lie `toString` or any other property that may be attached to the object.

src/queries/catalog/taint-query/taint-query-format.ts
EagleoutIce: on another note I'd like to have a small interface for the valid; invalid typing as this would avoid the cast duplication in the reduce init.

src/taint-analysis/builder/taint-analysis-definition.ts
```diff
@@ -11,7 +11,7 @@ export class TaintAnalysisDefinition<Name extends string, Domain extends AnyAbst
 	public mapper:          TaintMapper<Domain> = [];
 	public name:            Name;
 
-	private msg: string | undefined;
+	public msg: string | undefined;
```
EagleoutIce: if public, should it be readonly?

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -90,25 +90,30 @@ function taintQueryLineParser(output: ReplOutput, line: readonly string[], _conf
...
-			const lift = domains.value as StateDomainLift<AnyAbstractDomain>;
 			if(lift === Bottom) {
-				result.push(`      ╰ state: ${BottomSymbol}`);
+				resultStrings.push(`      ╰ state: ${BottomSymbol}`);
 				return true;
 			}
 
-			result.push(...lift.entries().take(20).map(([key, domain]) => {
+			resultStrings.push(...lift.entries().take(20).map(([key, domain]) => {
 				return `      ╰ ${key}: ${domain?.toString()}`;
```
EagleoutIce: theoretically you can write these one-liners more easily without braces, also saving on the `return`

src/queries/catalog/taint-query/taint-query-format.ts
```diff
@@ -90,25 +90,30 @@ function taintQueryLineParser(output: ReplOutput, line: readonly string[], _conf
...
+				resultStrings.push(`      ╰ finding: ${result.finding}`);
+			}
 
-		for(const [name, domains] of state) {
-			result.push(`   ╰ **${name}**:`);
-			const lift = domains.value as StateDomainLift<AnyAbstractDomain>;
 			if(lift === Bottom) {
-				result.push(`      ╰ state: ${BottomSymbol}`);
+				resultStrings.push(`      ╰ state: ${BottomSymbol}`);
 				return true;
```
EagleoutIce, line 114: isnt this implying that you'd like the default asccii summarizer to kick in?

src/taint-analysis/predefined/predefined.ts
```diff
@@ -5,7 +5,7 @@ export const predefinedTaintAnalyses = {
 	'scale': scaleAnalysis,
 } as const satisfies AnalysisMap<['scale']>;
 
-export type PredefinedTaintAnalysis = keyof typeof predefinedTaintAnalyses;
+export type AnyPredefinedAnalysisName = keyof typeof predefinedTaintAnalyses;
```
EagleoutIce: as mentioned above, I'd like some kind of taint branding on the type name

src/taint-analysis/predefined/scale-analysis.ts
```diff
@@ -49,4 +49,4 @@ export const scaleAnalysis = new TaintAnalysisDefinition('scale', scaleDomain)
 			}
 		}
 	}])
-	.report('Warning: Mean of scaled value is always zero');
+	.report('Mean of scaled value is always zero');
```
EagleoutIce, line 52: We can also add report severities later if oyu like, but for this, I'd hook into the existing linting system and avoid creating a new format. But so far this is fine :D 
one thing we could consider (in the far future/outside the thesis) would be string interpolation/report functions that could also use trace information/the specific function name etc.

EagleoutIce (commented):
So far so nice :) Do you want to hook the reports into the linting system? or what is your plan wrt this? generally we could also close it of here as well and just dump the findings list at the end for the benchmarking.

src/queries/catalog/taint-query/taint-query-executor.ts
MaxAtoms: Good point, I actually considered it before. I now added a type parameter for the analysis names, but set the empty array as default so that it is not necessary to explicitly use the type parameter in something like `.taint<[]>().add(new TaintAnalysis('myAnalysis'))`

src/taint-analysis/builder/taint-analysis-definition.ts
MaxAtoms: I added a proper getter

src/queries/catalog/taint-query/taint-query-format.ts
MaxAtoms: Which default summarizer do you mean? Where do I find it? 🙈

src/taint-analysis/predefined/scale-analysis.ts
MaxAtoms: Once I have more useful rules, I would like to plug the taint analysis into the linter system and also adopt the severity levels, if necessary. But first, I want to focus on evaluation/benchmarking 😅

src/queries/catalog/taint-query/taint-query-format.ts
EagleoutIce: That one:
https://github.com/flowr-analysis/flowr/blob/548298dce8f4ebbfe7689fc7025b8251dc43ab9c/src/queries/query-print.ts#L111

but true is actually the behavior I wanted :D

https://github.com/flowr-analysis/flowr/blob/548298dce8f4ebbfe7689fc7025b8251dc43ab9c/src/queries/query.ts#L145-L149

src/taint-analysis/predefined/scale-analysis.ts
EagleoutIce: No worries that works :)

src/queries/catalog/taint-query/taint-query-format.ts
MaxAtoms: As discussed, I will leave this as is for now. We might want to extract common logic shared with the linter query at some point

PR #2541, opened 2026-06-09
Fix `MultiValueDomain` reductions

### LINKED ISSUE #2540: [Bug]: Internal bug in MultiValueDomain - Reductions accessed before being set
### What happened?

Currently, the interplay of `MultiValueDomain` and `PartialProductDomain` is broken.
The constructor of `MultiValueDomain` accesses the super constructor before setting the reductions, leading to `TypeError: Cannot read properties of undefined (reading 'reduce')`. Executing the example code from the Wiki shows this behavior: https://github.com/flowr-analysis/flowr/blob/1c742d8fee153185755715ccb1a33d2cb4aad5aa/src/documentation/wiki-absint.ts#L88

_No response_

v2.10.7 (Jun 3, 2026)

### Relevant log output

```Shell

```

### Meta Information

- [ ] The issue is urgent

MaxAtoms:
Lift the default reductions processing from the `MultiValueDomain` to the `PartialProductDomain`.

@OliverGerstl Feel free to give feedback or edit the PR yourself as you see fit :)
Not sure whether we want to keep the new test cases and their structure, because they deviate from the domain tests we currently have.

The fact that we have two ways to define reductions (via either the constructor parameter or overriding the `reduce` function) adds complexity, but I see why we would want to have both.

Closes #2540 

_Disclaimer: Identified and fixed using OpenCode and Claude Opus 4.8, with some manual minor clean-up by hand_


OliverGerstl (approved):
Thank you for identifying and fixing the issue.
The solution is fine for me, but I am also not so happy about the generated tests.
I think we don't need them for now, as we will probably have more meaningful tests as soon as the multi-value domain is actually used :)

EagleoutIce:
_This pull request is included in v2.10.8 (see [Release v2.10.8 (MultiValue fixes, refined input sources, fixed linter rules)](https://github.com/flowr-analysis/flowr/releases/tag/v2.10.8))._

PR #2544, opened 2026-06-09
2495 interval pentagon ai improve pentagon ram usage

schuler-henry:
Contribution for Iteration 3.

Applied performance optimizations as described in the pentagon paper, where we do not explicitly store all implicit constraints but rather adapt the join and other operators to make these implicit constraints explicit if required.

This reduces the amount of required ram as well as the number of operations (and therefore runtime) significantly, but also possibly looses some information. However, since the interim eval showed that we did not infer more precise intervals with the pentagon domain compared to the interval domain, this might not be too much of a loss.

Additionally added `&` and `|` with the same abstract semantics as `&&` and `||` which should be fine for the usage with numeric scalar values.


OliverGerstl (approved):
Looks good to me :)

PR #2554, opened 2026-06-13
refactor: linter rule refactoring and improvement

### LINKED ISSUE #2550: [Linter] Refinements
* absolute file path should not trigger if you pass an URL to read.csv etc.
* file not found linter should not trigger if given a url instead of a file path (maybe check the url if web is enabled)

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.10.8 (see [Release v2.10.8 (MultiValue fixes, refined input sources, fixed linter rules)](https://github.com/flowr-analysis/flowr/releases/tag/v2.10.8))._

Issue #1823, opened 2025-07-20
Add list of unsupported data frame function to config


OliverGerstl:
The list of unsupported data frame functions that are mapped to top is currently only in the code. We want to add the list to the flowR config to allow users to add more data frame functions or to test the inference without some data frame functions.

PR #2573, opened 2026-06-17
Fix small abstract interpretation issues on variable definitions, assignments, unsupported functions

### LINKED ISSUE #2523: Abstract Interpreation fixpoint iteration is skipped early when processing function with unknown side effect
The code below describes the problem and suggests a potential fix.
The provided tests are described for the interval implementation based on commit [2cf2c6f](https://github.com/flowr-analysis/flowr/tree/2cf2c6f).
To reproduce them using a different domain, the tests need to be adapted, keeping the calls to functions with unknown side effects (assign, paste, load).

```
// Problem 1: Fixpoint iteration is not applied or at least only for the part after the call with unknown side effect.
// Problem 2: onVariableDefinition asserts that a variable has not been assigned before if it is not present in the current state.
//            However, as the assigned variable might be set to top in the current state (being undefined), we have to check the trace instead.
//            This way, the assigned variable does not appear as potential orign in the second loop iteration for the case that x <- x + 1.
// Problem 3: Neither assign nor paste are declared as function with unknown side effects which might cause underapproximations.

// Potential fix for 1: We have to check whether any variable in the loop body has changed to assess whether we need
//                      to re-run the loop, so that not only the state after the loop is a correct overapproximation
//                      but also all states within the loop are a correct overapproximation.
//                      Potentially it might be sufficient to just run the loop one more time instead or alternatively
//                      only check those states right before the function call with unknown side effect as they should
//                      contain all changed information of the predecessors
// Fix for 2: Simply replace "this.currentState.get(vertex.id) === undefined" with "this.trace.get(vertex.id) === undefined"

// Iterates only one time over the loop as load() clears the trace (unknown sideeffect),
// so the old and new state are the same (empty) after the first iteration
// Therefore, underapproximates to [1, 1] in l.3 when it should overapproximate to Top
// (or be [1, 4] if load does not override x).
testIntervalDomain(`
	x <- 0
	while (x <= 3) {
		x <- x + 1
		load("test.rda")
	}
`, {
	'3@x': { domain: IntervalTests.interval(1, 4), matching: DomainMatchingType.Overapproximation },
});

// Iterates two times over the loop as the y <- 1 assignment introduces new information after the load(),
// which stabilizes after one additional iteration.
// Additionally, the origin of the right x in l.3 does not resolve to the left x in l.3 in the second iteration,
// as the onVariableDefinition labels the left x as unassigned in the second iteration as the state has been
// cleared, although it has been assigned in the iteration before
// (this can be fixed by checking the trace instead of the currentState in the onVariableDefinition function).
// Without the fix, underapproximates to [1, 1] when it should overapproximate to Top
// (or be [1, 4] if load does not override x).
// With the fix, overapproximate to Top in the second iteration.
testIntervalDomain(`
	x <- 0
	while (x <= 3) {
		x <- x + 1
		load("test.rda")
		y <- 1
	}
`, {
	'3@x': { domain: IntervalTests.interval(1, 4), matching: DomainMatchingType.Overapproximation },
});

// Due to the paste call in the assign, flowr cannot resolve that x is reassigned to -1.
// The absint framework does not label assign nor paste as function with unknown side effect.
// Consequently, the framework ignores this assignment and infers [1, 4] for l.3 when it should be [0, 1]:
// 1 from the first iteration where x <- 0 + 1 and 0 from the following iterations where x <- -1 + 1.
testIntervalDomain(`
	x <- 0
	while (x <= 3) {
		x <- x + 1
		assign(paste("x"), -1)
	}
`, {
	'3@x': { domain: IntervalTests.interval(0, 1) },
});
```

### LINKED ISSUE #2542: Abstract Interpretation underapproximation due to non-explicit unknown variable information for joining states
When we evaluate a condition (in this case x==1) we can only keep the info x==1 if we update all origins, so that they (for intervals) have the inferred value [1, 1].
For the else branch we would do the same with x!=1 - in this case we cannot give more precise info than the `undefined` (meaning no information) which was provided by the assignment in the first line (as Hallo is a string and not a number)

When the states of the if and else branch are joined at the end, the abstract state in the if branch maps the origin of x to [1, 1], while the abstract state of the else branch maps the origin of x to `undefined`.

Currently we assume that `undefined` means that the variable is simply not present in the branch (as it does not exist there) and throw away de `undefined` and keep the [1, 1] value.
However, this is wrong as the variable exists in both branches and has just an unknown inferred value in the else branch, so we would expect the result to be `undefined` for x after the if/else rather than [1, 1].

```ts
x <- "Hallo"

if(x == 1) {
	x
} else {
	x
}
x
```

OliverGerstl:
This PR provides fixes to:
- Check trace for unassigned variables at variable definitions instead of current state (#2523)
- Allow setting variables to `undefined` (Top) in assignments (#2542)
- Functions with unknown environment side effects that are partially supported, such as `assign` (#2523)

The issue with early terminated fixpoint iterations due to unsupported functions at the end of loops is not fixed, but a test case is added to document the issue

Closes #2523, closes #2542

### DIFF
--- a/src/abstract-interpretation/unsupported-functions.ts
+++ b/src/abstract-interpretation/unsupported-functions.ts
@@ -1,10 +1,19 @@
 import { Identifier } from '../dataflow/environments/identifier';
-import { type DataflowGraphVertexArgument, isFunctionCallVertex } from '../dataflow/graph/vertex';
+import type { DataflowGraph } from '../dataflow/graph/graph';
+import { type DataflowGraphVertexArgument, type DataflowGraphVertexFunctionCall, isFunctionCallVertex } from '../dataflow/graph/vertex';
+
+/**
+ * Function info for unsupported functions that may change the environment unresolvable/implicitly.
+ */
+interface UnsupportedFunctionInfo {
+	package:    string;
+	condition?: (vertex: DataflowGraphVertexFunctionCall, dfg: DataflowGraph) => boolean;
+}
 
 /**
  * List of known function calls that may change the environment unresolvable/implicitly.
  */
-const UnsupportedFunctionsList = {
+const UnsupportedFunctionsList: Record<string, UnsupportedFunctionInfo> = {
 	'.Primitive':          { package: 'base' },
 	'.Internal':           { package: 'base' },
 	'.External':           { package: 'base' },
@@ -26,18 +35,25 @@ const UnsupportedFunctionsList = {
 	'rm':                  { package: 'base' },
 	'remove':              { package: 'base' },
 	'list2env':            { package: 'base' },
+	'assign':              { package: 'base', condition: (vertex, dfg) => dfg.unknownSideEffects.has(vertex.id) },
+	'delayedAssign':       { package: 'base', condition: (vertex, dfg) => dfg.unknownSideEffects.has(vertex.id) },
 	'assignInNamespace':   { package: 'utils' },
 	'assignInMyNamespace': { package: 'utils' },
-} as const satisfies Record<string, { package: string }>;
+};
 
 /**
  * Helper for unsupported functions that may change the environment.
  */
 export const UnsupportedFunctions = {
 	/**
-	 * Checks whether a data flow graph vertex represents a unsupported (environment-changing) function call (e.g. `eval`, `load`, `attach`, `rm`, ...)
+	 * Checks whether a data flow graph vertex represents an unsupported (environment-changing) function call (e.g. `eval`, `load`, `attach`, `rm`, ...)
 	 */
-	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined): boolean {
-		return isFunctionCallVertex(vertex) && Object.hasOwn(UnsupportedFunctionsList, Identifier.getName(vertex.name));
+	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined, dfg: DataflowGraph): boolean {
+		if(!isFunctionCallVertex(vertex)) {
+			return false;
+		}
+		const identifier = Identifier.getName(vertex.name);
+
+		return Object.hasOwn(UnsupportedFunctionsList, identifier) && (UnsupportedFunctionsList[identifier].condition?.(vertex, dfg) ?? true);
 	}
 };
--- a/src/abstract-interpretation/unsupported-functions.ts
+++ b/src/abstract-interpretation/unsupported-functions.ts
@@ -1,45 +1,50 @@
 import { Identifier } from '../dataflow/environments/identifier';
 import type { DataflowGraph } from '../dataflow/graph/graph';
 import { type DataflowGraphVertexArgument, type DataflowGraphVertexFunctionCall, isFunctionCallVertex } from '../dataflow/graph/vertex';
+import { Record } from '../util/record';
 
-/**
- * Function info for unsupported functions that may change the environment unresolvable/implicitly.
- */
 interface UnsupportedFunctionInfo {
-	package:    string;
-	condition?: (vertex: DataflowGraphVertexFunctionCall, dfg: DataflowGraph) => boolean;
+	readonly condition?: (vertex: DataflowGraphVertexFunctionCall, dfg: DataflowGraph) => boolean;
+}
+
+interface UnsupportedFunctionEntry extends UnsupportedFunctionInfo {
+	readonly identifier: Identifier;
+}
+
+function unsupportedFunctions(functions: Record<`${string}::${string}`, UnsupportedFunctionInfo>): readonly UnsupportedFunctionEntry[] {
+	return Record.entries(functions).map(([identifier, info]) => ({ identifier: Identifier.parse(identifier), ...info }));
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/abstract-interpretation/unsupported-functions.ts
```diff
@@ -1,10 +1,19 @@
...
+interface UnsupportedFunctionInfo {
+	package:    string;
+	condition?: (vertex: DataflowGraphVertexFunctionCall, dfg: DataflowGraph) => boolean;
+}
 
 /**
  * List of known function calls that may change the environment unresolvable/implicitly.
  */
-const UnsupportedFunctionsList = {
+const UnsupportedFunctionsList: Record<string, UnsupportedFunctionInfo> = {
```
EagleoutIce: in this setup it could also be a map so that it is easier to initialize it with more information / you could also move this into the flowR config into the futureu

src/abstract-interpretation/unsupported-functions.ts
```diff
@@ -26,18 +35,25 @@ const UnsupportedFunctionsList = {
...
 	 */
-	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined): boolean {
-		return isFunctionCallVertex(vertex) && Object.hasOwn(UnsupportedFunctionsList, Identifier.getName(vertex.name));
+	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined, dfg: DataflowGraph): boolean {
+		if(!isFunctionCallVertex(vertex)) {
+			return false;
+		}
+		const identifier = Identifier.getName(vertex.name);
+
+		return Object.hasOwn(UnsupportedFunctionsList, identifier) && (UnsupportedFunctionsList[identifier].condition?.(vertex, dfg) ?? true);
```
EagleoutIce: if its a map we could safe the hasOwn check just by adding another `?` but either is fine.

src/abstract-interpretation/unsupported-functions.ts
```diff
@@ -26,18 +35,25 @@ const UnsupportedFunctionsList = {
...
-	 * Checks whether a data flow graph vertex represents a unsupported (environment-changing) function call (e.g. `eval`, `load`, `attach`, `rm`, ...)
+	 * Checks whether a data flow graph vertex represents an unsupported (environment-changing) function call (e.g. `eval`, `load`, `attach`, `rm`, ...)
 	 */
-	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined): boolean {
-		return isFunctionCallVertex(vertex) && Object.hasOwn(UnsupportedFunctionsList, Identifier.getName(vertex.name));
+	isUnsupportedCall(this: void, vertex: DataflowGraphVertexArgument | undefined, dfg: DataflowGraph): boolean {
+		if(!isFunctionCallVertex(vertex)) {
+			return false;
+		}
+		const identifier = Identifier.getName(vertex.name);
```
EagleoutIce: shouldnt this also check the package/namespace attached? just to be sure? why otherwise would you include the pkg info

src/abstract-interpretation/unsupported-functions.ts
```diff
@@ -26,18 +35,25 @@ const UnsupportedFunctionsList = {
 	'rm':                  { package: 'base' },
 	'remove':              { package: 'base' },
 	'list2env':            { package: 'base' },
```
EagleoutIce: why not already supporting `base::rm` or something, flowR slowly moves to resolving these identifiers fully

EagleoutIce (approved):

src/abstract-interpretation/unsupported-functions.ts
OliverGerstl: I am not sure how a map would make it easier to provide more information here.
The plan is to add all this information to the flowR config in the future #1823

EagleoutIce: allright that is fine then, I just meant it is easier to set new values in a map then expecting users to create new properties (this also hurts the v8 optimizations)

src/abstract-interpretation/unsupported-functions.ts
OliverGerstl: This list collects all functions that may implicitly modify the environment. As long as we do not know all identifiers removed by `rm`, we have to treat `rm` as unknown function, over-approxinating the resulting abstract state by Top

EagleoutIce: no no sorry :see_no_evil: I meant that instead of making the mapping `"fn-name": { package: "package" }` which limits you to support potentially similarly named fns in multiple packages, why not directly map `base::rm`, `base::list2env` etc. as identifiers to search for

OliverGerstl: Ah ok, then I misunderstood you, sorry

OliverGerstl: The only issue is that most identifiers currently do not have a namespace, so I will keep the name-based record for now

EagleoutIce: I mean if you want but as soon as #2072 and its successor will be merged you'll have them for basically all

EagleoutIce:

PR #2577, opened 2026-06-18
feat-fix: safe exit protection

### LINKED ISSUE #2574: [Bug]: flowR REPL exits prematurely
### What happened?

When executing flowR within a container, long outputs might become truncated.

Suspected reason: When Node.js writes to a pipe instead of directly to a TTY, it does so asynchronously (https://nodejs.org/api/process.html#a-note-on-process-io). `flowr.js` exits via `process.exit(0)`, which does not wait for async write operations to finish.

> Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

_No response_

v2.10.9 (Jun 15, 2026)

### Relevant log output

```Shell

```

### Meta Information

- [ ] The issue is urgent
- [x] I used flowR as a docker image

EagleoutIce:


EagleoutIce:

PR #2586, opened 2026-06-20
refactor: improve performance of link resolve

### LINKED ISSUE #232: Improve the retrieval of function definitions
Currently `getAllLinkedFunctionDefinitions` is far from optimized and should be looked at. Especially if we *know* that no function by this name is currently defined - we do not even have to track (and save on a painfully expensive BFS).

https://github.com/Code-Inspect/flowr/blob/3ab77cc6836506eeb8da5bb23abaa652634d694d/src/dataflow/internal/linker.ts#L202-L240

EagleoutIce:


EagleoutIce:

PR #2591, opened 2026-06-25
feat-fix: conditional, cyclic read-edge linkage (dfg)

### LINKED ISSUE #2590: DFG: Missing reads edge
### What happened?

Missing reads edge from variable reference a in line 4 to variable definition a in line 4

In this example, a is incremented starting at 1 to infinity (endless loop, the condition inside always evaluates to true).
Consequently, the possible origins (reads edges) for the a in line 4 (`a + 1`) should be the definition of a in line 1 and the redefinition of a in line 4.
The DFG however only lists the definition of a in line 1 and the definition of a in line 6 as possible origins (reads edge).

> Note: the reads edge to the definition of a in line 6 is fine, as it is an overapproximation as flowR cannot resolve the condition `a > 0` without the info provided by for example the interval domain analysis

> Further, if the condition is unknown instead of `a > 0` there would also be a reads edge missing from variable reference a in line 6 to variable definition a in line 4.
> So, all in all if there is a condition at the end of the while loop, reads edges are only created to the else path (last occurrence of the variable) instead of both possible paths.

```ts
a <- 1
while (TRUE) {
  if (a > 0) {
    a <- a + 1
  } else {
    a <- a - 2
  }
}
```

---
flowR version: 2.10.6  
Extension version: 0.6.16 (Production mode)  
VS Code version: 1.123.0 (web false)  
Session: internal (active)  
OS: linux  

4.6.0

v2.10.6 (May 29, 2026)

### Relevant log output

```Shell

```

### Meta Information

- [ ] The issue is urgent

EagleoutIce:


EagleoutIce:

PR #2595, opened 2026-06-27
General abstract interpretation query

### LINKED ISSUE #2215: [AbsInt] Generalize the data frame shape query to an abstract interpretation query
The df-shape query should be generalized to an abstract interpretation query with configurable abstract interpretation options.
This may be inspired by the configuration format for abstract interpretation like in MOPSA.

OliverGerstl:


src/documentation/wiki-query.ts
```diff
@@ -539,29 +540,32 @@ enables quick statistics after each REPL command. Likewise, setting \`repl.dfPro
...
-	shortDescription: 'Returns the shapes inferred for all dataframes in the code.',
-	functionName:     executeDfShapeQuery.name,
-	functionFile:     '../queries/catalog/df-shape-query/df-shape-query-format.ts',
+	shortDescription: 'Returns the abstract values inferred for every expression or at specific locations.',
+	functionName:     executeAbsintQuery.name,
+	functionFile:     '../queries/catalog/absint-query/absint-query-format.ts',
 	buildExplanation: async(shell: RShell, ctx: GeneralDocContext) => {
-		const exampleCode = 'x <- data.frame(a=1:3)\nfilter(x, FALSE)';
-		const criterion = '2@x' as SlicingCriterion;
+		const criteria = ['1@df', '1@data.frame'] satisfies SlicingCriteria;
```
EagleoutIce: wouldnt as const also work here?

src/queries/catalog/absint-query/absint-query-executor.ts
```diff
@@ -0,0 +1,66 @@
...
+import { SlicingCriterion } from '../../../slicing/criterion/parse';
+import { log } from '../../../util/log';
+import type { BasicQueryData } from '../../base-query-format';
+import { AbsintQueryInferences, type AbsintQuery, type AbsintQueryDomain, type AbsintQueryResult, type AbsintQueryStateDomain, type AbsintQueryType } from './absint-query-format';
+
+/**
+ * Executes the given abstract interpretation queries using the provided analyzer.
+ */
+export async function executeAbsintQuery<AbsintType extends AbsintQueryType>({ analyzer }: BasicQueryData, queries: readonly AbsintQuery<AbsintType>[]): Promise<AbsintQueryResult<AbsintType>> {
+	if(queries.length === 0 || queries.some(query => query.inference !== queries[0].inference)) {
```
EagleoutIce, line 11: do you do the second part to make the return format easier?? because it feels weird to prevent someone asking the df-shapes and e.g. pentagons in the same query? why 1) make a named product or 2) use the same sub-key patterns that we e.g. use for the slicing requests?

src/queries/catalog/absint-query/absint-query-executor.ts
```diff
@@ -0,0 +1,66 @@
...
+		}
+		for(const criterion of query.criteria) {
+			if(result.has(criterion)) {
+				log.warn('Duplicate criterion in abstract interpretation query:', query.criteria);
+				continue;
+			}
+			try {
+				const nodeId = SlicingCriterion.parse(criterion, ast.idMap);
+				const node = ast.idMap.get(nodeId);
+				const value = visitor.getAbstractValue(node?.info.id);
```
EagleoutIce: why do you go over the id of the node? shouldnt this be nodeId directly?

src/queries/catalog/absint-query/absint-query-format.ts
```diff
@@ -0,0 +1,130 @@
...
+} satisfies Record<string, (config: AbsintVisitorConfiguration) => AbstractInterpretationVisitor<StateAbstractDomain<AnyAbstractDomain>>>;
+
+export type AbsintQueryType = keyof typeof AbsintQueryInferences;
+
+export type AbsintQueryStateDomain<AbsintType extends AbsintQueryType = AbsintQueryType> = DomainOfVisitor<ReturnType<typeof AbsintQueryInferences[AbsintType]>>;
+
+export type AbsintQueryDomain<AbsintType extends AbsintQueryType = AbsintQueryType> = ValueDomain<AbsintQueryStateDomain<AbsintType>>;
+
+
+function absintQueryCompleter(line: readonly string[], startingNewArg: boolean, _config: FlowrConfig): CommandCompletions {
```
EagleoutIce, line 42: I can check this in the REPL bu generally these completeres feel way to complex for these convenience features lol :P

src/queries/catalog/absint-query/absint-query-format.ts
```diff
@@ -0,0 +1,130 @@
...
+	const criteria = sliceCriteriaParser(line[1]);
+
+	if(criteria !== undefined && !criteria.every(SlicingCriterion.isValid)) {
+		output.stderr(output.formatter.format(`Invalid slicing criteria "${line[1]}"`, { color: Colors.Red, effect: ColorEffect.Foreground, style: FontStyles.Bold }));
+		return { query: [] };
+	}
+	const code = criteria ? line[2] : line[1];
+
+	if(code === undefined || code.length === 0) {
+		output.stderr(output.formatter.format('Missing R code to analyze', { color: Colors.Red, effect: ColorEffect.Foreground, style: FontStyles.Bold }));
```
EagleoutIce: why would you reject code.length === 0? I am not sure an empty file should beavre differently to, e.g. a file with a single space

src/documentation/wiki-query.ts
OliverGerstl: No, `as const` does not work here, but I am unsure why

src/queries/catalog/absint-query/absint-query-executor.ts
OliverGerstl: This is planned for the future, but as we currently cannot use the same visitor for multiple analyses, this is currently a restriction :)

src/queries/catalog/absint-query/absint-query-format.ts
OliverGerstl: I mean, if you have a shorter way feel free to change :P
But this is to complete the absint type, the slicing criteria (+ separators), and the file protocol or `""` :)

src/queries/catalog/absint-query/absint-query-format.ts
OliverGerstl: This does not affect file protocol case but only the case when providing the code as a string.
I want to prevent calling the query without a file nor code, e.g., `:query @absint df-shape`

src/queries/catalog/absint-query/absint-query-format.ts
EagleoutIce: nah it was just a general comment that a substential amount of the queries is the completer XD

src/queries/catalog/absint-query/absint-query-format.ts
EagleoutIce: But this is valid? 
Because the REPL will provide you with analyzer pre-initialized with the last data and also fail if there is nothing! 
This allows you to do:

```R
R> :df# "library(ggplot2)\nprint(\"hey\")"
Dataflow calculated in 1ms.
Edges:               4
Vertices:            4
 - Value:                     2
 - Use:                       0
 - FunctionCall:              2
 - VariableDefinition:        0
 - FunctionDefinition:        0
R> :query @dependencies
Query: dependencies (16 ms)
   ╰ Libraries
       ╰ `library`
           ╰ Node Id: 3, `ggplot2`
   ╰ Written Data
       ╰ `print`
           ╰ Node Id: 7, `stdout`
All queries together required ≈17 ms (1ms accuracy, total 21 ms)
R> 
```

which is imho. much nicer than needing to repeat the code

EagleoutIce:
_This pull request is included in v2.12.1 (see [Release v2.12.1 (Fix Docker Releases)](https://github.com/flowr-analysis/flowr/releases/tag/v2.12.1))._

PR #2597, opened 2026-06-27
feat: dicing

### LINKED ISSUE #902: Dicing
We want to extend the slicing by adding support for dicing.

This should allow us to only select paths from a selection of start nodes that reach a given selection of end nodes. For this we want to utilize the current backwards slicing from the end nodes and find only those paths that lead to the start nodes.

Example:
Input:
a = 3
b = 4
a + b
--diced from a to +
Output:
a = 3
a + b

EagleoutIce:


EagleoutIce:

PR #2603, opened 2026-07-03
refactor: clean up argument handling on script invocations

### LINKED ISSUE #1638: Incorrect parameter order is not caught when calling a script from flowr
This call:

`npm run flowr -- --verbose slicer` 

is not equal to this one:

`npm run flowr -- slicer --verbose`

The first one will call the slicer script and pass `slicer` as the first argument.
The second one will correctly pass `--verbose` as first argument to the `slicer` script.

The first parameter combination does not make sense and should not be allowed.

### Why does this happen?

"Any values unaccounted for by an option definition will be set on the `defaultOption`" (which is `script` in flowr)
https://github.com/75lb/command-line-args/blob/master/doc/option-definition.md#module_option-definition--OptionDefinition+defaultOption

https://github.com/flowr-analysis/flowr/blob/3900e7b2f6432adfe021a4e861def099637212d3/src/cli/flowr-main-options.ts#L64

We will then pass all parameters beginning with the third one to the script:

https://github.com/flowr-analysis/flowr/blob/3900e7b2f6432adfe021a4e861def099637212d3/src/cli/flowr.ts#L159

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.11.0 (see [Release v2.11.0 (Package Database Support)](https://github.com/flowr-analysis/flowr/releases/tag/v2.11.0))._

PR #2613, opened 2026-07-05
Improve network functions linting rule list

### LINKED ISSUE #2564: Improve network functions linting rule list
The list of functions for the network functions linter rule has some issues, namely:
- [ ] the function ist is very small, we should support more functions!
- [x] `onlyTriggerWithArgument` is not linked to functions in any way -> it should be toggleable for each function rather than globally! Maybe using the `FunctionInfo` construct we already have?

Ellpeck:


src/linter/rules/network-functions.ts
```diff
@@ -8,29 +8,50 @@ import type { ParentInformation } from '../../r-bridge/lang-4.x/ast/model/proces
...
-	/** only trigger if the function's read argument is linked to a value that matches this pattern */
+	/**
+	 * The list of function names or more detailed {@link NetworkFunction} information that should be marked in the given context if their arguments match.
+	 */
+	fns:                      readonly (string | NetworkFunction)[]
+	/**
+	 * Only trigger if the function's read argument is linked to a value that matches this pattern.
+	 * This value is only used for entries in {@link fns} that are either strings, or whose {@link NetworkFunction.onlyTriggerWithArgument} value is unset.
+	 */
+	onlyTriggerWithArgument?: RegExp | string
```
EagleoutIce: couldn't we move this *fully* into the NetworkFunction type as well? just for consistency? because having this globally feels unelegant

src/linter/rules/network-functions.ts
```diff
@@ -8,29 +8,50 @@ import type { ParentInformation } from '../../r-bridge/lang-4.x/ast/model/proces
...
 			async(es) => {
 				const res: (FlowrSearchElement<ParentInformation> & { certainty: LintingResultCertainty })[] = [];
 				for(const e of es) {
-					const val = await functionFinderUtil.requireArgumentValue(
-						e,
-						FnPool,
-						d,
-						c.onlyTriggerWithArgument);
-
+					const val = await functionFinderUtil.requireArgumentValue(e, fnPool, d, onlyTriggerLookup.get(e.node.lexeme ?? '') ?? c.onlyTriggerWithArgument);
```
EagleoutIce: if we do not have a lexeme we could skip?

src/linter/rules/network-functions.ts
```diff
@@ -53,11 +74,17 @@ export const NETWORK_FUNCTIONS = {
 		description:   'Marks network functions that execute network operations, such as downloading files or making HTTP requests.',
 		defaultConfig: {
 			fns: [
-				'read.table', 'read.csv', 'read.csv2', 'read.delim', 'read.delim2', 'readRDS', 'download.file', 'url', 'GET', 'POST', 'PUT',
+				// subset of default network functions provided by manually reviewed results from analysis based on crawlR, see the dirty-network-funcs-test branch in flowr-analysis/crawlr
+				'read.table', 'read.csv', 'read.csv2', 'read.delim', 'read.delim2', 'readRDS', 'download.file', 'url', 'GET', 'POST', 'PUT', 'BROWSE', 'RETRY', 'VERB',
```
EagleoutIce: 1) please do not mention crawlR here
2) could we please also migrate this all to `Identifier.make`? to have the packages?

EagleoutIce (approved):

src/linter/rules/network-functions.ts
Ellpeck: Hmm, but then the default value of `/^(https?|ftps?):\/\//` would have to be set explicitly for every default function, which seems much less elegant to me. I honestly quite like the "global" default.

src/linter/rules/network-functions.ts
Ellpeck: Should this PR also include a change to the Regex comparison that actually includes a comparison by package and not just ignoring it?

src/linter/rules/network-functions.ts
EagleoutIce: But the problem is it is harder to overwrite and when you want to change the global default it affects countless functions. It can still be a default object that gets passed to all the individual functions / we could use a loo/map to initialize them but the global default feels dirty to me especially because it's argument independent

src/linter/rules/network-functions.ts
EagleoutIce: That would be nice but only if it does not require too much machinery maybe we have to first revamp the call context query and if so this should be a separate issue

src/dataflow/environments/identifier.ts
```diff
@@ -166,6 +166,13 @@ export const Identifier = {
 		const targetInternal = Identifier.accessesInternal(target);
 		return idInternal === targetInternal;
 	},
+	/**
+	 * Helper to create a regular expression that matches against an array of {@link Identifier} values. If both the passed identifier and the matched identifier are namespaced, their namespaces are expected to match. If either is not namespaced, the namespace is ignored on both.
+	 */
+	regex(this: void, identifiers: readonly Identifier[]): RegExp {
```
EagleoutIce: maybe arrayRegex? or a spread?

EagleoutIce (approved):

EagleoutIce:
_This pull request is included in v2.13.5 (see [Release v2.13.5 (Resolve, Orphan, Incremental Improvements)](https://github.com/flowr-analysis/flowr/releases/tag/v2.13.5))._

PR #2630, opened 2026-07-09
[Linter]: Support for deprecated arguments

### LINKED ISSUE #1870: [Linter] Deprecated arguments support
Currently, the deprecated functions linter has no support for functions that themselves are not deprecated, but have deprecated arguments

gigalasr:
closes #1870

TODO:
- [x] Support for ifValue for args
- [x] Support version constraint
- [x] Support for Identifiers with package name


src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,135 @@
...
+	 */
+	fns:      Fns
+	/**
+	 * Additionally, constraint the detection of an fn in {@link DeprecatedFunctionsConfig.fns} if it is called with an Argument
+	 * defined by {@link DeprecatedFunctionsConfig.whenArgs}
+	 */
+	whenArgs: Partial<Record<Fns[number], DeprecatedArgumentInformation[]>>
+}
+// Little helper for constraining whenArgs to fns
+const defineConfig = <const Fns extends readonly string[]>(config: DeprecatedFunctionsConfig<Fns>) => config;
```
gigalasr: @EagleoutIce Is there a better way to constraint the `whenArgs`map to keys from `fns` ? I was only able to enforce this with generics in a small helper function.

gigalasr:
Questions:
1. Should the linter rule return the node id and source range of the entire function, or just of the deprecated arg?
2. How can I make this work with the new identifiers that are prefixed with the package name?

src/linter/rules/deprecated-functions.ts
EagleoutIce: can you please show me what you wantto have as the result? like should this be a record that only has properties with names from Fns?

gigalasr: Yes, whenArgs should only have properties with names from Fns, as in line 127 in the definition of DEPRECATED_FUNCTIONS

EagleoutIce:
>     1. Should the linter rule return the node id and source range of the entire function, or just of the deprecated arg?

Just for the argument please

>     2. How can I make this work with the new identifiers that are prefixed with the package name?

unsure what you mean here because the argument names cannot be namespaced? Could you please elaborate with an example?

src/linter/rules/deprecated-functions.ts
EagleoutIce: I mean you could also use the index signature but isnt there an as const or somethingmissing in the default config? but generally you do have to, in any way (as cast, const, ...) have to brin ghte precise type of the fucntion list into the Fns generic

gigalasr:
1. Oki, thank you!
2. In line 76 (https://github.com/flowr-analysis/flowr/pull/2630/changes#diff-08ba3901670b25d88a746877af089e407a93f5ec8a0b09db4b1bf47e1eb71c4aR76) I use the target from the CallTargetsEnrichment to index the whenArgs record. However, target can have names like `deply:someFn`, but the whenArgs record (currently) only contains names like `someFn`.

EagleoutIce:
Aah okay so you can use either `Identifier::toQualified` or also `Identifier::matches` (or match? not sure rn) to bridge the gap - is this what you needed?

gigalasr:
Sounds like it, I will try that! Thank you!

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+import { parseRRange } from '../../util/r-version';
+
+/**
+ * Information about an argument of a function that should be flagged as deprecated if it is called with this argument
+ */
+export interface DeprecatedArgumentInformation {
+	argIdx?:       number,
+	argName?:      string
+	ifValue?:      RegExp | string
+	replacedBy?:   string
```
EagleoutIce: please document how to encode this here? i.e., just the argument? or are we also allowing different functions etc.

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+	state?:        DeprecationState
+}
+
+/**
+ * Information about a deprecated function
+ */
+export interface DeprecatedFunctionInformation {
+	name:          string
+	/** Used to mark specific arguments as deprecated */
+	whenArgs?:     DeprecatedArgumentInformation[]
```
EagleoutIce: readonly?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+export interface DeprecatedArgumentResult extends DeprecatedFunctionResultBase {
+	type: 'deprecated-argument'
+	arg:  string | number
+}
+
+export type DeprecatedFunctionRuleResult = DeprecatedFunctionResult | DeprecatedArgumentResult;
+
+export enum DeprecationState {
+	/** Still works, but marked for removal */
+	Deprecated = 'deprecated',
```
EagleoutIce, line 105: can we reference an example for this, like with a url?

src/linter/rules/deprecated-functions.ts
EagleoutIce: and, only provide this if superseeded? or what?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+	/** No longer works, and marked for removal */
+	Defunct = 'defunct',
+	/** Replaced by another function but kept for compatibility */
+	Superseeded = 'superseeded'
+}
+
+export interface DeprecatedFunctionsConfig extends MergeableRecord{
+	/** Functions to mark as deprecated */
+	fns: (string | DeprecatedFunctionInformation)[]
```
EagleoutIce: Why not Identifier?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+	name:          string
+	/** Used to mark specific arguments as deprecated */
+	whenArgs?:     DeprecatedArgumentInformation[]
+	replacedBy?:   string
+	sinceVersion?: Range
+	state?:        DeprecationState
+}
+
+export interface DeprecatedFunctionResultBase extends LintingResult {
+	function:      BrandedIdentifier
```
EagleoutIce: Why just branded? please use the full Identifier here

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+export interface DeprecatedFunctionInformation {
+	name:          string
+	/** Used to mark specific arguments as deprecated */
+	whenArgs?:     DeprecatedArgumentInformation[]
+	replacedBy?:   string
+	sinceVersion?: Range
+	state?:        DeprecationState
+}
+
+export interface DeprecatedFunctionResultBase extends LintingResult {
```
EagleoutIce, line 70: doc for exported interface?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
-	createSearch:        (config) => functionFinderUtil.createSearch(config.fns),
-	processSearchResult: functionFinderUtil.processSearchResult,
-	prettyPrint:         functionFinderUtil.prettyPrint('deprecated'),
-	info:                {
+	createSearch:        (config) => functionFinderUtil.createSearch(config.fns.map(fn => typeof fn === 'string' ? fn : fn.name)),
+	processSearchResult: async(elements, config, data) => {
+		const fnInfosEx = Object.fromEntries(
+			config.fns
+				.filter(f => typeof f === 'object')
+				.map(f => [f.name, f])
```
EagleoutIce: why not directly expect a record with this mapping? :D

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+				} satisfies DeprecatedFunctionResult];
+			} else if(info.whenArgs !== undefined) {
+				// If whenArgs is set, the function should only be deprecated if it has one of the provided args
+				return info.whenArgs.map(deprecatedArgInfo => {
+					const vertex = dataflow.getVertex(e.node.info.id);
+					if(vertex === undefined || !isFunctionCallVertex(vertex)) {
+						return undefined;
+					}
+
+					const arg = vertex.args.find((arg, idx) =>
```
EagleoutIce: we have argument finders like pMatch for this, please use them :D 
There is also getValueOfArgument`

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+							arg:          (deprecatedArgInfo.argName ?? deprecatedArgInfo.argIdx) as string | number,
+							state:        deprecatedArgInfo.state,
+							replacedBy:   deprecatedArgInfo.replacedBy,
+							sinceVersion: deprecatedArgInfo.sinceVersion,
+							loc:          SourceLocation.fromNode(argNode) ?? e.sourceLocation
+						} satisfies DeprecatedArgumentResult;
+					}
+
+				// Don't mark function as deprecated, if we didn't find any deprecated args
+				}).filter(p => p !== undefined);
```
EagleoutIce: please use isNotUndefined here for more typesafety

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+					);
+					const argNode = arg === undefined || arg === EmptyArgument ? undefined : idMap.get(arg.nodeId);
+
+					if(argNode !== undefined) {
+						return {
+							type:         'deprecated-argument',
+							certainty:    LintingResultCertainty.Certain,
+							involvedId:   argNode.info.id,
+							function:     e.target,
+							arg:          (deprecatedArgInfo.argName ?? deprecatedArgInfo.argIdx) as string | number,
```
EagleoutIce: is the cast necessary?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,19 +1,194 @@
...
+		[LintingPrettyPrintContext.Query]: (result: DeprecatedFunctionRuleResult) => `${result.type === 'deprecated-argument' ? `Argument \`${result.arg}\` of ` : ''}Function \`${result.function}\` at ${SourceLocation.format(result.loc)}`,
+		[LintingPrettyPrintContext.Full]:  (result: DeprecatedFunctionRuleResult) => {
+			const str: string[] = [];
+			if(result.type === 'deprecated-argument') {
+				const argStr = typeof result.arg === 'number' ? `at position \`${result.arg}\`` : result.arg;
+				str.push(`Argument \`${argStr}\` of`);
+			}
+			str.push(`Function \`${result.function}\` is ${result.state ?? 'deprecated'}`);
+			if(result.sinceVersion) {
+				str.push(`since version ${result.sinceVersion.format()}`);
```
EagleoutIce, line 226: shouldn't we check whether the since Version range contract is even fulfilled?

src/linter/rules/deprecated-functions.ts
gigalasr: Hm, can't a package deprecate a function but already provide a replacement?

For example in [geom_violin](https://ggplot2.tidyverse.org/reference/geom_violin.html), the `draw_quantiles` parameter is marked as deprecated, and the warning says the user should use `quantiles.linetype` instead

EagleoutIce: they can yes, so it does not necessearily be linked to superseeded.

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+	target:         Identifier;
+	sourceLocation: SourceLocation
+}
+
+interface Metadata extends MergeableRecord {
+	sigdb:     number,
+	hardcoded: number
+}
+
+const AlwaysDeprecated = [
```
EagleoutIce, line 128: can we map these to the built in call props

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+];
+
+const ConditionallyDeprecated = {
```
EagleoutIce, line 132: nice as an example :) maybe reference the deprecation notice/url?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+}
+
+/**
+ * Returned by the {@link DEPRECATED_FUNCTIONS} linting rule, when a deprecated argument is detected.
+ * Provided for convince to differentiate between {@link DeprecatedArgumentResult} and {@link DeprecatedFunctionResult}
+ */
+export interface DeprecatedArgumentResult extends DeprecatedFunctionResultBase {
+	readonly type: 'deprecated-argument'
+	/** The name or index of the deprecated argument */
+	readonly arg:  string | number
```
EagleoutIce, line 96: this does not make it clear what this will contain. like why allow both? every parameter has a name so there should be one. is this automatically a name if it was given named and index if unnamed?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+
+export type DeprecatedFunctionRuleResult = DeprecatedFunctionResult | DeprecatedArgumentResult;
+
+export enum DeprecationState {
+	/** A better alternative is available, but the function is kept (softer alternative to deprecated) {@link https://lifecycle.r-lib.org/articles/stages.html#superseded} */
+	Superseeded = 'superseeded',
+	/** A better alternative is available, and the function is marked for removal {@link https://lifecycle.r-lib.org/articles/stages.html#deprecated} */
+	Deprecated = 'deprecated',
+	/** No longer works and is removed and replaced by another function {@link https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/Defunct} */
+	Defunct = 'defunct'
```
EagleoutIce, line 107: is defunct associated with exceptions? if so we could pipe this into flowR's exception handling

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+}
+
+interface PotentialFunction {
+	node:           RNode<ParentInformation>;
+	target:         Identifier;
+	sourceLocation: SourceLocation
+}
+
+interface Metadata extends MergeableRecord {
+	sigdb:     number,
```
EagleoutIce, line 124: does this code the number of hits from the sigdb, the number of entries in the sigdb.... this should be made clearer

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+	hardcoded: number
```
EagleoutIce, line 125: likewise here? in which sense do we want to map "hardcoded" what is that supposed to refer to?

EagleoutIce: the preset list?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+} satisfies Record<BrandedIdentifier, DeprecatedFunctionInformation>;
 
 export const DEPRECATED_FUNCTIONS = {
 	// unlike functionFinderUtil.createSearch(config.fns), this does not pre-filter to the hardcoded list: the
 	// sigdb-driven pass below needs every resolved call, so the `fns` filtering happens in processSearchResult instead
-	createSearch: (_config: FunctionsToDetectConfig) => Q.all().filter(VertexType.FunctionCall).with(Enrichment.CallTargets, {
+	createSearch: (_config) => Q.all().filter(VertexType.FunctionCall).with(Enrichment.CallTargets, {
 		onlyBuiltin:  true,
-		qualifyNames: false // we don't use qualified names for this rule yet
+		qualifyNames: true // we don't use qualified names for this rule yet
```
EagleoutIce, line 141: changing this should imply removing the comment :P

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
 	}),
 	processSearchResult: async(elements, config, data) => {
-		const matchesConfiguredFns = Identifier.regex(...config.fns);
-		const hardcoded = await functionFinderUtil.processSearchResult(elements, config, data, es =>
-			es.filter(e => enrichmentContent(e, Enrichment.CallTargets)?.targets
-				.some(t => typeof t === 'string' && matchesConfiguredFns.test(t))));
+		const matchesConfiguredFns = Identifier.regex(...config.always);
+		const graph = (await data.dataflow()).graph;
+		const idMap = (await data.normalize()).idMap;
```
EagleoutIce, line 146: you could also use the idmap from the graph.

src/linter/rules/deprecated-functions.ts
```diff
@@ -58,23 +209,141 @@ export const DEPRECATED_FUNCTIONS = {
...
 	}
-} as const satisfies LintingRule<FunctionsResult, FunctionsMetadata, FunctionsToDetectConfig>;
+} as const satisfies LintingRule<DeprecatedFunctionRuleResult, Metadata, DeprecatedFunctionsConfig>;
+
+/**
+ * This function is applied to function candidates that have an entry in the {@link DeprecatedFunctionsConfig.conditionally} map.
+ */
+async function deprecateFunctionConditionally(candidate: PotentialFunction, dataflow: DataflowGraph, idMap: AstIdMap, analyzer: ReadonlyFlowrAnalysisProvider<KnownParser>, info: DeprecatedFunctionInformation): Promise<DeprecatedFunctionRuleResult[]> {
+	const results: DeprecatedFunctionRuleResult[] = [];
+	const result = await analyzer.query([{
```
EagleoutIce, line 253: uuuh this is a performance hazard. check whether you need to do this first, then do it once and take the info. even if you guess just one package this has to solve a big system.

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+				}
+			});
+		}).filter(p => isNotUndefined(p));
 
+		// 2. Uses hardcoded information about deprecated arguments and deprecated functions
+		const results: DeprecatedFunctionRuleResult[] = (await Promise.all(detectedFunctions.map(async candidate => {
+			const name = candidate.target.includes('::') ? Identifier.getName(Identifier.parse(candidate.target)) : candidate.target;
+			const info = config.conditionally[name];
+			if(isNotUndefined(info)) {
+				// Check functions from DeprecatedFunctionsConfig.conditionally
```
EagleoutIce, line 165: any reasons for these comments?

src/linter/rules/deprecated-functions.ts
```diff
@@ -1,36 +1,186 @@
...
+					return {
+						node: e.node, target: target as string, sourceLocation
+					};
+				}
+			});
+		}).filter(p => isNotUndefined(p));
 
+		// 2. Uses hardcoded information about deprecated arguments and deprecated functions
+		const results: DeprecatedFunctionRuleResult[] = (await Promise.all(detectedFunctions.map(async candidate => {
+			const name = candidate.target.includes('::') ? Identifier.getName(Identifier.parse(candidate.target)) : candidate.target;
```
EagleoutIce, line 162: why even check for `::` Identifier shuold be doing this just fine! I nmean its technically fine as a safe guard but this could go south

src/linter/rules/deprecated-functions.ts
```diff
@@ -58,23 +209,141 @@ export const DEPRECATED_FUNCTIONS = {
...
+	const result = await analyzer.query([{
+		type:     'guess-dep-versions',
+		packages: [info.package]
+	}]);
+	const derrivedRangeRaw = result['guess-dep-versions'].dependencies.find(d => d.package === info.package)?.range;
+	const derrivedRange = derrivedRangeRaw !== undefined ? RRange.parse(derrivedRangeRaw) : undefined;
+
+	// Deprecated Argument: If `whenArgs` is provided, only mark deprecated arguments
+	if(info.whenArgs) {
+		const vertex = dataflow.getVertex(candidate.node.info.id);
```
EagleoutIce, line 262: could we please structure this into helper functions? :D

src/linter/rules/function-finder-util.ts
```diff
@@ -105,35 +107,55 @@ export const functionFinderUtil = {
...
 		}
+
 		return Ternary.Never;
 	}
 };
+
+/**
+ * Test if a function call has an argument with a specific value
+ */
+export function hasArgumentValue(
```
EagleoutIce, line 134: this could be.. should be moved in to the Rargument helper object i think

src/linter/rules/deprecated-functions.ts
gigalasr: Defunct comes after deprecated. In most cases, a deprecated function will eventually just be deleted. For very important functions, we’ll instead make the function defunct, which means that function continues to exist but the deprecation warning turns into an error. ([sauce](https://lifecycle.r-lib.org/articles/stages.html)).

Is this something for this pr, or a new pr?

src/linter/rules/function-finder-util.ts
gigalasr: Hm, I get a weird cyclic import when moving it there :/

EagleoutIce:

PR #2646, opened 2026-07-19
Version guessing

EagleoutIce:
- sigdb records S3/S4 class ownership, so a class use points at the package that owns it
- **version guessing**: proper solver over the version space, bounded by R release dates
- call graph can expand library calls into their internal callees
- **importFrom** support and namespace imports are tracked
- signature db is searchable by parameter
- new **logLevel config entry**, honored at load and at runtime
- `:version` lists registered plugins and which ones actually activated
- query keys fuzzy-complete in the repl
- **`--no-fs`** to run without filesystem loading
- linter rules can be specialized individually
- Integrate **RDA** support as a real **plugin**


EagleoutIce:
_This pull request is included in v2.13.0 (see [Release v2.13.0 (Guessing Dep. Versions, Specialization, RDA)](https://github.com/flowr-analysis/flowr/releases/tag/v2.13.0))._

PR #2675, opened 2026-07-29
Consistent Function Info

### LINKED ISSUE #2673: Consistent Function Info
1. **Identity return**: the function just hands back one of its arguments.                                                                                                                                                        
  2. **Result provenance**: which parameter positions the return value is computed from.                                                                                                                                            
  3. **Shape-only dependence**: the result depends only on shape and attributes of an argument, not its contents (`length`, `nrow`).                                                                                                
  4. **Parameter role**: which formals are control flags, which carry data, which are output sinks.                                                                                                                                 
  5. **Purity**: the function is free of side effects.                                                                                                                                                                              
  6. **Side-effect targets**: which inputs may be written, by reference (environments, R5/RC, `data.table`, external pointers) or via `<<-` and `assign`.                                                                           
  7. **Access mode**: read only vs write.                                                                                                                                                                                           
  8. **Can throw**: the call may raise a condition, and finer, may warn or signal.                                                                                                                                                  
  9. **Dispatch**: S3, S4, or S7 magic happens here, and whether the function is a generic.                                                                                                                                         
  10. **Strictness per parameter**: which arguments are always forced, never forced, or forced only conditionally.                                                                                                                  
  11. **Higher-order positions**: which formals are called rather than read.                                                                                                                                                        
  12. **NSE positions**: which arguments are quoted or evaluated in a masked frame.                                                                                                                                                 
  13. **Dots forwarding**: whether `...` is consumed here or passed on to a callee.                                                                                                                                                 
  14. **Scope effects**: it defines, removes, or rebinds names in the caller or global environment.                                                                                                                                 
  15. **Control flow**: non-local exits, never returns (`stop`, `q`), establishes a handler (`tryCatch`).                                                                                                                           
  16. **Determinism source**: non-deterministic, and whether that comes from the RNG or from clock and environment.                                                                                                                 
  17. **Resource class**: which resource the effect touches, such as filesystem, network, process, graphics device, or database.                                                                                                    
  18. **Abstractly evaluable**: flowR may constant-fold the call, so value resolution can cross it.                                                                                                                                 
  19. **Result aliasing**: the result aliases an argument object, the general case of item 1.                                                                                                                                       
  20. **Result visibility**: the result is returned invisibly.                                                                                                                                                                      
  21. **Presence sensitivity**: the result depends on whether an argument was supplied, not on its value.                                                                                                                           
  22. **Arity contract**: variadic vs fixed, and sensitivity to partial matching.

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.13.5 (see [Release v2.13.5 (Resolve, Orphan, Incremental Improvements)](https://github.com/flowr-analysis/flowr/releases/tag/v2.13.5))._

PR #2678, opened 2026-07-30
[Incremental Parsing]: Heuristics

### LINKED ISSUE #2676: Heuristics for triggering incremental parsing
Right now incremental parsing is a simple on/off switch through the config. We should add some heuristics to when the incremental parsing comes to action if it is activated.

First ideas:
- `mtime`: if `true`, skip reparsing entirely when the file's modification time is unchanged since the last parse.
-  `linesFrom`: only consider incremental parsing for files with at least this many lines.
- `bytesFrom`: only consider incremental parsing for files with at least this many bytes (rather this one).
- `alwaysWithEdits`: always take the incremental path whenever there is a computed edit region, regardless of other thresholds.
- `minFiles`: only apply these heuristics once a project has at least this many files loaded.

For this we need some sort of `shouldReparseHandler`, which determines, if a file should be reparsed based on the config.

See `call-context-query-executer.ts` `promoteCallName` as an orientation

stimjannik:
Related to #2676


src/documentation/wiki-interface.ts
```diff
@@ -313,7 +313,15 @@ ${codeBlock('json', JSON.stringify(
 				}
 			},
 			incrementalParsing: {
-				activated: false,
+				activated:  false,
+				heuristics: {
+					activated:       true,
```
EagleoutIce: maybe we could rephrase this into `alwaysIncremental: false` to make clear that this does not activate the incremental parsing in and of itself? 
Also i think i would prefer if we'd nest `incremental: { parsing: { ... } }` for the other heuristics for dataflow in the future

src/project/context/flowr-analyzer-files-context.ts
```diff
@@ -156,6 +156,8 @@ export interface ReadOnlyFlowrAnalyzerFilesContext {
 	 * folder for a single-file request. Useful to report back which inputs did not resolve to anything.
 	 */
 	getRequestedRoots(): readonly string[];
+
+	getFileCount(): number;
```
EagleoutIce, line 169: please doc :D even if this is supposed to be trivial - which files are counted? R source files? all files? ...
inline files...

EagleoutIce: How does this differ to `getFileByRole(...).length` etc.

src/project/context/flowr-analyzer-incremental-analysis-context.ts
```diff
@@ -17,6 +18,7 @@ export interface ReadOnlyFlowrAnalyzerIncrementalAnalysisContext {
 	getOldContentOf(filePath: FilePath): string | undefined;
 }
 
+type ShouldReparseTest = (filePath: FilePath, ctx: FlowrAnalyzerContext) => boolean;
```
EagleoutIce, line 30: Also a documentation of whether it is the job of this function to respect the config

src/project/context/flowr-analyzer-incremental-analysis-context.ts
```diff
@@ -50,6 +53,65 @@ export class FlowrAnalyzerIncrementalAnalysisContext implements ReadOnlyFlowrAna
...
+				let currentMtime: number | undefined;
+				try {
+					currentMtime = fs.statSync(filePath).mtimeMs;
+				} catch{
+					return true;
+				}
+				const lastMtime = ctx.inc.getLastKnownMtime(filePath);
+				ctx.inc.setLastKnownMtime(filePath, currentMtime);
+
+				return !(lastMtime !== undefined && lastMtime === currentMtime);
```
EagleoutIce: I mean we should argue whether this could not cause false updates (e.g., because the mtime gets updated no matter wehther we really reparse incrmentally but this is fine i guess)

EagleoutIce (approved):
Please also address the merge conflicts :D 

+ Can we please have at least one test were we check whether the length is respected?

EagleoutIce:
_This pull request is included in v2.13.10 (see [Release v2.13.10 (Loading from Disk, Bugfixes, Various Improvements)](https://github.com/flowr-analysis/flowr/releases/tag/v2.13.10))._

PR #2693, opened 2026-08-06
[AbsInt] Support multiple abstract interpretation analyses using a single visitor

### LINKED ISSUE #2679: [AbsInt] Support multiple abstract interpretation analyses using a single visitor

OliverGerstl:


src/abstract-interpretation/data-frame/dataframe-semantics.ts
```diff
@@ -0,0 +1,2062 @@
...
+		const byValue = Resolve.argument.value(byArg, info);
+
+		if(typeof byValue === 'string' || typeof byValue === 'number') {
+			byCols = [byValue];
+		} else if(Array.isArray(byValue) && (byValue.every(by => typeof by === 'string') || byValue.every(by => typeof by === 'number'))) {
+			byCols = byValue;
+		}
+	}
+
+	if(byCols?.some(by => typeof by === 'string')) {
```
EagleoutIce: maybe we should have more readable predicates for this? like stringColName or something?

src/abstract-interpretation/data-frame/dataframe-semantics.ts
```diff
@@ -0,0 +1,2062 @@
...
+		return;
+	}
+	const result: DataFrameOperations = [];
+	const summarizeArgs = args.filter(arg => arg !== dataFrame);
+
+	const summarizedCols = summarizeArgs.map(arg => Resolve.argument.toName(arg, info));
+
+	// only column names that are not created by summarize are preconditions on the operand
+	const accessedNames = summarizeArgs
+		.flatMap(arg => getUnresolvedSymbolsInExpression(arg, info.graph).map(Identifier.toString))
```
EagleoutIce: it feels weird to use Identifier to String to compare this to the argument.toName return

EagleoutIce (approved):

src/abstract-interpretation/data-frame/dataframe-semantics.ts
OliverGerstl: Yes, but the `toName` returns the argument names which are the column names that are newly created by the `summarize` operation. The identifier names of the unresolved symbols in the `summarize` expression args are the column names that are implicitly accessed in the `summarize` operation. Hence, the column names that existed before and are accessed in `summarize` (for `accessCols`) are all implicitly accessed column names that are not created by the `summarize` operation itself

OliverGerstl: Also, the `getUnresolvedSymbolsInExpression` only returns Identifiers that have no namespace. Maybe I make this explicit in the function signatue

EagleoutIce: the last point is what I wanted to get at :D maybe just use Identifier getName?

OliverGerstl: I will just make `getUnresolvedSymbolsInExpression` return an array of `BrandedIdentifiers` :+1:

EagleoutIce: that works too :D

EagleoutIce:
_This pull request is included in v2.15.0 (see [Release v2.15.0 (Semantic Props, Unescaped Arguments, Shared AbsInt Visitor, Bugfixes)](https://github.com/flowr-analysis/flowr/releases/tag/v2.15.0))._

PR #2723, opened 2026-08-19
[DF] remove killed symbols from out list

### LINKED ISSUE #2722: Remove killed symbols from out list
For the following example:
```R
x <- 1
rm(x)
``` 

the `graph.out` list contains `x`, but should instead not contain `x` as it was removed after its definition. 

---

The swapped case `rm(x); x<-1` works correctly, and is only detected as a def.

EagleoutIce:


EagleoutIce:

PR #2726, opened 2026-08-20
Track implicit echo in dependencies query

### LINKED ISSUE #2725: [Dep] Track top-level outputs with implicit echo, statistical tests
We want to have a new dep. view category for statistical tests (t.test, anova, etc.).
Also we want flowR to be configurable to whether 
1) assume implicit echo (also based on the project type) 
2) if so, collect implicit echo outputs (non-invisible etc.) for the dependency query

EagleoutIce:


EagleoutIce:

PR #2730, opened 2026-08-21
New CFG API

EagleoutIce:
Intertwined with Dataflow, no more exit markers, call helpers, edge order invert


EagleoutIce:

PR #2739, opened 2026-08-25
Fix id, interpolate ids, perf heavy loop

EagleoutIce:


EagleoutIce:
_This pull request is included in v2.15.1 (see [Release v2.15.1 (fix: index layout issue, double id for interpolation)](https://github.com/flowr-analysis/flowr/releases/tag/v2.15.1))._

Issue #2687, opened 2026-08-05
[taint-analysis] Integrate flowRs function classification system


MaxAtoms:
cf. `BuiltInIndex` in `query-fn-props.ts`

Issue #2686, opened 2026-08-05
[taint-analysis] Include mappings for the top-n functions


MaxAtoms:
Related: #2687

Prefer to use flowR's system, if applicable.

PR #2740, opened 2026-08-25
Add function categories and additional mappings for most called functions

MaxAtoms:
Closes #2687 and #2686

### DIFF
--- a/src/taint-analysis/builder/taint-analysis-definition.ts
+++ b/src/taint-analysis/builder/taint-analysis-definition.ts
@@ -1,5 +1,8 @@
-import type { TaintMapper, TaintRole } from '../function-mapper';
-import type { AbsintVisitorConfiguration, AbstractInterpretationVisitor } from '../../abstract-interpretation/absint-visitor';
+import type { TaintMapper, TaintMapping, TaintRole } from '../function-mapper';
+import type {
+	AbsintVisitorConfiguration,
+	AbstractInterpretationVisitor
+} from '../../abstract-interpretation/absint-visitor';
 import type { AnyStateDomain } from '../../abstract-interpretation/domains/state-domain-like';
 import type { TaintComponent, TaintProduct } from '../composite-taint-visitor';
 import { CompositeTaintInferenceVisitor } from '../composite-taint-visitor';
@@ -7,7 +10,12 @@ import type { TaintVisitorConfiguration } from '../taint-visitor';
 import { TaintInferenceVisitor } from '../taint-visitor';
 import { guard } from '../../util/assert';
 import type { ProductReduction } from '../../abstract-interpretation/domains/partial-product-domain';
-import type { AnyAbstractDomain } from '../../abstract-interpretation/domains/abstract-domain';
+import type { AbstractValue, AnyAbstractDomain } from '../../abstract-interpretation/domains/abstract-domain';
+import { AbstractDomain } from '../../abstract-interpretation/domains/abstract-domain';
+import { BuiltInIndex } from '../../dataflow/environments/query-fn-props';
+import type { ArgProps } from '../../dataflow/environments/built-in-props';
+import { ArgProp } from '../../dataflow/environments/built-in-props';
+import { Identifier } from '../../dataflow/environments/identifier';
 
 export type TaintAnalysisName<Definition> =
 	Definition extends RunnableTaintAnalysisDefinition<infer Name> ? Name : never;
@@ -22,6 +30,7 @@ export interface RunnableTaintAnalysisDefinition<Name extends string = string> {
 	readonly name: Name;
 	/** The optional message reported when the analysis produces a finding. */
 	readonly msg?: string;
+
 	/** Creates the abstract interpretation visitor that conducts the taint analysis for the given visitor configuration. */
 	createVisitor(config: AbsintVisitorConfiguration): AbstractInterpretationVisitor<AnyStateDomain>;
 }
@@ -38,7 +47,8 @@ export interface ComposeOptions {
 }
 
 /**
- * Fluent builder class for defining new taint analyses.
+ * A singular taint analysis definition or a combination of multiple taint analyses
+ * creatable via {@link TaintAnalysisDefinition.compose}.
  */
 export class TaintAnalysisDefinition<Name extends string = string, Domain extends AnyAbstractDomain = AnyAbstractDomain, Config extends AbsintVisitorConfiguration = AbsintVisitorConfiguration>
 implements RunnableTaintAnalysisDefinition<Name> {
@@ -59,6 +69,38 @@ implements RunnableTaintAnalysisDefinition<Name> {
 		this.config = config;
 	}
 
+	/** Adds default propagators (i.e. pure functions for which we know that the return value is based on at least one argument) */
+	public withDefaultPropagators(argProps: ArgProps = ArgProp.Alias | ArgProp.Value | ArgProp.Shape): this {
+		const idx = BuiltInIndex.default();
+		for(const i of idx.pure) {
+			const sig = idx.entries.find(e => Identifier.matches(i, e.name))?.sig;
+			if(!sig){
+				continue;
+			}
+
+			const containingArgProp = sig.filter(([_name, arg]) => (arg & argProps) !== 0);
+			const arg = containingArgProp.map(([name], idx) => {
+				return { pos: idx, name: name };
+			});
+
+			const handler: TaintMapping<Domain> = {
+				identifier: i,
+				role:       'through',
+				condition:  {
+					argTaints: arg,
+					condition: (_, taints) => {
+						const taintDomains = taints.map(t => this.domain.create(t));
+						return AbstractDomain.joinAll(taintDomains).value as AbstractValue<Domain>;
+					}
+				},
+			};
+
+			this.mapper.push(handler);
+		}
+
+		return this;
+	}
+
 	/** Adds source rules that determine the resulting taint of matching calls. */
 	public from(fnMapping: TaintMapper<Domain>): this {
 		this.mapper.push(...fnMapping.map(m => ({ ...m, role: 'from' as TaintRole })));
@@ -77,6 +119,7 @@ implements RunnableTaintAnalysisDefinition<Name> {
 		return this;
 	}
 
+	/** A human-readable message displayed when a finding occurred. */
 	public report(msg: string): this {
 		this._msg = msg;
 		return this;
--- a/src/taint-analysis/predefined/security-analysis.ts
+++ b/src/taint-analysis/predefined/security-analysis.ts
@@ -27,6 +27,7 @@ const protocolTaint = (path: unknown) =>
 	typeof path === 'string' && NetworkProtocolRegex.test(path) ? NetworkInput : FileInput;
 
 export const securityAnalysis = new TaintAnalysisDefinition('security', securityDomain)
+	.withDefaultPropagators()
 	.from([
 		{
 			identifier: [...BuiltInIndex.default().with(CallProp.User)],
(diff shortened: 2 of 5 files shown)


src/taint-analysis/builder/taint-analysis-definition.ts
```diff
@@ -59,6 +69,38 @@ implements RunnableTaintAnalysisDefinition<Name> {
 		this.config = config;
 	}
 
+	/** Adds default propagators (i.e. pure functions for which we know that the return value is based on at least one argument) */
+	public withDefaultPropagators(argProps: ArgProps = ArgProp.Alias | ArgProp.Value | ArgProp.Shape): this {
+		const idx = BuiltInIndex.default();
+		for(const i of idx.pure) {
+			const sig = idx.entries.find(e => Identifier.matches(i, e.name))?.sig;
```
MaxAtoms: I did not use `BuiltInIndex.get()` to be namespace-save

src/taint-analysis/builder/taint-analysis-definition.ts
```diff
@@ -59,6 +69,38 @@ implements RunnableTaintAnalysisDefinition<Name> {
...
+			if(!sig){
+				continue;
```
MaxAtoms: What should we do if there is no signature information?
Simply join all incoming taints, if there are any?

src/taint-analysis/builder/taint-analysis-definition.ts
```diff
@@ -59,6 +69,38 @@ implements RunnableTaintAnalysisDefinition<Name> {
...
+			});
+
+			const handler: TaintMapping<Domain> = {
+				identifier: i,
+				role:       'through',
+				condition:  {
+					argTaints: arg,
+					condition: (_, taints) => {
+						const taintDomains = taints.map(t => this.domain.create(t));
+						return AbstractDomain.joinAll(taintDomains).value as AbstractValue<Domain>;
```
MaxAtoms: I should probably make this customizable, as discussed

src/taint-analysis/builder/taint-analysis-definition.ts
EagleoutIce: revised in 2.15.2 its now supporting this

src/taint-analysis/builder/taint-analysis-definition.ts
EagleoutIce: go to top? 🙈 like then it is a general unknown and it should be at least treated like a sigdb entry with no more info

src/taint-analysis/predefined/security-analysis.ts
```diff
@@ -27,6 +27,7 @@ const protocolTaint = (path: unknown) =>
 	typeof path === 'string' && NetworkProtocolRegex.test(path) ? NetworkInput : FileInput;
 
 export const securityAnalysis = new TaintAnalysisDefinition('security', securityDomain)
+	.withDefaultPropagators()
```
EagleoutIce: not the javascript event system suppressdefault :sob:

EagleoutIce: and i was hoping one could get taint-analysis semantic labels and then write:

```
.on(GeneralClass.PureComputer, ....)
```

MaxAtoms:
There is at least one open to-do: Function categories should be addable on demand and via the higher-level `TaintAnalysis` class

src/taint-analysis/predefined/security-analysis.ts
```diff
@@ -26,100 +29,95 @@ const NetworkProtocolRegex = /^(https?|ftps?):\/\//;
...
 			],
 			condition: {
 				argValues:   [{ pos: 0, name: 'con' }],
 				conditionFn: ([path]) => protocolTaint(path)
 			}
 		},
 	])
-	.through([])
-	.to([
+	.through([
```
MaxAtoms, line 65: As we do not know of the origin of the taints, all sanitizers here are handled the same. We would either need to track taint origins or create separate analyses to make this more granular.

src/taint-analysis/predefined/security-analysis.ts
```diff
@@ -26,100 +29,95 @@ const NetworkProtocolRegex = /^(https?|ftps?):\/\//;
...
 		{
+			identifier: [ 'shQuote', PkgName.Base ],
+			taint:      Top
+		},
+		{
+			identifier: [['match.arg', PkgName.Base], ['make.names', PkgName.Base]],
```
MaxAtoms, line 71: Taken from linting rule 🙃

src/taint-analysis/predefined/security-analysis.ts
```diff
@@ -26,100 +29,95 @@ const NetworkProtocolRegex = /^(https?|ftps?):\/\//;
...
 
+/** Tainted input on any argument leads to Bottom taint at sinks */
+const securitySinkCondition: TaintConditionFunction<typeof securityDomain> =
+	(_args, taints) => taints.some(taint => taint.value === UserInput || taint.value === NetworkInput || taint.value === FileInput)
+		? Bottom : AbstractDomain.joinAll(taints).value;
+
 export const securityAnalysis = TaintAnalysisDefinition.create('security', securityDomain)
 	.from([
 		{
 			identifier: [...BuiltInIndex.default().with(SemanticCallTag.User)],
```
MaxAtoms, line 40: `scan` is a conflict here: it will be mapped twice, tainted as UserInput, but might be file input as well

src/taint-analysis/predefined/randomness-analysis.ts
```diff
@@ -12,40 +18,118 @@ export const randomnessDomain = new FiniteDomainBuilder<Top, Bottom, [Top, Botto
...
+			return Random;
+		}
+		if(taints.length > 0) {
+			return AbstractDomain.joinAll(taints).value;
+		}
+		return Top;
+	})
+	.from([
+		{
+			identifier: [
```
MaxAtoms, line 36: Compared to the previous version, I removed `c` as source and left out `data.frame` and others as they might be created with random data, invaliding Determinism assumptions

src/taint-analysis/predefined/scale-analysis.ts
```diff
@@ -45,62 +46,78 @@ export const scaleAnalysis = TaintAnalysisDefinition.create('scale', scaleDomain
...
 			taint: Unscaled
 		},
-		// dropping elements removes our assumptions
+		{
+			identifier: ['round', PkgName.Base],
+			condition:  {
+				argTaints:   [{ pos: 0, name: 'x' }],
+				// Rounding only removes centering and variance assumption, min-max taint is kept
+				conditionFn: (_args, [taint]) =>
+					taint.value == ZScore || taint.value == ZeroCentered || taint.value == UnitVariance ? Unscaled : taint.value
```
MaxAtoms, line 95: I am unsure here whether it makes sense to say that the data is "Unscaled" after rounding as it still significantly deviates from the original data. For the functions below I chose Top which is probably more appropriate in most/all (?) of the transformer cases

src/taint-analysis/predefined/scale-analysis.ts
```diff
@@ -45,62 +46,78 @@ export const scaleAnalysis = TaintAnalysisDefinition.create('scale', scaleDomain
...
-		{ identifier: 'sd', condition: constantAggregate(UnitVariance, ZScore) },
-		{ identifier: 'var', condition: constantAggregate(UnitVariance, ZScore) },
-		{ identifier: ['min', 'max', 'range'], condition: constantAggregate(MinMax) }
+		{ identifier: 'mean', condition: checkCalcOnNormalizedInput(ZeroCentered, ZScore) },
+		{ identifier: 'sd', condition: checkCalcOnNormalizedInput(UnitVariance, ZScore) },
+		{ identifier: 'var', condition: checkCalcOnNormalizedInput(UnitVariance, ZScore) },
+		{ identifier: ['min', 'max', 'range'], condition: checkCalcOnNormalizedInput(MinMax) }
 	])
-	.report('Aggregation of scaled data yields a known constant');
+	.report('Known summary statistic calculated on normalized data');
```
MaxAtoms, line 123: The analysis checks whether we appear to statically know the result because of normalization. In practice it would probably make sense to just say that all of these statistics appear to be mistakenly calculated on normalized data, regardless of the normalization specifics. However, that would negate the need for the more elaborate lattice...

src/taint-analysis/predefined/randomness-analysis.ts
```diff
@@ -12,40 +18,118 @@ export const randomnessDomain = new FiniteDomainBuilder<Top, Bottom, [Top, Botto
 	.addLeqOrder(Deterministic, Top)
 	.build();
 
+const randomnessSinkCondition: TaintConditionFunction<typeof randomnessDomain> =
+	(_args, taints) => taints.some(taint => taint.value === Random) ? Bottom : AbstractDomain.joinAll(taints).value;
```
MaxAtoms, line 22: I decided to not join the incoming taints here and rather set the taint to Random whenever any argument has a random taint. Wdyt?

src/taint-analysis/predefined/security-analysis.ts
EagleoutIce: okay but we could get around this right? we could filter for those with any of the tags of interest and thenprovide a mapping to taints

src/taint-analysis/predefined/security-analysis.ts
EagleoutIce: You could also, on request ask for origins in the handler but this is fine for now (or we could add a new taint for the shQuoting

src/taint-analysis/predefined/security-analysis.ts
```diff
@@ -26,100 +29,95 @@ const NetworkProtocolRegex = /^(https?|ftps?):\/\//;
...
 				argValues:   [{ pos: 0, name: 'con' }],
 				conditionFn: ([path]) => protocolTaint(path)
 			}
 		},
 	])
-	.through([])
-	.to([
+	.through([
 		{
+			identifier: [ 'shQuote', PkgName.Base ],
```
EagleoutIce, line 67: theoretically there should also be a way to get this from a semantic tag :D

src/taint-analysis/predefined/randomness-analysis.ts
EagleoutIce: ok

src/taint-analysis/predefined/scale-analysis.ts
EagleoutIce: I agree

src/taint-analysis/predefined/scale-analysis.ts
EagleoutIce: I think generally the detailed lattice can provide important context when addressing the report

src/taint-analysis/predefined/randomness-analysis.ts
EagleoutIce: I guess this is fine :D - I guess we could also write lattice generalizations on this but for now it is good just like it is :D (the main question would be why TOP will *not* map to a (potential) bottom in a sink)

MaxAtoms: > the main question would be why TOP will not map to a (potential) bottom in a sink

What do you mean by this?

src/taint-analysis/predefined/security-analysis.ts
MaxAtoms: We can filter for functions that have both, the `SemanticCallTag.File` and `SemanticCallTag.User`. But how do we decide which taint to assign (without introducing a new combined taint `UserOrFile`)?

EagleoutIce: I mean without the set-lift we cannot. but generally we compute the lub which might bring us back to top, which is less precise but still correct.

src/taint-analysis/predefined/randomness-analysis.ts
EagleoutIce: this way if any value is exactly `Random` we get bottom, but if all of them are, e.g. `Top`, which includes Random. we get top?

PR #2758, opened 2026-09-10
Improvements to the DSL

MaxAtoms:
- TaintAnalysis#run optionally accepts a flowR analyzer instance, removing the need to set an analyzer outright and allowing for analysis re-use
- Method call ordering is now enforced via interfaces:
  - at least one analysis definition needs to be added before TaintAnalysis#run is available
  - TaintAnalysisDefinitions must be created via the from -> through -> to -> report chain (in practice it might make sense to be more lenient here...)


src/queries/catalog/taint-query/taint-query-executor.ts
```diff
@@ -10,16 +10,24 @@ import type { AllPredefinedTaintAnalysisNames } from '../../../taint-analysis/pr
...
+			},
+		};
 	}
 
-	const start = Date.now();
-
-	const analysis = analyzer.taint<AllPredefinedTaintAnalysisNames>();
+	const [firstDef, ...restDefs] = flattened;
+	const analysis = analyzer
+		.taint<AllPredefinedTaintAnalysisNames>()
```
EagleoutIce, line 27: We could get the type from the queires :D but that is fine

src/queries/catalog/taint-query/taint-query-executor.ts
```diff
@@ -10,16 +10,24 @@ import type { AllPredefinedTaintAnalysisNames } from '../../../taint-analysis/pr
...
+		.addPredefined(firstDef);
 
-	for(const def of flattened) {
+	for(const def of restDefs) {
 		analysis.addPredefined(def);
```
EagleoutIce, line 31: is this to get the type correct? It feels weird to handle the first one "separate"

src/taint-analysis/builder/taint-analysis-definition.ts
```diff
@@ -38,11 +38,33 @@ export interface ComposeOptions {
...
+	report(msg: string): TaintAnalysisDefinition<Name, Domain>;
+}
+
+export interface TaintAnalysisToStage<Name extends string = string, Domain extends AnyAbstractDomain = AnyAbstractDomain> extends TaintAnalysisThroughStage<Name, Domain> {
+	/** Adds sink rules signaling findings by yielding Bottom. */
+	to(fnMapping: TaintMapper<Domain>): TaintAnalysisReportStage<Name, Domain>;
+}
+
+export interface TaintAnalysisThroughStage<Name extends string = string, Domain extends AnyAbstractDomain = AnyAbstractDomain> extends TaintAnalysisFromStage<Name, Domain> {
+	/** Adds propagator or sanitizer rules that determine the resulting taint of matching calls. */
```
EagleoutIce, line 52: why do through and from have the same docs?

src/taint-analysis/builder/taint-analysis.ts
```diff
@@ -66,66 +71,97 @@ export interface TaintInferenceResult {
...
+		return this as unknown as RunnableTaintAnalysis<readonly [...Defs, Name]>;
 	}
 
 	/**
 	 * Run one or multiple taint analyses.
-	 * Note: Requires a prior call to {@link TaintAnalysis.add}, {@link TaintAnalysis.addComposite}, or {@link TaintAnalysis.addPredefined} to add at least one taint analysis.
 	 */
-	public async run(): Promise<Map<Defs[number], TaintInferenceResult>> {
+	public async run(analyzer?: ReadonlyFlowrAnalysisProvider): Promise<Map<Defs[number], TaintInferenceResult>> {
+		const priorityAnalyzer = analyzer ?? this.analyzer;
```
EagleoutIce, line 153: you could also do a default in the run head: `run(analyzer: Read... = this.analyzer)`

src/taint-analysis/eval/eval.ts
```diff
@@ -90,8 +90,11 @@ async function analyzeScript(run: ScriptRun, analyzer: FlowrAnalyzer): Promise<s
 		analyzer.reset();
 		analyzer.addRequest(`${fileProtocol}${run.path}`);
 
-		const analysis = analyzer.taint<AllPredefinedTaintAnalysisNames>();
-		for(const def of allPredefinedTaintAnalysisNames) {
+		const [firstPredefined, ...restPredefined] = allPredefinedTaintAnalysisNames;
+		const analysis = analyzer
+			.taint<AllPredefinedTaintAnalysisNames>()
+			.addPredefined(firstPredefined);
```
EagleoutIce, line 96: aww a pattern - why not just offer a batch solution in the analyzer?

src/taint-analysis/predefined/randomness-analysis.ts
```diff
@@ -48,4 +48,4 @@ export const randomnessAnalysis = new TaintAnalysisDefinition('randomness', rand
 			Identifier.make('rsignrank', 'stats'),
 		],
 		taint: Random
-	}]);
+	}]).through([]).to([]).report('');
```
EagleoutIce, line 51: I guess its fine to make this explicit although it feels to be overly type safe? 
Like it would break the intuition to iterate over functions

test/functionality/taint-analysis/propagation.test.ts
```diff
@@ -20,23 +20,24 @@ const lattice = new FiniteDomainBuilder()
 	.addLeqOrder(TaintC, Top)
 	.build();
 
-const marker = new TaintAnalysisDefinition('marker', lattice)
+const marker = TaintAnalysisDefinition.create('marker', lattice)
 	.from([
 		{ identifier: Identifier.make('taint'), taint: TaintA },
 		{ identifier: Identifier.make('TaintB'), taint: TaintB },
-	]);
+	]).through([]).to([]).report('');
```
EagleoutIce, line 27: maybe we can also talk about the empty string?

MaxAtoms: Yeah, I was thinking about adding a method to "shortcut" the chain, although that kind of defeats its purpose. The predefined analyses now use all methods in the chain, but a shortcut could still be used in tests.

EagleoutIce: Its also that if feels like an anti-pattern to have report being a constant string vs. at least partially dynamic based on the handler which caused bot?

MaxAtoms: Which information would you like to consider per handler for a partially dynamic report message (besides source location and lexeme, _edit:_ and taint)?

EagleoutIce: I mean beside the taint the call which triggered it so just the NodeId would probably help.

MaxAtoms: I added a function `getPartialDefinition()` which shortcuts the method chaining.
And I added the possibility to create report messages dynamically based on a `FindingContext`:

image

Would be nice now to have the source function and to know the taint causing the bottom taint for the message...

EagleoutIce: Just as a random readability note: maybe we can compact all nodes that are top/bottom and render the others based on their location? :see_no_evil:

MaxAtoms: I added a separate teeny tiny issue #2766

